PR #2, opened 2020-09-12
Add documentation to Python backend

Tabrizian:
Add example and initial version of README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,47 @@
-# Repository Template
+<!--
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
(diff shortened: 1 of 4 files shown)


examples/add_sub.py
```diff
@@ -0,0 +1,146 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import numpy as np
+import sys
+import json
+
+import sys
+sys.path.append('../../')
```
deadeyegoodwin, line 32: Using relative paths like this is error prone. We should show triton_python_backend_utils.py in the same directory as add_sub.py.

README.md
```diff
@@ -1,4 +1,262 @@
...
+Also, you need to make a copy of [triton_python_backend_utils.py](src/resources/triton_python_backend_utils.py) available to your `model.py`.
+
+Your models directory should look like below:
+```
+models
+├── add_sub
+│   ├── 1
+│   │   └── model.py
+│   └── config.pbtxt
+└── triton_python_backend_utils.py
```
deadeyegoodwin, line 222: This level of the model repository should only have the model subdirs. We should show best practice as having triton_python_backend_utils.py in the same directory as model.py.

deadeyegoodwin (changes requested):

examples/add_sub.py
Tabrizian: Fixed

PR #7, opened 2020-09-17
Use the new TRITONBACKEND_BackendArtifacts API

Tabrizian:


src/python.cc
```diff
@@ -582,33 +582,26 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
...
 
   if (backend_config.Find("cmdline", &cmdline)) {
-    triton::common::TritonJson::Value python_lib;
-    if (cmdline.Find("python-lib", &python_lib)) {
-      RETURN_IF_ERROR(python_lib.AsString(&backend_state->python_lib));
-      found_py_lib_config = true;
-    }
-
     triton::common::TritonJson::Value python_runtime;
     if (cmdline.Find("python-runtime", &python_runtime)) {
```
deadeyegoodwin, line 589: Why do we need any command line flags? You don't document what this is so I'm guessing we don't need it.

Tabrizian: The only other flag is python runtime. By default it uses Python3. I will add it to the readme.

deadeyegoodwin (approved):

PR #10, opened 2020-09-25
Fix non-graceful shutdown

Tabrizian:


src/resources/startup.py
```diff
@@ -319,5 +358,11 @@ def sigterm_handler(signum, frame):
 
     server.add_insecure_port(FLAGS.socket)
     server.start()
+    time.sleep(4)
```
deadeyegoodwin, line 361: Why wait here on the main thread. Can't you just change the location of the sleep in the background thread?

src/resources/startup.py
```diff
@@ -293,10 +297,45 @@ def Execute(self, request, context):
         return execute_response
 
 
+def watch_connections(address, event):
+    number_of_failures = 0
+    timeout_sec = 1
+
+    while True:
```
deadeyegoodwin, line 304: So this makes sure that a connection is established within a timeout after the server starts? Is there a particular reason it is done this way? Why wait for 3 failures? If I have success, failure, failure, then things are ok? That's doesn't seem right. If the timeout is 12 seconds, why not just wait 12 seconds and then check once?

src/resources/startup.py
```diff
@@ -293,10 +297,45 @@ def Execute(self, request, context):
...
+        if len(sockets.socket_ref) == 1:
+            number_of_failures += 1
+        else:
+            number_of_failures = 0
+
+        # Sleep for 4 seconds before polling next time
+        time.sleep(4)
+
+
 if __name__ == "__main__":
```
deadeyegoodwin, line 334: I thought there was also a case where if triton just dies then the python process can be left running (zombie). How does this fix that case?

deadeyegoodwin (changes requested):

src/resources/startup.py
Tabrizian: The goal here is to check that always there is a socket other than this socket connect to the gRPC server. Two things can happen when there is a problem. First, the Python GRPC server may have stopped. Or the Triton might have exited non-gracefully. If the first case happened, the connection to the Python GRPC will fail. I just added this to make sure that there is not a temporary problem with the connection. I can probably reduce the number of failures to 2. But in the example you mentioned, if you have success, failure, failure, failure, then the Python backend will exit.

src/resources/startup.py
Tabrizian: Mentioned in the comment above how this fixes it. https://github.com/triton-inference-server/python_backend/pull/10/files/bb86d2ba9ddeba7df1a2dc3b461f7b423addd65e#r496131640

src/resources/startup.py
Tabrizian: Sure, I will update it and remove it from here.

src/resources/startup.py
deadeyegoodwin: I see... what you have is ok then.

deadeyegoodwin (approved):

PR #14, opened 2020-10-12
Fix triton_python_backend_utils and thread shutdown

Tabrizian:


src/resources/startup.py
```diff
@@ -40,6 +40,10 @@
 from grpc_channelz.v1 import channelz_pb2
 from grpc_channelz.v1 import channelz_pb2_grpc
 
+from grpc_channelz.v1 import channelz
```
deadeyegoodwin, line 43: Why do you import these again?

deadeyegoodwin (changes requested):

src/resources/startup.py
Tabrizian: I think it. I think I made a mistake when merging the conflicts.

deadeyegoodwin (approved):

PR #16, opened 2020-10-26
Fix KIND_GPU

Tabrizian:
Closes triton-inference-server/server#2157

PR #17, opened 2020-10-26
Fix environment variable propagation

Tabrizian:
Fixes the environment variable propagation issue mentioned in triton-inference-server/server#2157.

Should be merged after #16.

### DIFF
--- a/src/python.cc
+++ b/src/python.cc
@@ -24,24 +24,24 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-#include <chrono>
-#include <cstdio>
-#include <cstring>
-#include <ctime>
-#include <functional>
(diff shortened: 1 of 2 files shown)


deadeyegoodwin:
@Tabrizian  Please rebase so that is has only the env changes.

Tabrizian:
Rebased.

src/python.cc
```diff
@@ -234,7 +238,7 @@ TRITONSERVER_Error *ModelInstanceState::CreatePythonInterpreter() {
     subinterpreter_commandline[5] = pymodule_path_.c_str();
     subinterpreter_commandline[7] = name_.c_str();
     if (execvpe(subinterpreter_commandline[0],
```
deadeyegoodwin, line 240: Since you are just passing in the existing env you can just use execvp and not need to use "environ"

deadeyegoodwin (changes requested):

src/python.cc
Tabrizian: fixed. Also, @tanmayv25 pointed out the reason the formatting was inconsistent. I fixed that and reformatted the whole document.

deadeyegoodwin (approved):

PR #28, opened 2020-12-29
Fix np.object_ for TYPE_STRING

francescoclarifai:
`np.object` is `<class 'object'>`, while here, if I understand correctly the logic, we expect `<class 'numpy.object_'>`.
Note that this can alternatively be resolved by using the `==` operator (same value) instead of `is` (same reference).

```
>>> import numpy as np
>>> data = np.asarray(['text']).astype('object')
>>> data.dtype
dtype('O')
>>> data.dtype==np.object
True
>>> data.dtype==np.object_  
True
>>> data.dtype is np.object_
False
>>> data.dtype.type is np.object_
True
>>> np.object
<class 'object'>
>>> np.object_
<class 'numpy.object_'>
```

With this change I'm able to use the python backend for TYPE_STRING output.

### DIFF
--- a/src/resources/startup.py
+++ b/src/resources/startup.py
@@ -305,7 +305,7 @@ class PythonHost(PythonInterpreterServicer):
                 output_shape = output_np_array.shape
 
                 # We need to serialize TYPE_STRING
-                if output_np_array.dtype.type is np.object or output_np_array.dtype.type is np.bytes_:
+                if output_np_array.dtype.type is np.object_ or output_np_array.dtype.type is np.bytes_:
                     output_np_array = serialize_byte_tensor(output_np_array)
 
                 tensor = Tensor(name=output_tensor.name(),
--- a/src/resources/startup.py
+++ b/src/resources/startup.py
@@ -305,7 +305,7 @@ class PythonHost(PythonInterpreterServicer):
                 output_shape = output_np_array.shape
 
                 # We need to serialize TYPE_STRING
-                if output_np_array.dtype.type == np.object_ or output_np_array.dtype.type is np.bytes_:
+                if output_np_array.dtype == np.object or output_np_array.dtype.type is np.bytes_:
                     output_np_array = serialize_byte_tensor(output_np_array)
 
                 tensor = Tensor(name=output_tensor.name(),
--- a/src/resources/startup.py
+++ b/src/resources/startup.py
@@ -305,7 +305,7 @@ class PythonHost(PythonInterpreterServicer):
                 output_shape = output_np_array.shape
 
                 # We need to serialize TYPE_STRING
-                if output_np_array.dtype.type is np.object_ or output_np_array.dtype.type is np.bytes_:
+                if output_np_array.dtype.type == np.object_ or output_np_array.dtype.type is np.bytes_:
                     output_np_array = serialize_byte_tensor(output_np_array)
 
                 tensor = Tensor(name=output_tensor.name(),
--- a/src/resources/triton_python_backend_utils.py
+++ b/src/resources/triton_python_backend_utils.py
@@ -52,7 +52,7 @@ TRITON_TO_NUMPY_TYPE = {
     # TRITONSERVER_TYPE_FP64
     12: np.float64,
     # TRITONSERVER_TYPE_STRING
-    13: np.bytes_
+    13: np.object
 }
 
 TRITON_STRING_TO_NUMPY = {
@@ -68,7 +68,7 @@ TRITON_STRING_TO_NUMPY = {
     'TYPE_FP16': np.float16,
     'TYPE_FP32': np.float32,
     'TYPE_FP64': np.float64,
-    'TYPE_STRING': np.bytes_
+    'TYPE_STRING': np.object
 }
 
 NUMPY_TO_TRITON_TYPE = {v: k for k, v in TRITON_TO_NUMPY_TYPE.items()}


francescoclarifai:
@Tabrizian should I open an Issues in https://github.com/triton-inference-server/server instead of this PR? I noticed only now the instructions for contributing posted there.

Tabrizian:
@francescoclarifai The PR is opened correctly but I'm having trouble reproducing your error. Python backend already works with `TYPE_STRING`. You can use the example client below:

```
import argparse
import numpy as np
import sys

import tritonclient.grpc as grpcclient

if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('-v',
                        '--verbose',
                        action="store_true",
                        required=False,
                        default=False,
                        help='Enable verbose output')
    parser.add_argument('-u',
                        '--url',
                        type=str,
                        required=False,
                        default='localhost:8001',
                        help='Inference server URL. Default is localhost:8001.')
    parser.add_argument('-s',
                        '--ssl',
                        action="store_true",
                        required=False,
                        default=False,
                        help='Enable SSL encrypted channel to the server')
    parser.add_argument('-t',
                        '--client-timeout',
                        type=float,
                        required=False,
                        default=None,
                        help='Client timeout in seconds. Default is None.')
    parser.add_argument(
        '-r',
        '--root-certificates',
        type=str,
        required=False,
        default=None,
        help='File holding PEM-encoded root certificates. Default is None.')
    parser.add_argument(
        '-p',
        '--private-key',
        type=str,
        required=False,
        default=None,
        help='File holding PEM-encoded private key. Default is None.')
    parser.add_argument(
        '-x',
        '--certificate-chain',
        type=str,
        required=False,
        default=None,
        help='File holding PEM-encoded certicate chain. Default is None.')

    FLAGS = parser.parse_args()
    try:
        triton_client = grpcclient.InferenceServerClient(
            url=FLAGS.url,
            verbose=FLAGS.verbose,
            ssl=FLAGS.ssl,
            root_certificates=FLAGS.root_certificates,
            private_key=FLAGS.private_key,
            certificate_chain=FLAGS.certificate_chain)
    except Exception as e:
        print("channel creation failed: " + str(e))
        sys.exit()

    model_name = "python_object_object_object"

    # Infer
    inputs = []
    outputs = []
    inputs.append(grpcclient.InferInput('INPUT0', [1, 5], "BYTES"))
    inputs.append(grpcclient.InferInput('INPUT1', [1, 5], "BYTES"))

    # Create the data for the two input tensors. Initialize the first
    # to unique integers and the second to all ones.
    input0_data = np.array([["A", "B", "C", "D", "E"]], dtype=np.object)
    input1_data = np.array([["A", "B", "C", "D", "E"]], dtype=np.object)

    # Initialize the data
    inputs[0].set_data_from_numpy(input0_data)
    inputs[1].set_data_from_numpy(input1_data)

    outputs.append(grpcclient.InferRequestedOutput('OUTPUT0'))
    outputs.append(grpcclient.InferRequestedOutput('OUTPUT1'))

    # Test with outputs
    results = triton_client.infer(model_name=model_name,
                                  inputs=inputs,
                                  outputs=outputs,
                                  client_timeout=FLAGS.client_timeout,
                                  headers={'test': '1'})

    statistics = triton_client.get_inference_statistics(model_name=model_name)
    print(statistics)

    # Get the output arrays from the results
    output0_data = results.as_numpy('OUTPUT0')
    output1_data = results.as_numpy('OUTPUT1')

    # Test with no outputs
    results = triton_client.infer(model_name=model_name,
                                  inputs=inputs,
                                  outputs=None)

    # Get the output arrays from the results
    output0_data = results.as_numpy('OUTPUT0')
    output1_data = results.as_numpy('OUTPUT1')

    print('PASS: infer')
```

francescoclarifai:
@Tabrizian I realized I'm actually seeing two different issues, in 20.10 and 20.12 respectively. The fix in this PR addresses the 20.10 issue (also present in 20.12), but not the 20.12.

Here is a basic example to reproduce the problem:
model.py:
```
import numpy as np
import string
import triton_python_backend_utils as pb_utils

class TritonPythonModel:

    def initialize(self, args):
        self.character = list(string.printable[:-6])
        
    def execute(self, requests):

        responses = []
        
        for request in requests:
            in_0 = pb_utils.get_input_tensor_by_name(request, "indices")
            batch_indices = in_0.as_numpy()

            texts = []
            for indices in batch_indices:

                text = ''.join([self.character[idx] for idx in indices])
                texts.append(text)

            texts_out = np.asarray(texts).astype('object')
            out_tensor_0 = pb_utils.Tensor("text", texts_out)
            inference_response = pb_utils.InferenceResponse(output_tensors=[out_tensor_0])
            responses.append(inference_response)

        return responses
```
config.pbtxt:
```
max_batch_size: 1
input {
  name: "indices"
  data_type: TYPE_INT64
  dims: 26
}
output {
  name: "text"
  data_type: TYPE_STRING
  dims: 1
}
instance_group {
  count: 1
  kind: KIND_CPU
}
dynamic_batching {
  preferred_batch_size: 1
  max_queue_delay_microseconds: 500
}
backend: "python"
```
client script:
```
import numpy as np
from tritonclient.grpc import InferenceServerClient, InferInput, InferRequestedOutput

input_array = np.random.randint(0,90,(1,26)).astype(np.int64)

model_input = InferInput('indices', input_array.shape, 'INT64')
model_input.set_data_from_numpy(input_array)

model_output = InferRequestedOutput('text')

triton_url = '127.0.0.1:8001'
triton_client = InferenceServerClient(triton_url)

triton_client.load_model('python_model')
res = triton_client.infer(model_name='python_model', inputs=[model_input], outputs=[model_output])
print(res.as_numpy('text'))
```

In 20.10 (client 2.4.0) this throws the following exception (resolved by the change proposed here):
`tritonclient.utils.InferenceServerException: [StatusCode.INTERNAL] GRPC Execute Failed, message: Exception calling application: <class 'numpy.object_'>`

In 20.12 (client 2.6.0) this results into a different exception (maybe related to the ModelConfig protobuf parsing?):
`tritonclient.utils.InferenceServerException: [StatusCode.INTERNAL] GRPC Execute Failed, message: Exception calling application: 'TYPE_STRING' has type str, but expected one of: int, long`

src/resources/startup.py
```diff
@@ -305,7 +305,7 @@ def Execute(self, request, context):
                 output_shape = output_np_array.shape
 
                 # We need to serialize TYPE_STRING
-                if output_np_array.dtype.type is np.object or output_np_array.dtype.type is np.bytes_:
+                if output_np_array.dtype.type is np.object_ or output_np_array.dtype.type is np.bytes_:
```
Tabrizian, line 308: As you suggested, please change this to `output_np_array.dtype == np.object or output_np_array.dtype.type is np.bytes_`.

Tabrizian (changes requested):

Tabrizian:
Thanks for the complete report and PR. I was able to reproduce this error. You can change 
`np.asarray(texts).astype('object')` to `np.asarray(texts).astype(np.object)` for now until this PR is merged.

src/resources/startup.py
```diff
@@ -305,7 +305,7 @@ def Execute(self, request, context):
                 output_shape = output_np_array.shape
 
                 # We need to serialize TYPE_STRING
-                if output_np_array.dtype.type is np.object or output_np_array.dtype.type is np.bytes_:
+                if output_np_array.dtype.type == np.object_ or output_np_array.dtype.type is np.bytes_:
```
Tabrizian, line 308: You still need to remove `type` from `output_np_array.dtype.type`.

Tabrizian (changes requested):

src/resources/startup.py
francescoclarifai: Got it. Now it should be consistent with line 79.

src/resources/startup.py
```diff
@@ -305,7 +305,7 @@ def Execute(self, request, context):
                 output_shape = output_np_array.shape
 
                 # We need to serialize TYPE_STRING
-                if output_np_array.dtype.type is np.object or output_np_array.dtype.type is np.bytes_:
+                if output_np_array.dtype == np.object or output_np_array.dtype.type is np.bytes_:
```
CoderHam, line 308: Good catch. This also brought to my attention a relevant change that needs to be made in triton_python_backend_utils.py i.e. https://github.com/triton-inference-server/python_backend/blob/main/src/resources/triton_python_backend_utils.py#L71 and https://github.com/triton-inference-server/python_backend/blob/main/src/resources/triton_python_backend_utils.py#L55 where it should be `np.object` instead of `np.bytes_` since `np.bytes_` is not a dtype but a dtype.type. 

@Tabrizian the following lines need to be changed from `.dtype.type is np.object` to `.dtype == np.object` so you should open up a PR for the same
https://github.com/triton-inference-server/server/blob/master/qa/python_models/add_sub/model.py#L63 and https://github.com/triton-inference-server/server/blob/master/qa/python_models/sub_add/model.py#L63

CoderHam (changes requested):

src/resources/startup.py
Tabrizian: @CoderHam The decoder and encoder always create `np.bytes_` in these files: https://github.com/triton-inference-server/python_backend/blob/main/src/resources/triton_python_backend_utils.py#L71 and https://github.com/triton-inference-server/python_backend/blob/main/src/resources/triton_python_backend_utils.py#L55.

Check the deserializer that always creates np.bytes_: https://github.com/triton-inference-server/python_backend/blob/9e89c1018ef0a9cbd29c3c45ec0baffa7ccf0bc8/src/resources/startup.py#L130 (This is the same function from the Triton Python client)

I will create a separate PR to update the CI models.

Tabrizian:
@francescoclarifai Can you please revert your last commit?

Tabrizian:
Thanks for the PR. The issue is fixed in #31

PR #29, opened 2020-12-31
support get input buffer from GPU memory to CPU memory

ldengjie:
when cmake, add "-DTRITON_ENABLE_GPU=ON"


Tabrizian:
Thanks for the PR. I addressed this issue in #30.

PR #30, opened 2021-01-07
Add support for copying data from GPU to CPU when input is in GPU

Tabrizian:
* Refactored Python backend to use the new utility classes
* Used BackendInputCollector to copy data to the CPU.

### DIFF
--- a/src/python.cc
+++ b/src/python.cc
@@ -46,6 +46,9 @@
 
 #include "python_host.grpc.pb.h"
 #include "triton/backend/backend_common.h"
+#include "triton/backend/backend_input_collector.h"
+#include "triton/backend/backend_model.h"
+#include "triton/backend/backend_model_instance.h"
 #include "triton/common/triton_json.h"
 #include "triton/core/tritonbackend.h"
 #include "triton/core/tritonserver.h"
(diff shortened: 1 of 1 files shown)


src/python.cc
```diff
@@ -450,6 +394,12 @@ ModelInstanceState::GetInputTensor(
       in, &input_name, &input_dtype, &input_shape, &input_dims_count,
       &input_byte_size, &input_buffer_count));
 
+  // We need to create a new collector for every request because Python backend
+  // doesn't batch requests.
```
deadeyegoodwin, line 398: What does this mean? The backend can receive a batch of requests, but are you saying that we don't actually collect each tensor into a contiguous buffer because that is not how they are delivered to the python code?

src/python.cc
```diff
@@ -466,88 +416,65 @@ ModelInstanceState::GetInputTensor(
...
-    if (input_memory_type == TRITONSERVER_MEMORY_GPU) {
-      RETURN_IF_ERROR(TRITONSERVER_ErrorNew(
-          TRITONSERVER_ERROR_UNSUPPORTED,
-          "failed to get input buffer in CPU memory"));
-    }
-    data_buffer->append((const char*)input_buffer, buffer_byte_size);
-  }
+  RETURN_IF_ERROR(BackendMemory::Create(
+      model_state_->TritonMemoryManager(),
+      {BackendMemory::AllocationType::CPU_PINNED_POOL,
```
deadeyegoodwin, line 424: Pinned memory is "expensive" and so should only use it if going to copy to/from GPU.

deadeyegoodwin (changes requested):

src/python.cc
Tabrizian: I meant in the other backends like onnxruntime, a single buffer is created that contains all the requests and they do not call `ProcessTensor` for each request. But here we have to call it for every request because we do not batch multiple requests into a single request.

deadeyegoodwin: So say "We need to create a new collector for every request because python backend sends each request individually to the python model".

src/python.cc
```diff
@@ -466,88 +416,65 @@ ModelInstanceState::GetInputTensor(
...
+       BackendMemory::AllocationType::CPU},
+      input_memory_type_id, input_byte_size, &input_memory));
+
+  input_tensor_memories_.push_back(input_memory);
```
deadeyegoodwin, line 428: I don't see where you use input_tensor_memories_.
Also, you are allocating a buffer and then copying into it but later I assume you are then copying from this buffer into a protobuf buffer for sending to the python code. You should just copy directly into the protobuf to avoid the allocation and copy. You can size the protobuf field and then pass a pointer to it into the corrector. We have examples of this in grpc_server.cc I believe.

deadeyegoodwin (changes requested):

src/python.cc
Tabrizian: I found this. Thanks for the pointer!

deadeyegoodwin (approved):

PR #32, opened 2021-01-27
Fix: ModelInstanceState name_ property should not set #2457

zjjott:
fix https://github.com/triton-inference-server/server/issues/2457


Tabrizian:
Thanks for this PR. It is addressed in #33.

PR #36, opened 2021-02-18
Fix np.object_ for Python backend

Tabrizian:


src/resources/startup.py
```diff
@@ -128,7 +131,7 @@ def deserialize_bytes_tensor(encoded_tensor):
         sb = struct.unpack_from("<{}s".format(l), val_buf, offset)[0]
         offset += l
         strs.append(sb)
-    return (np.array(strs, dtype=bytes))
+    return (np.array(strs, dtype=np.bytes_))
```
deadeyegoodwin, line 134: We need to use np.object here to avoid removal of zero bytes. Also in the description we should state that the returned numpy array will have elements of type np.object.

src/resources/startup.py
```diff
@@ -56,14 +56,17 @@ def serialize_byte_tensor(input_tensor):
         Serializes a bytes tensor into a flat numpy array of length prepend bytes.
         Can pass bytes tensor as numpy array of bytes with dtype of np.bytes_,
```
deadeyegoodwin, line 57: When need to enforce (or at least educate) users to use np.object for their BYTES tensors and point explicitly to the problem with np.bytes_ and np.str. Even though these are internal we should be clear in what proper usage is.

deadeyegoodwin (changes requested):

src/resources/startup.py
Tabrizian: I will fix this in a subsequent PR. This is only to ensure that `np.object_` is parsed correctly.

src/resources/startup.py
Tabrizian: I improved this docstring.

src/resources/startup.py
```diff
@@ -53,29 +53,34 @@
...
-        -------
-        serialized_bytes_tensor : np.array
-            The 1-D numpy array of type uint8 containing the serialized bytes in 'C' order.
-        Raises
-        ------
-        InferenceServerException
-            If unable to serialize the given tensor.
-        """
+    Serializes a bytes tensor into a flat numpy array of length prepended bytes.
+    Can pass bytes tensor as numpy array of bytes with dtype of np.bytes_,
```
deadeyegoodwin, line 57: Shouldn't we delete the "Can pass bytes ... of np.bytes_" since that will strip the trailing zeros. Instead:
"The numpy array should use dtype of np.object_. For np.str_ and np.bytes_, numpy will remove ... "

deadeyegoodwin (changes requested):

src/resources/startup.py
deadeyegoodwin: You don't need to rerun CI

deadeyegoodwin (approved):

PR #40, opened 2021-03-17
Fix batch size reporting

Tabrizian:


GuanLuo (dismissed):

CoderHam (approved):

src/python.cc
```diff
@@ -355,6 +356,12 @@ ModelInstanceState::~ModelInstanceState()
   }
 }
 
+ModelState*
```
dzier, line 359: You should return a shared pointer. Is ModelInstance supposed to be a struct or class? If it is a struct, just make model_state a public variable. If not, then don't return raw data. Rather, create methods that can be applied in the object.

dzier: For example, you really only need `name` and `max_batch_size`, so create two methods that return just that data, so you are not passing pointers to internal data. That is bad and can be prone to problems down the road.  For example...

```
ModelState * state = instance_state->GetModelState();

.... Do useful work ....

delete state;  // oops, you just hosed the system by accident.
```

dzier (changes requested):

src/python.cc
dzier: Even more importantly, if you do return a pointer, make it a const pointer unless you intend to alter the data.

src/python.cc
```diff
@@ -118,6 +118,8 @@ class ModelInstanceState : public BackendModelInstance {
       std::vector<TRITONBACKEND_Response*>& responses, size_t r,
       uint32_t& batch_size);
 
+  ModelState* GetModelState();
```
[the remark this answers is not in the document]
GuanLuo: You just need to static cast it from `BackendModel` class to your `ModelState` class

src/python.cc
Tabrizian: > You should return a shared pointer. Is ModelInstance supposed to be a struct or class? If it is a struct, just make model_state a public variable. If not, then don't return raw data. Rather, create methods that can be applied in the object.

ModelInstanceState is a class. 

I see, thanks for the comment. I ended up deleting that method. It looks like that field was available in the parent class.

src/python.cc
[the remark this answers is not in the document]
Tabrizian: Right. I was going through the onnx backend code to see how it's done. It looks like similar changes can be applied there too.

src/python.cc
```diff
@@ -642,12 +640,78 @@ TRITONBACKEND_ModelInstanceExecute(
...
 
-  std::vector<TRITONBACKEND_Response*> responses;
-  responses.reserve(request_count);
-
+  ModelState* model_state =
+      reinterpret_cast<ModelState*>(instance_state->Model());
   uint64_t exec_start_ns = 0;
   SET_TIMESTAMP(exec_start_ns);
 
+  const int max_batch_size = model_state->MaxBatchSize();
```
dzier, line 648: primitive datatypes do not need to be const.

dzier (changes requested):

src/python.cc
Tabrizian: Fixed.

dzier (approved):

PR #42, opened 2021-03-19
Accelerate Python backend using shared memory

Tabrizian:
- [x] Create Shared Memory Pool
- [x] Create the request objects in shared memory (Main Process -> Child Process)
- [x] Pass the request objects to the Python model
- [x] Receive response back from the Python model
- [x] Create the response objects in shared memory (Child Process -> Main Process)
- [x] Proper clean-up of the mapped addresses.
- [x] Error handling
- [x] Non-graceful shutdown
- [x] Add a parameter to control the default memory mapped page.
- [x] Create separate Python stub
- [X] ~~Add test for shared memory growth~~  (Filed separate ticket)
- [X] Write perf tests (Will be included in nightly reports.)
- [X] ~~Add raw and NumPy interfaces~~ (Filed separate ticket)
- [X] Fix serialize byte size


src/pb_utils.h
```diff
@@ -0,0 +1,136 @@
...
+#include <memory>
+#include <vector>
+#include <pthread.h>
+
+namespace triton { namespace backend { namespace python {
+
+// Create a conditional variable.
+void CreateIPCCondVariable(pthread_cond_t** cv);
+
+// Create a mutex that is shared between different processes.
```
CoderHam, line 43: Duplicate

src/shm_manager.cc
```diff
@@ -0,0 +1,205 @@
...
+SharedMemory::SharedMemory(const std::string& shm_key) {
+  shm_fd_ = shm_open(shm_key.c_str(), O_RDWR | O_CREAT | O_TRUNC, S_IRUSR | S_IWUSR);
+  if (shm_fd_ == -1) {
+    TRITONSERVER_Error* err = TRITONSERVER_ErrorNew(TRITONSERVER_ERROR_INTERNAL,
+        ("unable to get shared memory descriptor for shared-memory key '" +
+        shm_key + "'").c_str());
+    throw BackendModelException(err);
+  }
+
+  size_t default_byte_size =  PAGE_SIZE;
```
CoderHam, line 49: Use a large size such as 128MB instead of page size which is usually 4/8/16kb. When needed mmap additional shm regions of size 128MB.

CoderHam (changes requested):

CoderHam:
Can you also test with a 16kb payload for throughput? Refer to L0_perf_nomodel

src/shm_manager.cc
Tabrizian: Fixed.

Tabrizian:
@CoderHam I tried the L0_perf_nomodel. The `max_throughput` increased from  2444.6 to 14990.6.

In the 16MB payload, throughput increased from 41.8 to 87.2.

The min latency for grpc decreased from 1821 to 925.

Tabrizian:
There are some minor test failures that I'm working on but I didn't want to make this PR any larger than this.

src/pb_utils.cc
```diff
@@ -0,0 +1,157 @@
...
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <cstring>
+#include <memory>
+#include <string>
+
+#include "pb_utils.h"
+#include "shm_manager.h"
+//#include "triton/backend/backend_common.h"
```
CoderHam, line 42: Remove this

src/shm_manager.cc
```diff
@@ -0,0 +1,189 @@
...
+    throw BackendModelException(err);
+  }
+}
+
+TRITONSERVER_Error*
+SharedMemory::Map(char** shm_addr, size_t byte_size, off_t& offset)
+{
+  while (*offset_ + byte_size >= *capacity_) {
+    // Increase the shared memory pool size by one page size.
+    *capacity_ = *offset_ + byte_size + PAGE_SIZE;
```
CoderHam, line 127: Why do we only create the capacity by one PAGE_SIZE instead of that fixed size we spoke about earlier? Wil you address this in a future PR?

src/shm_manager.cc
```diff
@@ -0,0 +1,189 @@
...
+      return err;
+    }
+  }
+
+  // Update shared memory pointer and capacity if necessary.
+  RETURN_IF_ERROR(UpdateSharedMemory());
+
+  *shm_addr = shm_addr_ + *offset_;
+  offset = *offset_;
+
```
CoderHam, line 143: Empty line not needed

src/python.cc
```diff
@@ -112,303 +78,1187 @@ class ModelState : public BackendModel {
...
+          responses, r,
+          TRITONBACKEND_ResponseOutput(
+              response, &triton_output, name, triton_dt, dims, dims_count));
+
+      uint64_t output_byte_size = raw_data->byte_size;
+
+      // Custom handling for TRITONSERVER_TYPE_BYTES
+      if (triton_dt == TRITONSERVER_TYPE_BYTES) {
+        //
+      } else {
```
CoderHam, line 531: Remove the empty if else, Add a TODO to the comment if needed

CoderHam (changes requested):

src/pb_utils.h
```diff
@@ -0,0 +1,212 @@
...
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include "shm_manager.h"
+#include "triton/backend/backend_common.h"
+#include "triton/core/tritonserver.h"
+
```
deadeyegoodwin, line 32: Convention is to not put blank line between system and local includes, and just let the code formatter order them.

deadeyegoodwin (changes requested):

src/shm_manager.cc
Tabrizian: I actually think 128MB is a bit large as the default value.  Because this shared memory region is per model instance which can be a lot if the user has a lot of models (e.g. 10 models will be 1280MB). But I'll fix the growth in this PR along with a paramter to make it controllable.

Tabrizian: I added the parameter.

PR #44, opened 2021-04-09
Set maximum tensor size

Tabrizian:


src/python.cc
```diff
@@ -1042,22 +1036,29 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
     }
 
     triton::common::TritonJson::Value shm_growth_size;
+    std::string shm_growth_byte_size;
     if (cmdline.Find("shm-growth-byte-size", &shm_growth_size)) {
-      RETURN_IF_ERROR(
-          shm_growth_size.AsInt(&backend_state->shm_growth_byte_size));
+      RETURN_IF_ERROR(shm_growth_size.AsString(&shm_growth_byte_size));
     }
+    backend_state->shm_growth_byte_size =
```
CoderHam, line 1043: You should add some error handling in case stoi fails

src/python.cc
```diff
@@ -912,27 +901,32 @@ ModelInstanceState::GetInputTensor(
...
 
-  RESPOND_AND_RETURN_IF_EXCEPTION(
-      request, TRITONBACKEND_InputProperties(
-                   in, &input_name, &input_dtype, &input_shape,
-                   &input_dims_count, &input_byte_size, &input_buffer_count));
+  RETURN_IF_ERROR(TRITONBACKEND_InputProperties(
+      in, &input_name, &input_dtype, &input_shape, &input_dims_count,
+      &input_byte_size, &input_buffer_count));
+
+  // If input_byte_size is larger than 2GBs, reject request the request.
```
CoderHam, line 908: You mention 4Gb and 2Gb here. Which is correct?

src/python.cc
```diff
@@ -707,7 +696,9 @@ ModelInstanceState::ProcessRequests(
 TRITONSERVER_Error*
 ModelInstanceState::SetupChildProcess()
 {
-  std::string shm_region_name = std::string("/") + Name();
+  std::string kind = TRITONSERVER_InstanceGroupKindString(kind_);
+  std::string shm_region_name =
```
CoderHam, line 700: Use '_' to separate name, kind and device_id for readability

CoderHam (changes requested):

PR #47, opened 2021-04-23
Make sure that the output buffers are in CPU

Tabrizian:
~~Needs to be rebased after https://github.com/triton-inference-server/python_backend/pull/45 is merged.~~ Done

### DIFF
--- a/src/python.cc
+++ b/src/python.cc
@@ -613,6 +613,17 @@ ModelInstanceState::ProcessRequests(
         &request, 1 /* request count */, &response, 0 /* max_batch_size */,
         model_state->TritonMemoryManager(), model_state->EnablePinnedInput(),
         CudaStream());
+
+    bool cuda_copy = false;
+    std::set<std::string> requested_output_names;
+    for (size_t j = 0; j < requested_output_count; ++j) {
+      const char* output_name;
+      GUARDED_RESPOND_IF_ERROR(
+          responses, r,
+          TRITONBACKEND_RequestOutputName(request, j, &output_name));
+      requested_output_names.insert(output_name);
+    }
+
     for (size_t j = 0; j < requested_output_count; ++j) {
       Tensor* output_tensor = &output_tensors[j];
       TRITONSERVER_DataType triton_dt = output_tensor->dtype;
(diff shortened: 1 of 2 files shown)


deadeyegoodwin (approved):

src/python.cc
```diff
@@ -607,14 +609,25 @@ ModelInstanceState::ProcessRequests(
             (char**)&output_tensors, sizeof(Tensor) * requested_output_count,
             response_shm->outputs));
 
+    BackendOutputResponder responder(
+        &request, 1 /* request count */, &response, 0 /* max_batch_size */,
```
GuanLuo, line 613: It is strange that you need one responder for every response instead of using one responder for all responses. What is the reason behinds it?

Tabrizian: @GuanLuo The reason is that I don't know how I can skip the Tensor processing for a request in case there is an error. As far as I have understood, `ProcessTensor` will do the processing of the specific tensor for all of the `requests` in the `responder`. I'm not sure how I can skip a single response using `ProcessTensor`.

GuanLuo: The output responder will handle the skipping when there is error or the response has been sent (due to earlier error)

Tabrizian: @GuanLuo I see. There is one other problem too. In Python backend, the output buffers are stored in separate `char *` pointers (as each request is handled independently) and are not in a single contiguous area. As far as I have understood, `ProcessTensor` expects all the output buffers inside a single pointer. Would it be possible to pass an array of separate buffers instead?

GuanLuo: `ProcessTensor` expects the data of a output is in contiguous buffer. So Python backend doesn't really support dynamic batching, otherwise I don't see why the output buffer will not be contiguous.

GuanLuo: Given that, I don't think we have a good reason to use output responder, all it does is the same as the code for [handling bytes data type](https://github.com/triton-inference-server/python_backend/pull/47/files#diff-9209a7d49d738ccae1862624eb9a79e7264f42779d73b3e66993b62aad8a981cR667-R688)

Tabrizian: Yeah, I'll change it to only use the `CopyBuffer`. The reason I made this change was because of an error I noticed when the output is in GPU. I wanted to make sure that the output is in CPU.

GuanLuo (approved):
Just need to update the PR title accordingly

PR #50, opened 2021-05-06
Modify incorrect information in READMD.md

qpakzk:
* Correct the wrong url for Triton Python Client Library

* Correct the wrong path for client.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ $ git clone https://github.com/triton-inference-server/python_backend -b r<xx.yy
 
 7. Run the example client.
 ```
-$ python3 python_backend/examples/add_sub_client.py
+$ python3 python_backend/examples/add_sub/client.py
 ```
 
 ## Building from Source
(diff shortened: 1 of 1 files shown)


Tabrizian:
Thanks for your contribution. However, I think that the user is able to find out how to use the gRPC client by reading the docs so I don't think it is necessary to include both clients for both of the clients for the examples.

I noticed that you found an incorrect documentation link. Can you please only include that change?

README.md
```diff
@@ -342,8 +342,8 @@ small.
 # Examples
 
 For using the Triton Python client in these examples you need to install
-the [Triton Python Client Library](https://github.com/triton-inference-server/server/blob/master/docs/client_libraries.md#getting-the-client-libraries).
-The Python client for each of the examples is in the `client.py` file.
+the [Triton Python Client Library](https://github.com/triton-inference-server/client#getting-the-client-libraries-and-examples).
```
Tabrizian, line 345: Please only include this change in your PR.

Tabrizian (changes requested):

README.md
qpakzk: @Tabrizian I updated my commit.

Tabrizian: Thanks, can you add "The Python client for each of the examples is in the `client.py` file."  back too?

qpakzk: I didn't delete that sentence. You can check it in my commit.
I don't know why that sentence is minus.

Tabrizian (approved):

PR #51, opened 2021-05-06
Fix a concurrency bug when the stub is starting

Tabrizian:


src/python.cc
```diff
@@ -799,6 +799,7 @@ ModelInstanceState::SetupChildProcess()
   }
 
   parent_pid_ = getpid();
+  pthread_mutex_lock(parent_mutex_);
```
deadeyegoodwin, line 802: InstanceFinalize is not called unless InstanceInitialize returns with a non-error code. Will the parent mutex be left in a locked state if initialization fails? Does that matter?

Tabrizian: The parent will not be in the locked state because the wait for the condition variable is timed. It will wait for some default value until the stub notifies the parent process. The problem here was that the parent didn't lock the mutex but child notified the parent and the notification was lost.

src/python.cc
```diff
@@ -1063,7 +1065,7 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
   triton::common::TritonJson::Value cmdline;
   backend_state->shm_default_byte_size = 64 * 1024 * 1024;  // 64 MBs
   backend_state->shm_growth_byte_size = 64 * 1024 * 1024;   // 64 MBs
-  backend_state->stub_timeout_seconds = 10;
+  backend_state->stub_timeout_seconds = 30;
```
CoderHam, line 1068: If you are changing the default value of stub timeout, you should update the docs as well. However what is the motivation to increase the default timeout if you are anyway increasing it to 120 in the L0_infer test

CoderHam (changes requested):

deadeyegoodwin (approved):

PR #55, opened 2021-05-20
Add support for custom python environments

Tabrizian:
Directory structure of Python models using custom Python backend stubs:

```
.
├── python_3_6
│   ├── 1
│   │   ├── model.py
│   │   └── __pycache__
│   │       └── model.cpython-36.pyc
│   ├── config.pbtxt
│   └── triton_python_backend_stub
└── python_3_9
    ├── 1
    │   ├── model.py
    │   └── __pycache__
    │       └── model.cpython-39.pyc
    ├── config.pbtxt
    └── triton_python_backend_stub
```

An example model configuration:
```
name: "python_3_6"
backend: "python"

input [
  {
    name: "INPUT"
    data_type: TYPE_FP32
    dims: [ 1 ]
  }
]
output [
  {
    name: "OUTPUT"
    data_type: TYPE_FP32
    dims: [ 1 ]
  }
]

instance_group [{ kind: KIND_CPU }]
parameters: {key: "EXECUTION_ENV_PATH", value: {string_value: "/opt/tritonserver/qa/L0_backend_python/python3.6.tar.gz"}}
```

### DIFF
--- a/src/python.cc
+++ b/src/python.cc
@@ -248,7 +248,7 @@ class ModelInstanceState : public BackendModelInstance {
 
 ModelInstanceState::ModelInstanceState(
     ModelState* model_state, TRITONBACKEND_ModelInstance* triton_model_instance)
-    : BackendModelInstance(model_state, triton_model_instance),
+    : BackendModelInstance(model_state, triton_model_instance), child_pid_(0),
       initialized_(false)
 {
 }
@@ -745,8 +745,15 @@ ModelInstanceState::SetupChildProcess()
   int64_t shm_default_size =
       model_state->StateForBackend()->shm_default_byte_size;
 
-  shm_pool_ = std::make_unique<SharedMemory>(
-      shm_region_name, shm_default_size, shm_growth_size, true /* truncate */);
+  try {
+    shm_pool_ = std::make_unique<SharedMemory>(
+        shm_region_name, shm_default_size, shm_growth_size,
+        true /* truncate */);
+  }
+  catch (const PythonBackendException& pb_exception) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL, pb_exception.what());
+  }
 
   // Child Mutex and CV
   pthread_mutex_t* child_mutex;
@@ -802,7 +809,7 @@ ModelInstanceState::SetupChildProcess()
   }
 
   // Path to the extracted Python env
-  std::string python_execution_env = "none";
+  std::string python_execution_env = "";
 
   // Path to environment activation script
   std::string path_to_activate;
@@ -811,7 +818,7 @@ ModelInstanceState::SetupChildProcess()
   if (model_state->PythonExecutionEnv() != "") {
     try {
       python_execution_env =
-          model_state->StateForBackend()->env_manager->Extract(
+          model_state->StateForBackend()->env_manager->ExtractIfNotExists(
               model_state->PythonExecutionEnv());
     }
     catch (PythonBackendException& pb_exception) {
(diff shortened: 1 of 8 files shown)


src/python.cc
```diff
@@ -798,6 +801,35 @@ ModelInstanceState::SetupChildProcess()
             .c_str());
   }
 
+  // Path to the extracted Python env
+  std::string python_execution_env = "none";
```
deadeyegoodwin, line 805: Why not just leave the string empty to indicate it is not set/intialized?

src/python.cc
```diff
@@ -798,6 +801,35 @@ ModelInstanceState::SetupChildProcess()
...
+
+  // Path to environment activation script
+  std::string path_to_activate;
+  std::string path_to_libpython;
+
+  if (model_state->PythonExecutionEnv() != "") {
+    try {
+      python_execution_env =
```
deadeyegoodwin, line 813: We shouldn't need a copy of the execution environment for each instance, or even each model. If multiple models/instances use the same execution environment they should all use the same extracted version of the env. So it seems like the execution environments should be managed at the backend level.

deadeyegoodwin (changes requested):

src/python.cc
Tabrizian: @deadeyegoodwin We do not copy the execution environment for each instance. The `Extract` function will extract the environment only if it does not exist in the `env_map_`. Otherwise, it will return the path to the previously extracted environment. Perhaps I should rename the `Extract` function to `ExtractIfNotExists` so that it is more clear.

src/python.cc
Tabrizian: I fixed this.

src/python.cc
Tabrizian: Improved the comments and function name to address this.

src/python.cc
```diff
@@ -745,8 +745,15 @@ ModelInstanceState::SetupChildProcess()
...
-  shm_pool_ = std::make_unique<SharedMemory>(
-      shm_region_name, shm_default_size, shm_growth_size, true /* truncate */);
+  try {
+    shm_pool_ = std::make_unique<SharedMemory>(
+        shm_region_name, shm_default_size, shm_growth_size,
+        true /* truncate */);
+  }
+  catch (const PythonBackendException& pb_exception) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL, pb_exception.what());
```
CoderHam, line 755: How about:
Unable to create shared memory region + "shm_region_name" + ": " + pb_exception.what()

src/python.cc
CoderHam: ExtractIfNotExists feels like it could be misinterpreted. How about ExtractIfNotExtracted or similar?

CoderHam (changes requested):

src/python.cc
Tabrizian: @CoderHam  all the returned exceptions have `shm_region_name` in them so I think it would be redundant.

src/python.cc
Tabrizian: Sounds good, I'll use `ExtractIfNotExtracted `.

src/python.cc
CoderHam: Oh yes, I missed that you were mentioning it there.

deadeyegoodwin (dismissed):
This seems ok but need separate tickets and PRs for the documentation and testing for both custom stub and execution env.

Tabrizian:
@deadeyegoodwin I have added testing for both in this PR: https://github.com/triton-inference-server/server/pull/2897

This tests two Python versions (3.6 and 3.9) and compiles a custom Python stub to make sure it works properly. I filed a ticket for documentation changes.

deadeyegoodwin (approved):

CoderHam (approved):

PR #57, opened 2021-05-26
Add docs for custom execution environments

Tabrizian:


README.md
```diff
@@ -287,20 +299,125 @@ models
     └── config.pbtxt
 ```
 
-## Using Your Own Python Interpreter
+## Using Custom Python Execution Environments
 
 Python backend shipped in the [NVIDIA GPU Cloud](https://ngc.nvidia.com/)
 containers uses Python 3.8. Changing your environment using
```
deadeyegoodwin, line 305: "uses Python 3,8". If your Python model is compatible with Python 3.8 and requires only modules already included in the Triton container, then you can skip this section. If you need to use a different version of Python or if you have additional dependencies, you need to recompile the stub executable and create an execution environment as described below and include that with your model."

I would delete "Changing your..." to the end of the section.

README.md
```diff
@@ -287,20 +299,125 @@ models
...
+
+parameters: {
+  key: "EXECUTION_ENV_PATH",
+  value: {string_value: "/home/iman/miniconda3/envs/python-3-6/python3.6.tar.gz"}
+}
+```
+
+### Important Notes
+
+1. The version of the Python interpreter in the tar file must match the version of triton_python_backend_stub.
```
deadeyegoodwin, line 403: "... in the execution environment must match ..."

README.md
```diff
@@ -287,20 +299,125 @@ models
...
+[Building Custom Python Backend Stub Step](#1-building-custom-python-backend-stub).
+In this case you only need to pack your environment using `conda-pack` and
+provide the path to tar file in the model config. However, the previous note
+still applies here and the version of the Python interpreter inside the conda
+environment must match the Python version of stub used by Python backend. The
+default version of the stub is Python 3.8.
+
+3. Note that if the Python version of your environment matches the Python
+version available inside the container, the packages installed in
+[conda](https://conda.io) or [virtualenv](https://virtualenv.pypa.io/) will be
```
deadeyegoodwin, line 415: Don't mention virtualenv

README.md
```diff
@@ -287,20 +299,125 @@ models
...
+
+2. If you don't want to use a different Python interpreter, you can skip
+[Building Custom Python Backend Stub Step](#1-building-custom-python-backend-stub).
+In this case you only need to pack your environment using `conda-pack` and
+provide the path to tar file in the model config. However, the previous note
+still applies here and the version of the Python interpreter inside the conda
+environment must match the Python version of stub used by Python backend. The
+default version of the stub is Python 3.8.
+
+3. Note that if the Python version of your environment matches the Python
```
deadeyegoodwin, line 413: I don't understand this point.

README.md
```diff
@@ -287,20 +299,125 @@ models
...
+version available inside the container, the packages installed in
+[conda](https://conda.io) or [virtualenv](https://virtualenv.pypa.io/) will be
+picked up by Python backend and applied to all the Python models. In this case
+you can skip both [Building Custom Python Backend
+Stub](#1-building-custom-python-backend-stub) and [Packaging the Conda
+Environment](#2-packaging-the-conda-environment) steps. You only need to make
+sure that the environment is activated before you start the Triton Server.
```
deadeyegoodwin, line 421: Add a note describing that the same env can be used for multiple models by having the models point to the same tar file.

deadeyegoodwin (changes requested):

README.md
Tabrizian: This point says that if you install packages in the triton container or a conda environment with a matching Python interpreter version, you don't need to use conda-pack or the `EXECUTION_ENV_PATH`. If you activate the environment before the Python backend starts it will work properly.

deadeyegoodwin: I would just remove the whole bullet. It is very confusing and there is no need to us to describe how they can change the contents of the triton container itself.

README.md
```diff
@@ -287,20 +299,132 @@ models
...
+
+3. Note that if the Python version of your environment matches the Python
+version available inside the container, the packages installed in
+[conda](https://conda.io) will be picked up by Python backend and applied to all
+the Python models. In this case you can skip both [Building Custom Python
+Backend Stub](#1-building-custom-python-backend-stub) and [Packaging the Conda
+Environment](#2-packaging-the-conda-environment) steps. You only need to make
+sure that the environment is activated before you start the Triton Server.
+
+4. You can use a single execution environment for multiple models. You need to
```
deadeyegoodwin, line 425: "You can share an execution environment across multiple models."

deadeyegoodwin (changes requested):

README.md
Tabrizian: Removed.

deadeyegoodwin (approved):

PR #58, opened 2021-05-28
Add automatic restart for Python backend stubs

Tabrizian:
This should be merged after #59.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,13 @@ FetchContent_MakeAvailable(pybind11)
 find_package(ZLIB REQUIRED)
 find_package(Threads REQUIRED)
 
+#
+# Boost
+#
+find_package(Boost REQUIRED)
+message(STATUS "Using Boost ${Boost_VERSION}")
+include_directories(${Boost_INCLUDE_DIRS})
+
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 configure_file(src/libtriton_python.ldscript libtriton_python.ldscript COPYONLY)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -209,7 +209,6 @@ install(
   DESTINATION
     ${CMAKE_INSTALL_PREFIX}/backends/python
 )
-
 
 include(CMakePackageConfigHelpers)
 configure_package_config_file(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,10 @@ set(TRITON_CORE_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/co
 #
 include(FetchContent)
 
+# We need to use ExternalProject because we want to use Boost 1.76 which is not
+# available in Ubuntu 20.04
+include(ExternalProject)
+
 FetchContent_Declare(
   repo-common
   GIT_REPOSITORY https://github.com/triton-inference-server/common.git
@@ -80,17 +84,25 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(pybind11)
 
-find_package(ZLIB REQUIRED)
-find_package(Threads REQUIRED)
-
 #
 # Boost
 #
-find_package(Boost REQUIRED)
-message(STATUS "Using Boost ${Boost_VERSION}")
-include_directories(${Boost_INCLUDE_DIRS})
+ExternalProject_Add(
+  boostorg
+  URL https://boostorg.jfrog.io/artifactory/main/release/1.76.0/source/boost_1_76_0.tar.gz
+  URL_HASH SHA256=7bd7ddceec1a1dfdcbdb3e609b60d01739c38390a5f956385a12f3122049f0ca
+  PREFIX "boost-src"
+  CONFIGURE_COMMAND ${CMAKE_COMMAND} -E copy_directory
+                    <SOURCE_DIR>/boost/ ${CMAKE_BINARY_DIR}/boost
+  INSTALL_COMMAND ""
+  BUILD_COMMAND ""
+)
+set(boostorg_INCLUDE_DIRS "${CMAKE_BINARY_DIR}/boost/")
 
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
+find_package(ZLIB REQUIRED)
+find_package(Threads REQUIRED)
+
+include_directories(${CMAKE_BINARY_DIR})
 configure_file(src/libtriton_python.ldscript libtriton_python.ldscript COPYONLY)
 
 add_library(
@@ -112,16 +124,14 @@ add_executable(
   src/shm_manager.cc
   src/shm_manager.h
 )
+
+add_dependencies(triton-python-backend boostorg)
+add_dependencies(triton-python-backend-stub boostorg)
+
 set_property(TARGET triton-python-backend-stub PROPERTY OUTPUT_NAME triton_python_backend_stub)
 
 add_library(
   TritonPythonBackend::triton-python-backend ALIAS triton-python-backend
-)
-
-target_include_directories(
-  triton-python-backend
-  PRIVATE
-    ${CMAKE_CURRENT_SOURCE_DIR}/src
 )
 
 target_compile_features(triton-python-backend PRIVATE cxx_std_11)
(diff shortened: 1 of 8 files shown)


src/pb_stub.cc
```diff
@@ -132,6 +132,7 @@ class Stub {
   py::object deserialize_bytes_;
   py::object serialize_bytes_;
   ResponseBatch* response_batch_;
+  bool* health_;
```
CoderHam, line 135: Is this used?

src/pb_stub.cc
```diff
@@ -712,13 +715,15 @@ main(int argc, char** argv)
...
   bool non_graceful_exit = false;
+  bool &health = stub->Health();
 
   bool background_thread_running = true;
   std::thread background_thread(
-      [&parent_pid, &background_thread_running, &stub, &non_graceful_exit] {
+      [&parent_pid, &background_thread_running, &stub, &non_graceful_exit, &health] {
         while (background_thread_running) {
-          // Every two seconds check if the parent process is alive.
+          // Every 10ms check if the parent process is alive.
```
CoderHam, line 724: How does moving from 2s to 10ms impact performance?

src/python.cc
```diff
@@ -208,21 +208,24 @@ class ModelInstanceState : public BackendModelInstance {
...
   IPCMessage* ipc_message_;
   std::unique_ptr<SharedMemory> shm_pool_;
-  // Child process pid
-  pid_t child_pid_;
+  // Stub process pid
+  pid_t stub_pid_;
 
   // Parent process pid
   pid_t parent_pid_;
-
```
CoderHam, line 223: Keep this empty line

src/python.cc
```diff
@@ -208,21 +208,24 @@ class ModelInstanceState : public BackendModelInstance {
...
+  pthread_mutex_t* stub_mutex_;
+  pthread_cond_t* stub_cond_;
   pthread_mutex_t* parent_mutex_;
   pthread_cond_t* parent_cond_;
   std::string model_path_;
   IPCMessage* ipc_message_;
   std::unique_ptr<SharedMemory> shm_pool_;
-  // Child process pid
-  pid_t child_pid_;
+  // Stub process pid
```
CoderHam, line 218: Add empty line before comment line

CoderHam (changes requested):

src/python.cc
```diff
@@ -956,32 +950,147 @@ ModelInstanceState::SetupChildProcess()
...
-    pthread_cond_wait(parent_cond_, parent_mutex_);
-    pthread_mutex_unlock(parent_mutex_);
+    ipc_message_->health = false;
+
+    // Sleep 1 second so that the child process has a chance to change the
+    // health variable
+    sleep(1);
+
+    if (ipc_message_->health) {
+      // Create Python inference request
```
deadeyegoodwin, line 1074: Why are you sending a request to the stub at this point? Is this how you tell the stub to exit? Why not a more direct way... at the very least this comment needs updating.

src/pb_stub.cc
deadeyegoodwin, line 724: Does it really need to reset the health every 10ms? That seems unnecessarily frequent.

src/pb_stub.cc
```diff
@@ -712,13 +715,15 @@ main(int argc, char** argv)
...
           sleep(0.01);
+          health = true;
```
deadeyegoodwin, line 726: You are communicating between threads using "plain" variables without mutexes or atomics. This one is in shared memory and 'background_thread_running' and others are on stack. There is no guarantee if/when other threads will see this changes. On x86 it might just work but that may not be true on ARM. You need to make there variables atomics or protect them with a mutex or some other sync that enforces memory ordering/visibility.
https://stackoverflow.com/questions/3114124/in-c-c-are-volatile-variables-guaranteed-to-have-eventually-consistent-semant
https://stackoverflow.com/questions/34382440/shared-variables-in-c11

deadeyegoodwin (changes requested):

src/python.cc
Tabrizian: It is sending a request to trigger the shutdown and calling of `finalize` on the Python stub side. I agree that the messaging mechanism needs an update/refactor. When the main process creates a request of size zero, it triggers the shutdown on the stub side. I'll update the comment to be more clear.

src/pb_stub.cc
Tabrizian: Yes, that is not really required. I'll increase it to 300ms so that it has 3 chances to set this variable when the parent process is polling this variable (parent process polls this variable every 1s).

src/pb_stub.cc
Tabrizian: I see, I will fix this. I will add a mutex only in the shared memory for the health variable and make the variables that are shared between the threads atomic.

src/pb_stub.cc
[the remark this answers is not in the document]
Tabrizian: My bad, it is not used. I'll remove it.

src/python.cc
Tabrizian: I think we should only have empty lines before the comments so I think the removal is ok here.

Tabrizian:
@deadeyegoodwin I decided to switch to boost for using mutexes because of better abstractions. Boost provides `scoped_lock` in shared memory. I added a `health_mutex` which is used for synchronizing the `health` variable which resides in shared memory. If locking on the `health_mutex` fails, we assume that the stub is not responsive and we'll kill the stub process. I also looked at other alternatives such as [gcc atomic builtins](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html) and using C++11 atomic variables inside the shared memory. The latter can be used if the implementation of the atomic in the toolchain is lock-free. The implementation is lock-free on some toolchains and architectures, however, it may use locks on some other architectures.  The problem with  [gcc atomic builtins](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html) is similar to using atomics in shared memory. These functions only work on Intel processors which makes them platform dependent and unsuitable.

deadeyegoodwin (approved):

deadeyegoodwin (approved):

CMakeLists.txt
```diff
@@ -193,6 +210,7 @@ install(
     ${CMAKE_INSTALL_PREFIX}/backends/python
 )
 
+
```
CoderHam, line 213: remove extra line

src/pb_stub.cc
```diff
@@ -38,12 +43,13 @@
 #include "pb_utils.h"
 #include "shm_manager.h"
 
-
 namespace py = pybind11;
 using namespace pybind11::literals;
 
 namespace triton { namespace backend { namespace python {
 
+namespace bi = boost::interprocess;
```
CoderHam, line 51: move this together with the other namespace definitions

CoderHam (changes requested):

CMakeLists.txt
```diff
@@ -80,10 +84,25 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(pybind11)
 
+#
+# Boost
+#
+ExternalProject_Add(
```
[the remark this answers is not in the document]
Tabrizian: Removed.

CMakeLists.txt
```diff
@@ -52,6 +52,10 @@ set(TRITON_CORE_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/co
 #
```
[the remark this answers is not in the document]
Tabrizian: Fixed.

src/shm_manager.h
```diff
@@ -47,13 +49,15 @@ class SharedMemory {
...
   int64_t shm_growth_bytes_;
 
-  // List of old shared memory addresses that should be deallocated.
-  // First element of the pair is size and second element is the address.
-  std::vector<std::pair<size_t, char*>> old_shm_addresses_;
   void UpdateSharedMemory();
 
   // Get the amount of shared memory available.
   size_t GetAvailableSharedMemory();
+  boost::interprocess::shared_memory_object shm_obj_;
```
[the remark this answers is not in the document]
Tabrizian: It doesn't have a comment so I don't think blank line is required.

src/pb_stub.cc
```diff
@@ -711,15 +743,19 @@ main(int argc, char** argv)
   py::scoped_interpreter guard{};
 
   stub->Initialize(model_version, argv[6] /* triton install path */);
-  bool non_graceful_exit = false;
+  std::atomic<bool> non_graceful_exit(false);
 
-  bool background_thread_running = true;
+  std::atomic<bool> background_thread_running = {true};
```
CoderHam, line 748: More of a cleanliness fix but I think you should stick to the same convention for initializing these atomic bool variables ie. () or {} but not both.

CoderHam (changes requested):
Small fix, otherwise LGTM

deadeyegoodwin (approved):

CoderHam (approved):

PR #62, opened 2021-06-18
Add support for DLPack Tensors

Tabrizian:
- [x] Support for input Tensors in DLPack
- [x] Support output Tensors in DLPack
- [x] Fix TYPE_BYTES support in GPU (DLPack does not support GPU tensors so I ended up not supporting this.)
- [ ] Add support for CUDA Array Interface (It will be addressed in a future PR)
- [x] Add testing for DLPack and CUDA Array Interface (I will add another PR for testing on the main repo)

Opening this PR for reviews. I will add testing to the server repo for this PR. The interfaces below show some examples of using Python backend tensors:

```python
# Create tensors using DLPack
tensor = pb_utils.Tensor.from_dlpack("TENSOR_NAME", dlpack_capsule_object)

# If the tensor is stored in CPU, it is possible to get NumPy representation
numpy_representation = tensor.as_numpy()

# Convert Python backend tensor to DLPack
tensor_dlpack = tensor.to_dlpack()
pytorch_tensor = torch.utils.dlpack.from_dlpack(tensor_dlpack)

# Legacy interfaces using NumPy
pb_tensor_from_numpy = pb_utils.Tensor("TENSOR_NAME", np.array([1, 2, 3]))
```

One issue that is remaining is if the output tensor is already an exported pointer, we cannot get the cudaIpcMemHandle. The only way that I can think of to mitigate this issue is to copy the tensor in another buffer and export the pointer for the new buffer. I'll address this in a future PR.


src/pb_stub_utils.cc
```diff
@@ -0,0 +1,281 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "pb_stub_utils.h"
+#include "pb_utils.h"
+
+namespace triton { namespace backend { namespace python {
+
+int
+numpy_to_triton_type(py::object data_type)
```
deadeyegoodwin, line 33: Why is data_type an int instead of some enum, and why are plain numbers used for the datatypes instead of sum enum values?

src/pb_tensor.cc
```diff
@@ -0,0 +1,301 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "pb_stub_utils.h"
+#include "pb_tensor.h"
+#include "pb_utils.h"
+
+namespace py = pybind11;
+
+namespace triton { namespace backend { namespace python {
+PbTensor::PbTensor(std::string name, py::object numpy_array)
```
deadeyegoodwin, line 34: const std::string& name
Can you padd numpy_array by const ref?

Similar comment for below functions

src/pb_tensor.h
```diff
@@ -0,0 +1,146 @@
...
+typedef enum PYTHONBACKEND_tensortype_enum {
+  PYTHONBACKEND_RAW,
+  PYTHONBACKEND_NUMPY,
+  PYTHONBACKEND_DLPACK
+} PYTHONBACKEND_TensorType;
+
+// PbTensor class is the representation of Triton tensors
+// inside Python backend.
+class PbTensor {
+  std::string name_;
```
deadeyegoodwin, line 49: class members should be private ?

src/python.cc
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
deadeyegoodwin, line 1: Remove "(c)" from copyright to avoid line wrap. These wrapped lines will fail in the copyright checker

deadeyegoodwin (changes requested):

src/pb_tensor.h
Tabrizian: By default class members are private. I'll add `private` to be more explicit.

src/pb_stub_utils.cc
Tabrizian: I'll fix this to be an enum.

src/pb_tensor.cc
Tabrizian: Adding const ref for `numpy_array` leads to some compile errors. This doesn't copy the whole numpy_array. It only copies the reference so there should not be any performance cost.

deadeyegoodwin (approved):

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #64, opened 2021-07-08
Fix a bug in detecting the health of the stub process

Tabrizian:


src/pb_stub.cc
```diff
@@ -771,6 +769,8 @@ main(int argc, char** argv)
         }
       });
 
+  stub->Initialize(model_version, argv[6] /* triton install path */);
```
CoderHam, line 772: Add comment about why initialization needs to be delayed till here

CoderHam (changes requested):

src/pb_stub.cc
Tabrizian: Done

CoderHam (approved):

PR #65, opened 2021-07-09
Refactor communication of stub and main process

Tabrizian:


src/shm_manager.cc
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
deadeyegoodwin, line 1: Remove '(c)' to avoid line wrapping by formatter

deadeyegoodwin (changes requested):

src/shm_manager.cc
Tabrizian: Fixed.

deadeyegoodwin (approved):

PR #81, opened 2021-09-02
Add support for relative paths for execution envs

Tabrizian:
Closes triton-inference-server/server#3097 and triton-inference-server/server#3227. `$$TRITON_MODEL_DIRECTORY` can be used for pointing to the model folder in model repository.

### DIFF
--- a/src/python.cc
+++ b/src/python.cc
@@ -1086,6 +1086,18 @@ ModelInstanceState::StartStubProcess()
       python_backend_stub = model_python_backend_stub;
     }
 
+    // Give the execute permission to the owner.
+    int error = chmod(python_backend_stub.c_str(), S_IXUSR);
+    if (error != 0) {
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_INTERNAL,
+          (std::string("Failed to give execute permission to "
+                       "triton_python_backend_stub in ") +
+           python_backend_stub + " " + Name() +
+           " Error No.: " + std::to_string(error))
+              .c_str());
+    }
+
     std::string bash_argument;
 
     // This shared memory variable indicates whether the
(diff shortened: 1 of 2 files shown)


Tabrizian:
Closes triton-inference-server/server#3097.

Tabrizian:
`$$TRITON_MODEL_DIRECTORY` can be used for point to the model folder in model repository.

src/python.cc
```diff
@@ -1086,6 +1086,18 @@ ModelInstanceState::StartStubProcess()
       python_backend_stub = model_python_backend_stub;
     }
 
+    // Give the execute permission to the owner.
+    int error = chmod(python_backend_stub.c_str(), S_IXUSR);
```
deadeyegoodwin, line 1090: Depending on how the container is set up the user running Triton may not be able to chmod the file. And the file may already have execute permissions. In general it seems hard to determine if the stub will be executable or not by Triton except by trying it. Wouldn't it be better to just give an appropriate error when attempting to launch the stub if that fails due to inadequate permission?

deadeyegoodwin (changes requested):

src/python.cc
Tabrizian: @deadeyegoodwin The problem is that S3 does not retain permissions. When triton downloads model repository from S3, triton_python_backend_stub does not have execute permission. A user faced a similar issue [here](triton-inference-server/server#3227). https://www.reddit.com/r/aws/comments/d8lw4z/how_to_preserve_file_permissions_on_a_file_when/

deadeyegoodwin: But what if the stub is executable by Triton, but Triton does not have ownership to chmod the stub? Then the chmod call will fail but it is not even needed.

Tabrizian: @deadeyegoodwin I see. I can check whether stub is executable by Triton (using stat) and chmod only if it is not executable.

deadeyegoodwin: I believe stat will just tell you information about the owner/group/other executable permissions of the file. It doesn't tell you if triton itself will be able to run the stub. You would have to know which user (and group) triton was running as and then compare that to the owner and group of the file and the corresponding permissions. Perhaps there is a another API that more directly tells you if it is executable? Perhaps you just need try the chmod only after running the stub has failed?

Tabrizian: > Perhaps you just need try the chmod only after running the stub has failed?

I can do this. I was also looking into the Triton API to see if I can use the artifact type to detect where the model repository is stored and chmod only if it is not using the local file system. The problem with this is that artifact type enum has only a single value which is TRITONBACKEND_ARTIFACT_FILESYSTEM: https://github.com/triton-inference-server/core/blob/e5bde79031d6557ee770d51cb51ed2df5083c6e0/include/triton/core/tritonbackend.h#L119

Tabrizian: Fixed in the next commit. First, I use `system` to make sure that stub can be executed and then use `execvp`. I only chmod when `system` returns unexpected return code.

deadeyegoodwin (approved):

oeway:
Hi, thanks for adding this variable. I am wondering how one can use this to share custom environments across models (as mentioned in the README). I have many models share the same env and each package is 3.5GB hosted on S3.

(Sorry if I posted the question in the wrong place)

oeway:
@Tabrizian Thanks for the tips! I will try it when the new docker image is available.

PR #87, opened 2021-10-04
Clarify FORCE_CPU_ONLY_INPUT_TENSORS documentation

Tabrizian:


README.md
```diff
@@ -713,6 +713,9 @@ model config, you are letting Triton decide the placement of input Tensors:
 parameters: { key: "FORCE_CPU_ONLY_INPUT_TENSORS" value: {string_value:"no"}}
 ```
 
+Using this configuration parameter, you **cannot** force your input tensors to
```
deadeyegoodwin, line 716: This is still not clear... it doesn't describe what happens with FORCE is false. See above comment... you might want to remove this.

README.md
```diff
@@ -698,7 +698,7 @@ is not C-order contiguous an exception will be raised.
 
 This function can be used to check whether a tensor is placed in CPU or not.
 
-## Controlling Input Tensor Device Placement
+## Input Tensor Device Placement
 
 By default Python backend moves all the input tensors to CPU. Starting from
 21.09 release, you can control whether you want to move input tensors to CPU or
```
deadeyegoodwin, line 704: Instead of the new paragraph below, I would just change this text...
"By default, the Python backend moves all input tensors to CPU before providing them to the python model. Starting from 21.09 you can change this default behavior. By setting ... to "no", Triton will not move input tensors to CPU for the model. Instead, Triton will provide the input tensors to the python model in either CPU or GPU memory, depending on how those tensors were last used. You cannot predict which memory will be used for each input tensor so your python model must be able to handle tensors in both CPU and GPU memory..."

Something like that.

deadeyegoodwin (changes requested):

README.md
Tabrizian: Thanks for the feedback. Modified the paragraph according to your comment.

deadeyegoodwin (approved):

PR #92, opened 2021-11-04
Update inferentia build scripts to new Neuron SDK and added TF build support

jbkyang-nvi:
Update inferentia build scripts to new Neuron SDK and added TF build support


inferentia/README.md
```diff
@@ -113,6 +124,8 @@ Now, the server can be launched with the model as below:
 tritonserver --model-repository <path_to_model_repository>
 ```
 
-Note: The `config.pbtxt` and `model.py` should be treated as
+Note: 
+1. The `config.pbtxt` and `model.py` should be treated as
 starting point. The users can customize these files as per
 their need.
+2. Triton Inferentia currently only works with **single** models.
```
tanmayv25, line 131: **single** model.

inferentia/README.md
```diff
@@ -47,33 +47,44 @@ After logging into the inf1* instance, you will need to clone
...
+```
+ sudo ./python_backend/setup-pre-container.sh
+```
+
 Then, start the Triton instance with:
 ``` 
-docker run -v /home/ubuntu/python_backend:/home/ubuntu/python_backend -v /lib/udev:/mylib/udev -v /run:/myrun --shm-size=1g --ulimit memlock=-1 -p 8000:8000 -p 8001:8001 -p 8002:8002 --ulimit stack=67108864 -ti nvcr.io/nvidia/tritonserver:<xx.yy>-py3
+docker run --device /dev/neuron0 <more neuron devices> -v /home/ubuntu/python_backend:/home/ubuntu/python_backend -v /lib/udev:/mylib/udev --shm-size=1g -e "AWS_NEURON_VISIBLE_DEVICES=ALL" --ulimit memlock=-1 -p 8000:8000 -p 8001:8001 -p 8002:8002 --ulimit stack=67108864 -ti nvcr.io/nvidia/tritonserver:<xx.yy>-py3
+```
+Note 1: The user would need to list any new neuron device to run during container initialization.
```
tanmayv25, line 60: What do you mean by **new** neuron device?

inferentia/README.md
```diff
@@ -47,33 +47,44 @@ After logging into the inf1* instance, you will need to clone
...
+For example, to use 4 neuron devices on an instance, the user would need to run with:
```
tanmayv25, line 61: Isn't there a way to launch container with an access to all Neuron cores?

inferentia/scripts/setup.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#/bin/bash
```
tanmayv25, line 1: Why are you removing the shebang? A typo?

inferentia/scripts/setup.sh
```diff
@@ -28,28 +28,31 @@
 USAGE="
 usage: test_script.sh [options]
 
-Sets up enviroment for neuron-pytorch
+Sets up enviroment for Inferentia-neuron
```
tanmayv25, line 31: Sets up python execution environment for AWS Neuron SDK for execution on Inferentia chips.

inferentia/scripts/setup.sh
```diff
@@ -106,36 +127,32 @@ rm -rf build && mkdir build && cd build
...
+    pip install --upgrade tensorflow-neuron==1.15.5.* neuron-cc
+    # Update Neuron TensorBoard
+    pip install --upgrade tensorboard-plugin-neuron
+fi
+
+if [ $USE_PYTORCH -eq 0 ]
+then
+    conda install torch-neuron torchvision -y
+    # Upgrade torch-neuron and install transformers
+    pip install --upgrade torch-neuron neuron-cc[tensorflow] torchvision "transformers==4.6.0"
```
tanmayv25, line 149: Why `neuron-cc[tensorflow]` for `$USE_PYTORCH`?

inferentia/scripts/setup.sh
```diff
@@ -106,36 +127,32 @@ rm -rf build && mkdir build && cd build
...
-# https://aws.amazon.com/blogs/developer/neuron-conda-packages-eol/
-pip config set global.extra-index-url https://pip.repos.neuron.amazonaws.com
-pip install --upgrade torch-neuron torchvision "transformers==4.6.0"
+if [ $USE_TENSORFLOW -eq 0 ]
+then
+    conda install tensorflow-neuron pillow -y
+    #Update Neuron TensorFlow
+    pip install --upgrade tensorflow-neuron==1.15.5.* neuron-cc
+    # Update Neuron TensorBoard
+    pip install --upgrade tensorboard-plugin-neuron
```
tanmayv25, line 142: Is tensorboard really required?

inferentia/scripts/setup.sh
```diff
@@ -64,11 +67,30 @@ for OPTS; do
         echo "Inferentia path set to ${INFRENTIA_PATH}"
         shift 2
         ;;
+        -t|--use-tensorflow)
+        USE_TENSORFLOW=0
+        echo "Installing tensorflow-neuron"
+        shift 1
+        ;;
+        -p|--use-pytorch)
+        USE_PYTORCH=0
```
tanmayv25, line 76: Why using 0 as the value to indicate using pytorch runtime? Usually we stick with the global truth values of 0 meaning false and 1 meaning true..

Same for USE_TENSORFLOW above.

inferentia/scripts/setup.sh
```diff
@@ -64,11 +67,30 @@ for OPTS; do
...
+        echo "Installing pytorch-neuron"
+        shift 1
+        ;;
     esac
 done
+
+if [ $USE_TENSORFLOW -ne 0 ] && [ $USE_PYTORCH -ne 0 ]
```
tanmayv25, line 83: See my comments above. We must keep 0 as false and 1 as true.

inferentia/scripts/setup.sh
```diff
@@ -106,36 +127,32 @@ rm -rf build && mkdir build && cd build
...
+    conda install tensorflow-neuron pillow -y
+    #Update Neuron TensorFlow
+    pip install --upgrade tensorflow-neuron==1.15.5.* neuron-cc
+    # Update Neuron TensorBoard
+    pip install --upgrade tensorboard-plugin-neuron
+fi
+
+if [ $USE_PYTORCH -eq 0 ]
+then
+    conda install torch-neuron torchvision -y
```
tanmayv25, line 147: I guess these dependency comes from the demos from AWS Neuron SDK? There might be some more dependencies that a user might need to successfully deploy their models. 
Just something to keep in mind for future.

tanmayv25 (changes requested):

inferentia/README.md
jbkyang-nvi: No not right now.

inferentia/scripts/setup.sh
jbkyang-nvi: This is part of their installation instructions so I kept it

inferentia/scripts/setup.sh
jbkyang-nvi: Also part of their official installation instructions https://awsdocs-neuron.readthedocs-hosted.com/en/latest/neuron-intro/tensorflow-setup/tensorflow-install.html

inferentia/scripts/setup.sh
jbkyang-nvi: I was following the bash success/no success where anything but 0 means no-success. I'll change it

inferentia/scripts/setup.sh
jbkyang-nvi: same as above. I think it might be something necessary to run the examples. However these are the minimum set of instructions for installation in their instructions. What else do you think the user would need? I can add them now

tanmayv25: I don't think we might tell that right now. It depends really on their model right. I was wondering for making them user configurable in the script. But I guess for now let's just focus on mlperf models. Take a note of this however for future.

inferentia/scripts/setup.sh
tanmayv25: This is strange... I am fine keeping it for now.

inferentia/scripts/setup.sh
tanmayv25: Those are for exit codes for the functions..For truth values it is better to stick with 0(false) and 1(true)

tanmayv25 (approved):

PR #96, opened 2021-11-18
Add testing setup script and some more notes to README

jbkyang-nvi:
- Add test.sh setup and run script to inferentia/qa folder. Corresponding `server` PR: https://github.com/triton-inference-server/server/pull/3586
- Add to README: 1) pointers to how to compile a model 2) instructions on how to run test


inferentia/README.md
```diff
@@ -87,6 +100,18 @@ you can run:
 source /home/ubuntu/python_backend/inferentia/scripts/setup.sh -v 3.6
 ```
 Please use the `-h` or `--help` options to learn about more configurable options.
+## Compile Inferentia Model
```
tanmayv25, line 103: This can be removed as compiling the model for inferentia is a part of section [Setting up the Inferentia model](setting-up-the-inferentia-model). See below. For TF, we don't have `*.pt` or `*.pb` models, but have TF savedmodel. If you have something to add, then you can add in that section.

inferentia/README.md
```diff
@@ -202,4 +227,22 @@ Note:
...
\ No newline at end of file
+2. Triton Inferentia is currently tested with a **single** model. 
+
+## Testing Inferentia Setup for Accuracy
+The [qa folder](https://github.com/triton-inference-server/python_backend/tree/main/inferentia/qa)
+contains the necessary files to set up testing with a simple add_sub model. The test
+requires an instance with more than 8 inferentia cores to run, eg:`inf1.6xlarge`.
+start the test, run 
+```
+source <triton path>/python_backend/inferentia/qa/setup_test_enviroment_and_test.sh
```
tanmayv25, line 238: Let's use a consistent style. Following the ReadMe in the repo https://github.com/triton-inference-server/python_backend#quick-start

All the shell command must be preceded by $.

inferentia/README.md
```diff
@@ -202,4 +227,22 @@ Note:
...
+``` 
+where `<triton path>` is usually `/home/ubuntu`/.
+This script will pull the corresponding [server](https://github.com/triton-inference-server/server)
```
tanmayv25, line 241: What does corresponding server mean here? I think by default it picks up 'main' right?
This can be changed by exporting TRITON_SERVER_BRANCH_NAME which you describe below.
Just remove the 'corresponding', and describe in detail in the note.

inferentia/qa/Dockerfile.QA
```diff
@@ -0,0 +1,83 @@
+
+# Copyright 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tanmayv25, line 2: Fix the copytight.. and remove the extra line.

inferentia/README.md
```diff
@@ -202,4 +227,22 @@ Note:
...
+that contains the tests for inferentia. It will then build the most recent 
+Triton Server and Triton SDK.
```
tanmayv25, line 243: I do understand we are using perf_analyzer for output validation for some models. I believe we must also enable L0_infer in for inferentia models as it would test along all the range of datatypes and must take lesser time than perf_analyzer tools.

inferentia/qa/setup_test_enviroment_and_test.sh
```diff
@@ -0,0 +1,134 @@
...
+
+# Build QA container
+docker build -t ${QA_IMAGE} \
+                   -f ${TRITON_PATH}/python_backend/inferentia/qa/Dockerfile.QA \
+                   --build-arg "TRITON_PATH=${TRITON_PATH}" \
+                   --build-arg "BASE_IMAGE=${BASE_IMAGE}"   \
+                   --build-arg "BUILD_IMAGE=${BUILD_IMAGE}" \
+                   --build-arg "SDK_IMAGE=${SDK_IMAGE}"     .
+
+export TEST_JSON_REPO=/opt/tritonserver/qa/common/inferentia_perf_analyzer_input_data_json
```
tanmayv25, line 96: Can't we run the single and multiple instance cases in single test script?

tanmayv25 (changes requested):

inferentia/qa/setup_test_enviroment_and_test.sh
jbkyang-nvi: In the future, we probably want to add more tests for multiple instances that focus on how compute is distributed so it _might_ be cleaner to just have two separate test folders. If the user doesn't care about multiple instances, this way it's also easier to comment out the second test.

inferentia/README.md
jbkyang-nvi: since it is going to take more work to set up each model with gen script, would you be ok with adding this to a separate PR?

inferentia/qa/setup_test_enviroment_and_test.sh
jbkyang-nvi: I think this might be over optimizing. I'll combining the two for now

tanmayv25 (approved):

PR #97, opened 2021-11-19
pb_utils.Tensor.from_dlpack() — correct signature in README

okdimok:
pb_utils.Tensor.from_dlpack()  is implemented in https://github.com/triton-inference-server/python_backend/blob/main/src/pb_tensor.h#L135 and takes the tensor name as its first argument.

However, in the README it is shown as taking only a PyCapsule, this results in errors like
```
from_dlpack(): incompatible function arguments.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,7 +707,7 @@ class TritonPythonModel:
 
     # Create a Python backend tensor from the DLPack encoding of a PyTorch
     # tensor.
-    input0 = pb_utils.Tensor.from_dlpack(to_dlpack(pytorch_tensor))
+    input0 = pb_utils.Tensor.from_dlpack("INPUT0", to_dlpack(pytorch_tensor))
 ```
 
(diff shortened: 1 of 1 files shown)


tanmayv25:
@Tabrizian I wonder how the models work here:
https://github.com/triton-inference-server/server/blob/main/qa/python_models/dlpack_add_sub/model.py#L79-L80

The `from_dlpack()` is getting only a PyCapsule.

Tabrizian:
@tanmayv25 The lines that you have shared are from PyTorch DLPack not Python backend DLPack. PyTorch's "from_dlpack" only accepts a PyCapsule, but Python backend requires a name in addition to the PyCapsule.

Tabrizian (approved):

Tabrizian:
@okdimok Thanks for your contribution!

PR #99, opened 2021-11-24
Separate shared memory region by version

GuanLuo:


Tabrizian (approved):

src/python.cc
```diff
@@ -1308,7 +1308,7 @@ ModelInstanceState::SetupStubProcess()
 {
   std::string kind = TRITONSERVER_InstanceGroupKindString(kind_);
   std::string shm_region_name =
-      std::string("/") + Name() + "_" + kind + "_" + std::to_string(device_id_);
+      std::string("/") + Name() + "_" + std::to_string(Model()->Version()) + "_" + kind + "_" + std::to_string(device_id_);
```
Tabrizian, line 1311: Could you pass `shm_region_name` to `StartStubProcess` at the end of the function so that the region name is not built multiple times?

Tabrizian (approved):

PR #101, opened 2021-11-27
Increase the reference count when DLPack capsule is created

Tabrizian:


src/pb_tensor.cc
```diff
@@ -578,11 +598,14 @@ PbTensor::SaveToSharedMemory(
 
 #ifdef TRITON_ENABLE_GPU
 void
-PbTensor::LoadGPUData(
-    std::unique_ptr<SharedMemory>& shm_pool, std::mutex& gpu_load_mutex)
+PbTensor::LoadGPUData(std::unique_ptr<SharedMemory>& shm_pool)
 {
-  std::lock_guard<std::mutex> lock{gpu_load_mutex};
```
tanmayv25, line 584: Why is this not needed anymore?

Tabrizian: @tanmayv25 In the previous versions of cuda, cudaIpcOpenMemHandle required that no cudaIpcMemHandle is opened twice. This mutex was added to make sure that the previous handle has been closed before the new handle is opened. Looks like this is no longer an issue in CUDA 11.2+ and you can open a handle multiple times.

CoderHam: > Looks like this is no longer an issue in CUDA 11.2+ and you can open a handle belonging to the same allocation multiple times.
Good to know!

tanmayv25 (dismissed):

CoderHam (approved):

PR #102, opened 2021-11-30
Fix inferentia to support single output case

tanmayv25:


CoderHam (approved):
Do we have test coverage for this?

tanmayv25:
> Do we have test coverage for this?

Currently no. @jbkyang-nvi you should keep in mind to enable L0_infer_zero alongwith other tests from L0_infer whenever you get to it. That should test the single input/output case.

PR #105, opened 2021-12-07
Inferentia Tensorflow Test

jbkyang-nvi:
Add add_sub test for tensorflow. Test related changes are in the [server repo](https://github.com/triton-inference-server/server/pull/3665/). 
This introduced a couple of changes:
1. During model.py creation, if input/output is specified use it even if model type is Tensorflow
2. Using input sizes as multiples of neuron core size to address TF splicing issue (this will be fixed later).
3. Change all input for tests to be multi-dimensional `[8,4]` so that validation data json can be reused

### DIFF
--- a/inferentia/qa/setup_test_enviroment_and_test.sh
+++ b/inferentia/qa/setup_test_enviroment_and_test.sh
@@ -104,7 +104,7 @@ docker create --name ${CONTAINER_NAME}             \
             -e TEST_REPO=${TEST_REPO}              \
             -e TEST_JSON_REPO=${TEST_JSON_REPO}    \
             -e TRITON_PATH=${TRITON_PATH}          \
-            -e USE_TENSORFLOW="0"                  \
+            -e USE_PYTORCH="1"                     \
             --net host -ti ${QA_IMAGE}             \
             /bin/bash -c "bash -ex ${TEST_REPO}/${TEST_SCRIPT}" && \
             docker cp /lib/udev ${CONTAINER_NAME}:/mylib/udev && \
--- a/inferentia/qa/setup_test_enviroment_and_test.sh
+++ b/inferentia/qa/setup_test_enviroment_and_test.sh
@@ -47,7 +47,7 @@ CONTAINER_NAME="qa_container"
 
 
 cd ${TRITON_PATH}
-echo $TRITON_SERVER_REPO_TAG
+echo "Using server repo tag: $TRITON_SERVER_REPO_TAG"
 # Clone necessary branches
 rm -rf ${TRITON_PATH}/server
 git clone --single-branch --depth=1 -b ${TRITON_SERVER_REPO_TAG} \
--- a/inferentia/qa/setup_test_enviroment_and_test.sh
+++ b/inferentia/qa/setup_test_enviroment_and_test.sh
@@ -34,7 +34,7 @@ export TRITON_THIRD_PARTY_REPO_TAG=${DEFAULT_REPO_TAG}
 export IDENTITY_BACKEND_REPO_TAG=${DEFAULT_REPO_TAG}
 export PYTHON_BACKEND_REPO_TAG=${DEFAULT_REPO_TAG}
 export CHECKSUM_REPOAGENT_REPO_TAG=${DEFAULT_REPO_TAG}
-export TRITON_SERVER_BRANCH_NAME=${TRITON_SERVER_BRANCH_NAME:=${DEFAULT_REPO_TAG}}
+export TRITON_SERVER_REPO_TAG=${TRITON_SERVER_REPO_TAG:=${DEFAULT_REPO_TAG}}
 export TRITON_CLIENT_REPO_TAG=${TRITON_CLIENT_REPO_TAG:=${DEFAULT_REPO_TAG}}
 export TRITON_VERSION="2.17.0dev"
 export TRITON_CONTAINER_VERSION="21.12dev"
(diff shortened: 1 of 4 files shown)


inferentia/scripts/setup.sh
```diff
@@ -133,13 +133,16 @@ make triton-python-backend-stub -j16
 pip config set global.extra-index-url https://pip.repos.neuron.amazonaws.com
 conda config --env --add channels https://conda.repos.neuron.amazonaws.com
 
-if [ $USE_TENSORFLOW -eq 1 ]
-then
+if [ $USE_TENSORFLOW -eq 1 ] || [ $USE_TENSORFLOW -eq 2 ]; then
     conda install tensorflow-neuron pillow -y
```
tanmayv25, line 137: Don't we need to install the specific version of tensorflow-neuron here as well?

inferentia/qa/setup_test_enviroment_and_test.sh
```diff
@@ -109,6 +122,7 @@ docker create --name ${CONTAINER_NAME}             \
             -e TEST_REPO=${TEST_REPO}              \
             -e TEST_JSON_REPO=${TEST_JSON_REPO}    \
             -e TRITON_PATH=${TRITON_PATH}          \
+            -e USE_TENSORFLOW="1"                  \
```
tanmayv25, line 125: Don't we want to add test for TF version 2 as well?

jbkyang-nvi: No there's something that doesn't work for tf2. Since we don't need it for MLPerf right now I think we can just stick with testing with TF1

inferentia/scripts/setup.sh
jbkyang-nvi: No I think the --upgrade step will deal with the resolution

inferentia/scripts/setup.sh
```diff
@@ -34,11 +34,11 @@ Sets up python execution environment for AWS Neuron SDK for execution on Inferen
 -v|--python-version        Python version, default is 3.7
 -i|--inferentia-path       Inferentia path, default is: /home/ubuntu
 -p|--use-pytorch           Install pytorch-neuron if specified
--t|--use-tensorflow        Install tensorflow-neuron if specified
+-t|--use-tensorflow        Install tensorflow-neuron of specified version (1 or 2)
```
tanmayv25, line 37: A separate flag to set the TF version...

inferentia/scripts/setup.sh
```diff
@@ -34,18 +34,20 @@ Sets up python execution environment for AWS Neuron SDK for execution on Inferen
 -v|--python-version        Python version, default is 3.7
 -i|--inferentia-path       Inferentia path, default is: /home/ubuntu
 -p|--use-pytorch           Install pytorch-neuron if specified
--t|--use-tensorflow        Install tensorflow-neuron if specified
+-t|--use-tensorflow        Install tensorflow-neuron is specified
+--tensorflow-version       Version of Tensorflow used. Ignored if installing pytorch-neuron
```
tanmayv25, line 38: Should also mention the default version is 1.

tanmayv25 (approved):

inferentia/qa/setup_test_enviroment_and_test.sh
```diff
@@ -34,22 +34,24 @@ export TRITON_THIRD_PARTY_REPO_TAG=${DEFAULT_REPO_TAG}
...
 export BUILD_IMAGE=tritonserver_build
 export QA_IMAGE=tritonserver_qa
+export TEST_JSON_REPO=/opt/tritonserver/qa/common/inferentia_perf_analyzer_input_data_json
+export TEST_REPO=/opt/tritonserver/qa/L0_inferentia_perf_analyzer
+export TEST_SCRIPT="test.sh"
+CONTAINER_NAME="qa_container"
+
 
 cd ${TRITON_PATH}
+echo $TRITON_SERVER_REPO_TAG
```
CoderHam, line 50: Remove this

inferentia/scripts/setup.sh
```diff
@@ -34,18 +34,20 @@ Sets up python execution environment for AWS Neuron SDK for execution on Inferen
...
-OPTS=$(getopt -o hb:v:i:tp --long help,python-backend-path:,python-version:,inferentia-path:,use-tensorflow,use-pytorch -- "$@")
+OPTS=$(getopt -o hb:v:i:tp --long help,python-backend-path:,python-version:,inferentia-path:,use-tensorflow,use-pytorch,tensorflow-version: -- "$@")
 
 
 export INFRENTIA_PATH=${TRITON_PATH:="/home/ubuntu"}
 export PYTHON_BACKEND_PATH="/home/ubuntu/python_backend"
 export PYTHON_VERSION=3.7
 export USE_PYTORCH=0
 export USE_TENSORFLOW=0
+export TENSORFLOW_VERSION="1"
```
CoderHam, line 50: Why is this a string?

inferentia/scripts/setup.sh
```diff
@@ -133,13 +153,16 @@ make triton-python-backend-stub -j16
...
 
-if [ $USE_TENSORFLOW -eq 1 ]
-then
+if [ $USE_TENSORFLOW -eq 1]; then
     conda install tensorflow-neuron pillow -y
-    #Update Neuron TensorFlow
-    pip install --upgrade tensorflow-neuron==1.15.5.* neuron-cc
     # Update Neuron TensorBoard
     pip install --upgrade tensorboard-plugin-neuron
+    #Update Neuron TensorFlow
```
CoderHam, line 160: space after #

CoderHam (changes requested):

inferentia/scripts/setup.sh
jbkyang-nvi: We're checking against strings in the if statements

inferentia/qa/setup_test_enviroment_and_test.sh
jbkyang-nvi: I think it's helpful for the user to read the server repo tag.. I imagine it to be frequent that the user will need to input the server repo that's not "main"

CoderHam: It is not useful to just echo a branch name. Echo a more readable message such as `server_repo_tag: <branch>`

CoderHam (approved):

PR #107, opened 2021-12-17
Enable batching on inferentia and add tests

jbkyang-nvi:
Enabled batching for inferentia models, added tests for batching and fixed tensorflow bug where first dimension of inputs need to be multiples of num_threads_per_core  * core_per_instance. 
Also added extra error reporting in python backend stub
Testing PR: https://github.com/triton-inference-server/server/pull/3714

### DIFF
--- a/inferentia/README.md
+++ b/inferentia/README.md
@@ -39,7 +39,9 @@ and the [Neuron Runtime](https://awsdocs-neuron.readthedocs-hosted.com/en/latest
   - [Inferentia setup](#inferentia-setup)
   - [Setting up the Inferentia model](#setting-up-the-inferentia-model)
     - [PyTorch](#pytorch)
+      - [Batching with Pytorch](#batching-with-pytorch)
     - [TensorFlow](#tensorflow)
+      - [Batching with Tensorflow](#batching-with-tensorflow)
   - [Serving Inferentia model in Triton](#serving-inferentia-model-in-triton)
   - [Testing Inferentia Setup for Accuracy](#testing-inferentia-setup-for-accuracy)
 
@@ -156,6 +158,24 @@ loaded on cores 2-3. To best engage inferentia device, try setting
 the number of neuron cores to be a proper multiple of the instance
 count.
 
+#### Batching with Pytorch
+
+There are three flags related to batched inputs:
+
+1. `--max_batch_size`: If set to 0, then Triton server will send no batched
+   requests. Otherwise, Triton will send batched requests up to specified
+   `max_batch_size`.
+2. `--preferred_batch_size`: The preferred number of batches. To optimize
+   performance, this is recommended to be multiples of engaged neuron cores.
+   For example, if each instance is using 2 neuron cores, `preferred_batch_size`
+   could be 2, 4 or 6.
+3. `--disable_batch_requests_to_neuron`: Enable the non-default way for Triton to
+   handle batched requests. Triton backend will send each request to neuron
+   separately, irrespective of if the Triton server requests are batched.
+   This flag can be set with `max_batch_size` > 0, and is recommended when users
+   want to optimize performance with models that do not perform well with batching
+   without the flag.
+
 ### TensorFlow
 For TensorFlow, the model must be compiled for AWS Neuron. See
 [AWS Neuron TensorFlow](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/neuron-guide/neuron-frameworks/tensorflow-neuron/tutorials/index.html
@@ -188,6 +208,10 @@ neuron core in the chip. For TensorFlow, we use deprecated feature of
 next available Neuron cores and not specific ones. See
 [Parallel Execution using NEURONCORE_GROUP_SIZES](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/neuron-guide/appnotes/perf/parallel-ncgs.html?highlight=NEURONCORE_GROUP_SIZES)
 for more information.
+
+#### Batching with Tensorflow
+
+The batching flags have the same functionality for Tensorflow
 
 Please use the `-h` or `--help` options in `gen_triton_model.py` to
 learn about more configurable options.
--- a/inferentia/README.md
+++ b/inferentia/README.md
@@ -39,10 +39,9 @@ and the [Neuron Runtime](https://awsdocs-neuron.readthedocs-hosted.com/en/latest
   - [Inferentia setup](#inferentia-setup)
   - [Setting up the Inferentia model](#setting-up-the-inferentia-model)
     - [PyTorch](#pytorch)
-      - [Batching with Pytorch](#batching-with-pytorch)
     - [TensorFlow](#tensorflow)
-      - [Batching with Tensorflow](#batching-with-tensorflow)
   - [Serving Inferentia model in Triton](#serving-inferentia-model-in-triton)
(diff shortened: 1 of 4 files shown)


inferentia/scripts/gen_triton_model.py
```diff
@@ -336,7 +342,8 @@ def _validate_output_dict(self, expected_count):
     return init_impl
 
 
-def get_tensorflow_execute_impl():
+def get_tensorflow_execute_impl(enable_batching,
```
tanmayv25, line 345: Same here.. I don't think you need to pass enable_batching flag here. Just look whether execute function was called with single request or multiple.

tanmayv25 (changes requested):

src/pb_stub.cc
```diff
@@ -494,9 +494,12 @@ Stub::Execute(RequestBatch* request_batch, ResponseBatch* response_batch)
   // If the number of request objects do not match the number of resposne
   // objects throw an error.
   if (response_size != batch_size) {
-    throw PythonBackendException(
+    std::string err =
         "Number of InferenceResponse objects do not match the number of "
-        "InferenceRequest objects.");
+        "InferenceRequest objects. InferenceRequest size is:" +
```
Tabrizian, line 499: "InferenceRequest(s) size is"

src/pb_stub.cc
```diff
@@ -494,9 +494,12 @@ Stub::Execute(RequestBatch* request_batch, ResponseBatch* response_batch)
...
+        std::to_string(batch_size) +
+        ", and response size is:" + std::to_string(response_size) + "\n";
```
Tabrizian, line 501: "InferenceResponse(s) size is"

inferentia/scripts/gen_triton_model.py
```diff
@@ -628,13 +796,20 @@ def create_model_file(using_tensorflow_model):
     elif FLAGS.model_type == 'pytorch':
         is_tensorflow_model = False
 
+    enable_batching = FLAGS.max_batch_size > 0
```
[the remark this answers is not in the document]
jbkyang-nvi: Does `max_batch_size` matter at all if dynamic batching is disabled?

inferentia/scripts/gen_triton_model.py
jbkyang-nvi: I was trying to speed up execution for the default case (no batching) so there doesn't need to be extra checks

inferentia/scripts/gen_triton_model.py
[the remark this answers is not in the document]
tanmayv25: Yes.. The server will reject requests client sends with a batch size greater than `max_batch_size`...
It is like a validation parameter.

inferentia/scripts/gen_triton_model.py
tanmayv25: You can speed up the execution when the server sent only a single request to the backend. That is even if the dynamic batching is enabled, the server does not always send multiple requests to the backend. For instance, if the delay time expired and only a single request was received by the server, we don't need to perform extra logic...

inferentia/README.md
```diff
@@ -156,6 +158,24 @@ loaded on cores 2-3. To best engage inferentia device, try setting
 the number of neuron cores to be a proper multiple of the instance
 count.
 
+#### Batching with Pytorch
+
+There are three flags related to batched inputs:
```
tanmayv25, line 163: `gen_triton_model.py`  supports following three options for configuring (triton's dynamic batching)\<link to the triton dynamic batching\>:

1. --preferred_batch_size: <description with link to triton's docs>
2. --max_queue_delay_microseconds: <desc with link to triton's docs>
3. --disable_batch_requests_to_neuron: \<desc\>

inferentia/README.md
```diff
@@ -189,6 +209,10 @@ next available Neuron cores and not specific ones. See
 [Parallel Execution using NEURONCORE_GROUP_SIZES](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/neuron-guide/appnotes/perf/parallel-ncgs.html?highlight=NEURONCORE_GROUP_SIZES)
 for more information.
 
+#### Batching with Tensorflow
+
+The batching flags have the same functionality for Tensorflow
```
tanmayv25, line 214: This is confusing. You should not have separate sections for PyTorch and TensorFlow. Add a common section after 'Serving Inferentia on Triton' section. You can call it `Using Triton's Dynamic Batching` with Inferentia models.

inferentia/scripts/gen_triton_model.py
```diff
@@ -336,7 +351,8 @@ def _validate_output_dict(self, expected_count):
     return init_impl
 
 
-def get_tensorflow_execute_impl():
+def get_tensorflow_execute_impl(enable_dynamic_batching,
```
tanmayv25, line 354: These python models doesnot do anything to enable triton's batching. When dyanmic batching in triton is enabled then Triton Server will itself provide multiple requests in one execute call if it was able to batch multiple requests. It is inefficient to run the extra logic even if the server was unable to get multiple requests for the python backend. 

enable_dynamic_batching should not appear in python models.. Instead it should rely on the requests list received. If the list contains single request then nothing extra needs to be done.

inferentia/scripts/gen_triton_model.py
```diff
@@ -374,54 +391,133 @@ def execute(self, requests):
...
+            inference_response = pb_utils.InferenceResponse(output_tensors=output_tensors)
+            responses.append(inference_response)
 
         return responses
 '''
     return exec_impl
 
 
-def get_pytorch_execute_impl():
+def get_pytorch_execute_impl(enable_dynamic_batching,
```
tanmayv25, line 519: See my comment above.. enable_dynamic_batching should not appear here.

tanmayv25 (changes requested):

inferentia/scripts/gen_triton_model.py
jbkyang-nvi: When `enable_dynamic_batching` is specified though, Triton will send inputs of `[X, input dim]` according to my tests. eg: Input provided from perf_analyzer is [6, 4], model accepts [X, 4] but all the requests sent are of type [1,6,4]. 
This is why `squeeze` is required for batched requests.

However, if batching is not enabled, the inputs are [6,4].

Are you suggesting I squeeze the inputs regardless?

tanmayv25: Ohh.. I didn't see that before..  You should not design the python model on the basis of what perf analyzer generates. That is just a tool for running inferencing on server. Perf analyzer can be modified to generate requests expected by server but we must not loose generality.

So this is what your python model should be doing.. If the shape expected is in the following format: [batch size, X]. 
If the python model in its exec function gets three requests in the list as follows: [6, 4], [6,4] and [6, 4],
It should simply form a tensor of size [18, 4] and run that in neuron SDK. 

Now to test this you should run perf_analyzer as follows:

`perf_analyzer -m <model_name> --concurrency-range 10 -b 6` 

Note the model config should specify the dim as [4] and have max_batch_size specified as greater than or equal to 18. To form a batch of size 6, perf_analyzer goes through the tensors in a round-robin fashion.  Look at these documentation for more context.. https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#inputs-and-outputs
https://github.com/triton-inference-server/server/blob/main/docs/perf_analyzer.md#real-input-data

tanmayv25 (approved):

PR #109, opened 2022-01-05
Fix zero length io in BLS

Tabrizian:


src/pb_main_utils.cc
```diff
@@ -265,10 +265,18 @@ RequestExecutor::Infer(
...
-              sname, dims_vector, datatype, memory_type, memory_type_id,
-              const_cast<void*>(base), byte_size, nullptr /* DLManagedTensor */,
-              *(reinterpret_cast<off_t*>(userp))));
+          if (byte_size != 0) {
+            output_tensors.push_back(std::make_shared<PbTensor>(
+                sname, dims_vector, datatype, memory_type, memory_type_id,
+                const_cast<void*>(base), byte_size,
+                nullptr /* DLManagedTensor */,
+                *(reinterpret_cast<off_t*>(userp))));
+          } else {
```
CoderHam, line 274: We do want to support zero length/empty tensors as I/O in Triton. I don't think other backend support empty tensors for I/O? 

Additionally this only adds support for zero length outputs. Do we not need to add extra handling for zero length inputs?

CoderHam (changes requested):

src/pb_main_utils.cc
Tabrizian: >We do want to support zero length/empty tensors as I/O in Triton. I don't think other backend support empty tensors for I/O?

I am not sure about other backends.

> Additionally this only adds support for zero length outputs. Do we not need to add extra handling for zero length inputs?

The test in the server repo tests both inputs and outputs.

CoderHam (approved):

PR #110, opened 2022-01-05
Fix error handling to include failed counts in infer stat

tanmayv25:


Tabrizian (commented):
Instead of using `goto`, can we create a class that reports the batch statistics when the object is being destructed? Then, we can change the `RESPOND_ALL_AND_SET_TRUE_IF_ERROR` to return from the function. Returning from the function will always trigger the destructor and the batch statistics will be reported.

Tabrizian (approved):

PR #111, opened 2022-01-07
Improve the error checking for execute return

Tabrizian:


src/pb_stub.cc
```diff
@@ -478,15 +478,25 @@ Stub::Execute(RequestBatch* request_batch, ResponseBatch* response_batch)
...
 
   if (is_coroutine) {
-    responses = asyncio.attr("run")(execute_return);
+    responses_obj = asyncio.attr("run")(execute_return);
   } else {
-    responses = execute_return;
+    responses_obj = execute_return;
   }
 
+  // Check the return type of execute function.
```
tanmayv25, line 490: Should add a test case in our CI for execute functions that don't return list.

Tabrizian: https://github.com/triton-inference-server/server/pull/3783

CoderHam (approved):

tanmayv25 (approved):

PR #112, opened 2022-01-10
Fix background thread sleep

Tabrizian:
I made a mistake and it looks like `sleep` function only accepts an unsigned int as the argument.

Closes https://github.com/triton-inference-server/server/issues/3779.

It should probably fix https://github.com/triton-inference-server/server/issues/3386 too.

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -790,7 +790,7 @@ main(int argc, char** argv)
           // shared memory and will be set to false by the parent process.
           // The parent process expects that the stub process sets this
           // variable to true within 1 second.
-          sleep(0.3);
+          std::this_thread::sleep_for(std::chrono::milliseconds(300));
 
           stub->UpdateHealth();
 


CoderHam (approved):
What was sleep(0.3) doing? Sleeping for 1 second instead of 0.3?

Tabrizian:
It was not sleeping at all and casted to zero which lead to the CPU consumption.

KshitizLohia:
@Tabrizian How do we know as to what version of `tritonserver` to pick in order in which this fix is present?

KshitizLohia:
@Tabrizian How do we know what version of `tritonserver` to pick in which this fix is present?

Tabrizian:
There is not any released version of Triton that has this fix. You need to compile the Python backend from the main branch by yourself using the instructions [here](https://github.com/triton-inference-server/python_backend#building-from-source)

KshitizLohia:
@Tabrizian Till when can we expect a new version to be released with this fix?

Tabrizian:
This fix will be included in the 22.01 release which will be released in the next few weeks.

tanmayv25:
@hegc  Please create a new github issue here: https://github.com/triton-inference-server/server/issues
with the reproduction steps.

PR #115, opened 2022-01-19
Fix BLS support for sequence models

Tabrizian:
Fix the support for sequence models in BLS. New `sequence_start` and `sequence_end` flags have been added to indicate the starting and ending of sequences in BLS requests. An example BLS request for sequence models:

```python
 infer_request = pb_utils.InferenceRequest(
            model_name='onnx_nobatch_sequence_int32',
            inputs=[input],
            requested_output_names=['OUTPUT'],
            sequence_start=True,
            sequence_end=False,
            correlation_id=correlation_id)
```

Testing in the server PR:

https://github.com/triton-inference-server/server/pull/3826

Closes https://github.com/triton-inference-server/server/issues/3770

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -40,11 +40,10 @@ InferRequest::InferRequest(
     const std::vector<std::shared_ptr<PbTensor>>& inputs,
     const std::vector<std::string>& requested_output_names,
     const std::string& model_name, const int64_t model_version,
-    bool sequence_start, bool sequence_end)
+    const uint32_t flags)
     : request_id_(request_id), correlation_id_(correlation_id), inputs_(inputs),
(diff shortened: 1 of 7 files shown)


src/infer_request.cc
```diff
@@ -82,6 +84,18 @@ InferRequest::ModelVersion()
   return model_version_;
 }
 
+bool
+InferRequest::SequenceStart()
```
deadeyegoodwin, line 88: Existing Triton code just uses Flags(), so it would be good to continue that pattern. See sequence_batcher for example.

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

tanmayv25 (approved):

PR #117, opened 2022-01-25
Docs to let user know async BLS needed Python >= 3.7

CoderHam:


README.md
```diff
@@ -661,6 +661,9 @@ class TritonPythonModel:
 A complete example for sync and async BLS in Python backend is included in the
 [Examples](#examples) section.
 
+Note: Async BLS is not supported on Python 3.6 or lower due to the async keyword
```
Tabrizian, line 664: Small nit: use `async` instead of async.

README.md
```diff
@@ -661,6 +661,9 @@ class TritonPythonModel:
...
+and asyncio.run being introduced in Python 3.7.
```
Tabrizian, line 665: Same for `asyncio.run`.

Tabrizian (changes requested):

README.md
CoderHam: Good catch

README.md
CoderHam: Done

README.md
```diff
@@ -661,6 +661,9 @@ class TritonPythonModel:
 A complete example for sync and async BLS in Python backend is included in the
 [Examples](#examples) section.
 
+Note: Async BLS is not supported on Python 3.6 or lower due to the `async` keyword
+and `asyncio.run` being introduced in Python 3.7.
```
Tabrizian, line 665: Also, I think we should add a sentence that the Python version used in Jetson is 3.6.

CoderHam: Do we call out the Python version used for x86? I can add it there

CoderHam: PS: I call out that Async BLS is not supported on Jetson 
"The Python backend on Jetson does not support GPU Tensors and Async BLS."

Tabrizian: This sounds good to me. We mention the x86 version in a lot of places in the doc (you can grep for 3.8 and see all the places its mentioned).

Tabrizian (approved):

PR #122, opened 2022-02-10
Use the new shared memory manager for initialize and finalize

Tabrizian:
In this PR, the new shared memory interface is implemented. Currently, only `initialize` and `finalize` functions can work properly. The lifecycle management of the objects is handled such that if one makes a mistake and forgets to handle the lifecycle properly, they will see a double free instead of shared memory leak which is silent and can go unnoticed.

All the other logic (request execution, BLS, etc.) has been removed temporarily from the code and will be added back in separate PRs.


naor2013:
@Tabrizian Out of interest, can you explain what would be different with this?
Will it impact us as users in some way (performance improvement, or something else) or is that only affecting things internally?

Tabrizian:
@naor2013 This mainly improves the lifecycle management of objects in shared memory and addresses limitation 1 for BLS:

https://github.com/triton-inference-server/python_backend#limitations

src/pb_tensor.h
```diff
@@ -72,23 +72,15 @@ class PbTensor {
   PYTHONBACKEND_TensorType tensor_type_;
   uint64_t byte_size_;
   DLManagedTensor* dl_managed_tensor_;
-  Tensor* tensor_shm_;
-  RawData* raw_data_shm_;
 
 #ifdef TRITON_ENABLE_GPU
-  bool is_cuda_handle_set_;
-  cudaIpcMemHandle_t* cuda_ipc_mem_handle_ = nullptr;
+
```
tanmayv25, line 77: remove this newline

src/pb_utils.h
```diff
@@ -76,34 +76,40 @@ namespace bi = boost::interprocess;
...
+#define THROW_IF_ERROR(MSG, X)           \
+  do {                                   \
+    int return__ = (X);                  \
+    if (return__ != 0) {                 \
+      throw PythonBackendException(MSG); \
+    }                                    \
+  } while (false)
 
-struct InitializeResponse {
+struct InitializeResponseShm {
```
tanmayv25, line 87: Are these Shm renaming required?

tanmayv25 (changes requested):

src/message_queue.cc
```diff
@@ -175,50 +169,59 @@ MessageQueue::Pop(int const& duration, bool& success)
...
+  new (SemFullMutable()) bi::interprocess_semaphore(0);
+  new (SemEmptyMutable()) bi::interprocess_semaphore(Size());
+  new (MutexMutable()) bi::interprocess_mutex;
 }
 
 std::unique_ptr<MessageQueue>
 MessageQueue::LoadFromSharedMemory(
-    std::unique_ptr<SharedMemory>& shm_pool, off_t message_queue_offset)
+    std::unique_ptr<SharedMemoryManager>& shm_pool,
+    bi::managed_external_buffer::handle_t message_queue_offset)
```
CoderHam, line 199: message_queue_offset -> message_queue_handle?

src/ipc_message.cc
```diff
@@ -29,61 +29,121 @@
...
+
+  return std::unique_ptr<IPCMessage>(
+      new IPCMessage(ipc_message_shm, response_mutex_shm, response_cond_shm));
+}
 
 std::unique_ptr<IPCMessage>
 IPCMessage::LoadFromSharedMemory(
-    std::unique_ptr<SharedMemory>& shm_pool, off_t message_offset)
+    std::unique_ptr<SharedMemoryManager>& shm_pool,
+    bi::managed_external_buffer::handle_t message_offset)
```
CoderHam, line 59: message_offset is no longer an offset but a handle right?

src/ipc_message.h
```diff
@@ -54,65 +54,55 @@ typedef enum PYTHONSTUB_commandtype_enum {
...
+/// \param response_offset determines the request offset.
 struct IPCMessageShm {
   PYTHONSTUB_CommandType command;
-  off_t args;
+  bi::managed_external_buffer::handle_t args;
   bool inline_response = false;
-  off_t request_offset;
-  off_t response_mutex;
-  off_t response_cond;
+  bi::managed_external_buffer::handle_t response_offset;
```
CoderHam, line 63: handle?

CoderHam (changes requested):

src/message_queue.h
```diff
@@ -29,64 +29,95 @@
...
-/// Struct holding the represenation of a message queue inside the shared
+/// Struct holding the represenation of a message stack inside the shared
 /// memory.
-/// \param size Total size of the message queue.
+/// \param size Total size of the message stack.
 /// \param mutex Offset of the mutex variable protecting index.
 /// \param index Used element index.
 /// \param sem_empty Semaphore object counting the number of empty buffer slots.
 /// \param sem_full Semaphore object counting the number of used buffer slots.
 struct MessageQueueShm {
```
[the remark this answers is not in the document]
Tabrizian: Yes, this was message stack but has been converted to a message queue in #123.

src/pb_utils.h
Tabrizian: It is not required just a new convention I'm trying to follow. All the structs that reside in shared memory are suffixed with `Shm`.

src/ipc_message.cc
Tabrizian: Boost typedefs the handles to a simple number. So they are basically the same thing. But I'll rename all the offsets to handle to be consistent.

PR #127, opened 2022-03-14
Move BLS to use the new interface

Tabrizian:
- [x] Enable BLS for CPU Tensors
- [x] Fix GPU tensors for output tensors
- [x] Fix GPU tensors for the input of BLS requests

`L0_backend_python` passes successfully with this change.


src/scoped_defer.cc
```diff
@@ -0,0 +1,52 @@
...
+{
+  if (!done_) {
+    task_();
+    done_ = true;
+  }
+}
+
+ScopedDefer::~ScopedDefer()
+{
+  if (!done_) {
```
tanmayv25, line 47: Can we directly call Complete here?

src/shm_manager.h
```diff
@@ -101,7 +102,8 @@ class SharedMemoryManager {
   }
 
   template <typename T>
-  AllocatedSharedMemory<T> Load(bi::managed_external_buffer::handle_t handle)
+  AllocatedSharedMemory<T> Load(
+      bi::managed_external_buffer::handle_t handle, bool unsafe = false)
```
tanmayv25, line 106: Where and why do we need unsafe=true. It looks like I missed where it was being used.

tanmayv25 (changes requested):

src/infer_request.h
```diff
@@ -61,16 +62,22 @@ class InferRequest {
...
   bi::managed_external_buffer::handle_t ShmHandle();
 
+#ifdef TRITON_PB_STUB
+  std::unique_ptr<InferResponse> Exec();
+#endif
+
   /// Save an Inference Request to shared memory.
   /// \param shm_pool Shared memory pool to save the inference request.
-  void SaveToSharedMemory(std::unique_ptr<SharedMemoryManager>& shm_pool);
+  void SaveToSharedMemory(
```
tanmayv25, line 71: add copy_gpu in documentation above.

src/pb_tensor.h
```diff
@@ -142,7 +142,8 @@ class PbTensor {
 
   static std::shared_ptr<PbTensor> LoadFromSharedMemory(
```
tanmayv25, line 143: We must document these functions as well.

src/request_executor.cc
```diff
@@ -139,8 +154,15 @@ ResponseRelease(
   switch (memory_type) {
     case TRITONSERVER_MEMORY_CPU:
     case TRITONSERVER_MEMORY_CPU_PINNED: {
-      off_t* offset = reinterpret_cast<off_t*>(buffer_userp);
-      delete offset;
+      AllocationInfo* allocation_info =
+          reinterpret_cast<AllocationInfo*>(buffer_userp);
+      {
+        // Load the data so that it is deallocated automatically.
+        auto result = allocation_info->shm_manager_->Load<char>(
```
tanmayv25, line 161: Found it. Why are we sending unsafe=true?

tanmayv25 (changes requested):

src/infer_request.cc
```diff
@@ -173,10 +177,10 @@ InferRequest::SaveToSharedMemory(std::unique_ptr<SharedMemoryManager>& shm_pool)
   requested_output_names_shm_ = std::move(requested_output_names_shm);
 }
 
-std::unique_ptr<InferRequest>
+std::shared_ptr<InferRequest>
```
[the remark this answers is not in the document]
Tabrizian: The reason we return a shared pointer for the infer request object is that in the `async_exec` call we want to keep the infer request object around until the request execution is finished. Because of this the callback needs to keep a reference to the infer request object until it is the request's turn to execute. This is done in the L722 of pb_stub.cc.

src/shm_manager.h
Tabrizian: It is used in the request executor. The reason is that when response alloc is called we release the object to match the C interface of ResponseAlloc. When it is loaded with unsafe=true, it will be loaded without changing the reference count to make up for the released object.

src/request_executor.cc
Tabrizian: Explained in other comment.

src/infer_request.h
```diff
@@ -61,16 +62,22 @@ class InferRequest {
...
+      std::unique_ptr<SharedMemoryManager>& shm_pool, bool copy_gpu = true);
 
   /// Create an Inference Request object from shared memory.
   /// \param shm_pool Shared memory pool
   /// \param request_handle Shared memory handle of the request.
-  static std::unique_ptr<InferRequest> LoadFromSharedMemory(
+  static std::shared_ptr<InferRequest> LoadFromSharedMemory(
```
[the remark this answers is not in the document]
Tabrizian: Converted to unique_ptr.

tanmayv25 (approved):

PR #134, opened 2022-04-07
Improve argument type checking

Tabrizian:
Will not merge it before the code-freeze.

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -41,10 +41,26 @@ InferRequest::InferRequest(
     const std::vector<std::string>& requested_output_names,
     const std::string& model_name, const int64_t model_version,
     const uint32_t flags)
-    : request_id_(request_id), correlation_id_(correlation_id), inputs_(inputs),
-      requested_output_names_(requested_output_names), model_name_(model_name),
-      model_version_(model_version), flags_(flags)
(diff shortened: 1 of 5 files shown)


src/infer_request.cc
```diff
@@ -41,10 +41,26 @@ InferRequest::InferRequest(
...
+  for (auto& input : inputs) {
+    if (!input) {
+      throw PythonBackendException(
+          "Input tensor for request with id '" + request_id +
+          "' and model name '" + model_name + "' should not be empty.");
+    }
+  }
+
+  for (auto& requested_output_name : requested_output_names) {
+    if (requested_output_name == "") {
```
dyastremsky, line 56: For consistency, it may make sense to make the empty string check be the same here and in the above if statement for inputs.

dyastremsky: Unless you have a reason for doing otherwise. Then, a comment would be good.

dyastremsky (changes requested):

tanmayv25 (approved):

src/infer_request.cc
Tabrizian: Added the same to pb_tensor.cc

dyastremsky (approved):

tanmayv25 (approved):

CoderHam (approved):

PR #137, opened 2022-04-18
Some examples for using decoupled API

tanmayv25:
These example model.py can be easily swapped with repeat_int32 and square_int32 in L0_decoupled_test for a decent test coverage. Other special cases will be covered in L0_backend_python tests. 
The detailed explanation of the examples will be added with another PR.

### DIFF
--- a/examples/decoupled/repeat_client.py
+++ b/examples/decoupled/repeat_client.py
@@ -1,0 +1,116 @@
+# Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import sys
+from functools import partial
+import numpy as np
+
+from tritonclient.utils import *
+import tritonclient.grpc as grpcclient
+
+
+class UserData:
+
+    def __init__(self):
+        self._completed_requests = queue.Queue()
+
+
+def callback(user_data, result, error):
+    if error:
+        user_data._completed_requests.put(error)
+    else:
+        user_data._completed_requests.put(result)
+
+
+# This client sends a single request to the model with the
+# following tensor data. In compliance with the behavior
+# of repeat_int32 model, it will expect the 4 responses
+# with output: [4], [2], [0] and [1] respectively.
+model_name = "repeat_int32"
+in_value = [4, 2, 0, 1]
+delay_value = 2
+wait_value = 5
+
+inputs = []
+inputs.append(grpcclient.InferInput('IN', [len(in_value)], "INT32"))
+inputs.append(grpcclient.InferInput('DELAY', [1], "UINT32"))
+inputs.append(grpcclient.InferInput('WAIT', [1], "UINT32"))
+
+outputs = []
+outputs.append(grpcclient.InferRequestedOutput('OUT'))
+outputs.append(grpcclient.InferRequestedOutput('IDX'))
+
+with grpcclient.InferenceServerClient(url="localhost:8001",
+                                      verbose=True) as triton_client:
+    # Establish stream
+    triton_client.start_stream(callback=partial(callback, user_data))
+
+    in_data = np.array(in_value, dtype=np.int32)
+    inputs[0].set_data_from_numpy(in_data)
+    delay_data = np.array([delay_value], dtype=np.uint32)
+    inputs[1].set_data_from_numpy(delay_data)
+    wait_data = np.array([wait_value], dtype=np.uint32)
+    inputs[2].set_data_from_numpy(wait_data)
+
+    request_id = "0"
+    triton_client.async_stream_infer(model_name=model_name,
+                                     inputs=inputs,
+                                     request_id=request_id,
+                                     outputs=outputs)
+
+    # Retrieve results...
+    recv_count = 0
+    expected_count = len(in_value)
+    result_dict = {}
+    while recv_count < expected_count:
+        data_item = user_data._completed_requests.get()
+        if type(data_item) == InferenceServerException:
+            raise data_item
+        else:
+            this_id = data_item.get_response().id
+            if this_id not in result_dict.keys():
+                result_dict[this_id] = []
+            result_dict[this_id].append((recv_count, data_item))
+
+        recv_count += 1
+
+    # Validate results...
+    if len(result_dict[request_id]) != len(in_values):
+        print("expected {} many responses for request id {}, got {}".format(
+            len(in_values), request_id, len(result_dict[request_id])))
+        sys.exit(1)
+
+    result_list = result_dict[request_id]
+    for i in range(len(result_list)):
+        expected_data = np.array([in_values[i]], dtype=np.int32)
+        this_data = result_list[i][1].as_numpy('OUT')
+        if not np.array_equal(expected_data, this_data):
+            print("incorrect data: expected {}, got {}".format(
+                expected_data, this_data))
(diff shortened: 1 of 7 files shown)


examples/decoupled/square_model.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns
+        -------
+        None
+        """
+
+        for request in requests:
+            self.process_request(request)
+
+        return None
```
GuanLuo, line 141: Why `repeat` and `square` models don't wait for the execution? Doesn't that mean there can be multiple concurrent executions even with 1 instance?

tanmayv25: >  Doesn't that mean there can be multiple concurrent executions even with 1 instance?

That is true. The behavior is documented here: https://github.com/triton-inference-server/python_backend/blob/tanmayv-decoupled/examples/decoupled/repeat_model.py#L59

Once the issue thread exits it can re-enter if there are new requests in the model. There can be multiple responses from multiple requests in-flight from the model/server. I can make the documentation more explicit. 

See here for how we do it in our C++ example: 
repeat: https://github.com/triton-inference-server/repeat_backend/blob/main/src/repeat.cc#L488-L500
square: https://github.com/triton-inference-server/square_backend/blob/main/src/square.cc#L360-L372

examples/decoupled/repeat_model.py
```diff
@@ -0,0 +1,206 @@
...
+    def execute(self, requests):
+        """`execute` MUST be implemented in every Python model. `execute`
+        function receives a list of pb_utils.InferenceRequest as the only
+        argument. This function is called when an inference request is made
+        for this model. The request.get_response_sender() must be used to
+        get an InferenceResponseSender object associated with the request.
+        Use the InferenceResponseSender.send() and InferenceResponseSender.close()
+        calls to send responses and indicate no responses will be sent for
+        the corresponding request respectively. If there is an error, you can
+        set the error argument when creating a pb_utils.InferenceResponse
```
Tabrizian, line 121: A dot is missing at the end of the sentence.

examples/decoupled/repeat_model.py
```diff
@@ -0,0 +1,206 @@
...
+        requests : list
+          A list of pb_utils.InferenceRequest
+
+        Returns
+        -------
+        None
+        """
+
+        # This model does not support batching, so 'request_count' should always be 1.
+        if len(requests) != 1:
```
Tabrizian, line 134: Is this check required? With `max_batch_size: 0` the requests size will be always 1.

examples/decoupled/repeat_model.py
```diff
@@ -0,0 +1,206 @@
...
+
+        thread.start()
+
+        # Read WAIT input for wait time, then return so that Triton can call execute again with
+        # another request.
+        wait_input = pb_utils.get_input_tensor_by_name(requests[0],
+                                                       'WAIT').as_numpy()
+        time.sleep(wait_input[0] / 1000)
+
+        return None
```
Tabrizian, line 159: Can you add a comment above this sentence and contrast this with the non-decoupled API mode, where the has to be a list of inference response objects here?

examples/decoupled/repeat_model.py
```diff
@@ -0,0 +1,206 @@
...
+        the model to perform any necessary clean ups before exit.
+        Here we will wait for all response threads to complete sending
+        responses.
+        """
+        print('Finalize invoked')
+
+        inflight_threads = True
+        while inflight_threads:
+            with self.inflight_thread_count_lck:
+                inflight_threads = (self.inflight_thread_count != 0)
```
Tabrizian, line 202: Can you add a print statement here that prints the number of inflight threads? Similar to what we have in Triton when the responses have not been released.

examples/decoupled/square_model.py
```diff
@@ -0,0 +1,189 @@
...
+    def execute(self, requests):
+        """`execute` MUST be implemented in every Python model. `execute`
+        function receives a list of pb_utils.InferenceRequest as the only
+        argument. This function is called when an inference request is made
+        for this model. The request.get_response_sender() must be used to
+        get an InferenceResponseSender object associated with the request.
+        Use the InferenceResponseSender.send() and InferenceResponseSender.end()
+        calls to send responses and indicate no responses will be sent for
+        the corresponding request respectively. If there is an error, you can
+        set the error argument when creating a pb_utils.InferenceResponse
```
Tabrizian, line 126: Add a dot at the end of the sentence.

examples/decoupled/square_model.py
```diff
@@ -0,0 +1,189 @@
...
+          A list of pb_utils.InferenceRequest
+
+        Returns
+        -------
+        None
+        """
+
+        for request in requests:
+            self.process_request(request)
+
```
Tabrizian, line 140: Can you add a comment above this sentence and contrast this with the non-decoupled API mode, where the has to be a list of inference response objects here?

Tabrizian (changes requested):

Tabrizian (commented):
Can you change the branch that this PR is being merged to from `main` to `deocupled-api`? Also, you need to add the client example similar to the `add_sub`. https://github.com/triton-inference-server/python_backend/blob/main/examples/add_sub/client.py

examples/decoupled/repeat_model.py
tanmayv25: request size will not be 1 only when using dynamic batching. Anyways, this check makes it clear the model is not a batching model. And as the config is a separate file than this model.py, I think it makes it more comprehensible.

examples/decoupled/square_model.py
GuanLuo: I see, it makes sense to showcase the flexibility of the decouple style: execution / request / response are not closely coupled, and their lifecycle can be managed in arbitrary order.

Should probably point out that this bypasses the instance limitation and shouldn't be adopted to actual backends as it can result in over-committing work in the backend

tanmayv25: Added a comment in the execute function to explain the pitfall.

examples/decoupled/repeat_client.py
```diff
@@ -0,0 +1,116 @@
...
+inputs = []
+inputs.append(grpcclient.InferInput('IN', [len(in_value)], "INT32"))
+inputs.append(grpcclient.InferInput('DELAY', [1], "UINT32"))
+inputs.append(grpcclient.InferInput('WAIT', [1], "UINT32"))
+
+outputs = []
+outputs.append(grpcclient.InferRequestedOutput('OUT'))
+outputs.append(grpcclient.InferRequestedOutput('IDX'))
+
+with grpcclient.InferenceServerClient(url="localhost:8001",
```
GuanLuo, line 66: Okay syntax for Python, but should wrap the process in `if __name__ == '__main__':`

examples/decoupled/repeat_model.py
```diff
@@ -0,0 +1,228 @@
...
+# and converting Triton input/output types to numpy types.
+import triton_python_backend_utils as pb_utils
+
+
+class TritonPythonModel:
+    """Your Python model must use the same class name. Every Python model
+    that is created must have "TritonPythonModel" as the class name.
+
+    This model demonstrates how to write a decoupled model where each
+    request can generate 0 to many responses.
```
GuanLuo, line 42: Should make sure we have test Python decouple in L0_decoupled test. Should be able to just swapping the existing repeat / square models in the test?

examples/decoupled/repeat_model.py
```diff
@@ -135,7 +135,16 @@ def execute(self, requests):
...
+        # Start a separate thread to send the responses for the request. A model using
+        # decoupled transaction policy is not required to send all responses for the
+        # current request before returning from the execute. The sending the responses
+        # is delegated to this thread.
+        # This means the model can process multiple requests at the same time in
+        # the same model instance, as it defers responding to requests in an
+        # earlier invocation until after responding to requests from a later request.
+        # Even though this demonstrates the flexibility the decoupled API,
+        # it can lead to many requests into pipeline whose responses are being
+        # processed. A model developer should be mindful of this.
```
Tabrizian, line 147: Can you clarify more on why the user should be mindful of this? Are you saying that there could be many requests that Triton core needs to keep track of or you are saying that there could be many responses in the queue that Python backend is copying/retrieving from shared memory?

Tabrizian (changes requested):

examples/decoupled/repeat_model.py
Tabrizian: > request size will not be 1 only when using dynamic batching. 

Can you clarify a little bit? Are you saying that the requests size can be larger than 1 even with max_batch_size equal to zero?

examples/decoupled/repeat_model.py
Tabrizian: Exactly. That is the plan.

examples/decoupled/repeat_model.py
tanmayv25: No.. requests size will be one if max_batch_size == 0. 
I am saying it will also be 1 for max_batch_size > 1. 
The request size will not be 1 if max_batch_size > 1 and dynamic batching is enabled.
I have added this test to make it explicit that the model is not a batching model. So, it will fail if it is tried to be used with config.pbtxt setting max_batch_size > 1 and enabling dynamic batching. A user can change the config.pbtxt as per their needs and this check only makes sure that model will return an appropriate error.

examples/decoupled/repeat_client.py
tanmayv25: I didn't wrap the script in main because none of our client scripts in the example directory are currently within main.
@Tabrizian Was this intentional? If yes, then why?

examples/decoupled/repeat_model.py
tanmayv25: Yes. For this release we are aiming to be able to successfully run L0_decoupled test with these python models. Currently, the feature development is still in progress but we will have another run where these python models are used instead of square and repeat backends.

examples/decoupled/repeat_model.py
```diff
@@ -135,7 +135,16 @@ def execute(self, requests):
...
                                                 len(requests))
 
-        # Start a separate thread to send the responses for the request.
+        # Start a separate thread to send the responses for the request. A model using
+        # decoupled transaction policy is not required to send all responses for the
+        # current request before returning from the execute. The sending the responses
+        # is delegated to this thread.
+        # This means the model can process multiple requests at the same time in
+        # the same model instance, as it defers responding to requests in an
+        # earlier invocation until after responding to requests from a later request.
```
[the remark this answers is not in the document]
tanmayv25: Add some more clear comments. Let me know if they bring pitfalls clearly.

examples/decoupled/repeat_client.py
Tabrizian: There is not any particular reason. We can wrap those clients in `if __name__ == '__main__'` too.

Tabrizian (approved):

examples/decoupled/repeat_client.py
GuanLuo: Should use `__main__` then, otherwise the code will be executed if the code is accidentally imported in other Python script.

tanmayv25: I will do this in another PR.

GuanLuo (approved):

Tabrizian (commented):
Found some minor issues while running the model for testing purposes.

PR #138, opened 2022-04-21
Python models should have API where they can provide max_batch/inputs/outputs for auto-complete

krishung5:
To-Dos:
- [x] Implement the auto-complete logic
- [x] Update documentation
- [x] Add test to cover the changes : https://github.com/triton-inference-server/server/pull/4322


README.md
```diff
@@ -189,6 +190,45 @@ class TritonPythonModel:
     that is created must have "TritonPythonModel" as the class name.
     """
 
+    def auto_complete_config(self):
```
Tabrizian, line 193: This should accept an additional argument named `model_config` that the user will use to auto_complete.

examples/add_sub/model.py
```diff
@@ -38,6 +38,50 @@ class TritonPythonModel:
...
+        Must return a `pb_utils.ModelConfig` object which contains the updated
+        model configuration.
+        Note: All the objects in this function will go out of scope after exiting.
+        Should not store any objects in this function.
+
+        Parameters
+        ----------
+        args : dict
+          Both keys and values are strings. The dictionary keys and values are:
+          * model_config: A JSON string containing the current model configuration
```
Tabrizian, line 62: How about constructing a `ModelConfig` object instead of JSON and passing it to the `auto_complete_config` here?

examples/add_sub/model.py
```diff
@@ -38,6 +38,50 @@ class TritonPythonModel:
     that is created must have "TritonPythonModel" as the class name.
     """
 
+    def auto_complete_config(self, args):
```
Tabrizian, line 41: How about making this a static method?

src/python.cc
```diff
@@ -224,12 +227,58 @@ class ModelState : public BackendModel {
...
+
  private:
   ModelState(TRITONBACKEND_Model* triton_model);
   BackendState* backend_state_;
   std::string python_execution_env_;
   bool force_cpu_only_input_tensors_;
   bool decoupled_;
+
+  // Stub process pid
+  pid_t stub_pid_;
```
Tabrizian, line 259: Can we refactor the stub functionality used in the ModelState and ModelInstanceState to another class named `StubLauncher` to avoid code duplication?

src/resources/triton_python_backend_utils.py
```diff
@@ -294,5 +294,208 @@ def triton_string_to_numpy(triton_type_string):
...
+        self._model_config["input"].append(new_input._get_input())
+
+    def set_output(self, new_output):
+        """Set the output for the model.
+        Parameters
+        ----------
+        new_output : list
+            The output to be set.
+        """
+        self._model_config["output"].append(new_output._get_output())
```
Tabrizian, line 377: After the discussions we had, we can support set_raw_json too.

Tabrizian (changes requested):

README.md
```diff
@@ -41,6 +41,7 @@ any C++ code.
   - [Quick Start](#quick-start)
   - [Building from Source](#building-from-source)
   - [Usage](#usage)
+    - [`auto_complete_config`](#auto_complete_config)
```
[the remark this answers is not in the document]
krishung5: Thanks for the catch! I think the link works but it's better to use the dash so that it looks more like the other ones. Will modify.

README.md
```diff
@@ -189,6 +190,45 @@ class TritonPythonModel:
     that is created must have "TritonPythonModel" as the class name.
     """
 
+    def auto_complete_config(self):
+        """`auto_complete_config` is called only once when the server is started
+        with `--strict-model-config=false`. Implementing `auto_complete_config`
```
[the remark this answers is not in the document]
krishung5: You're right. The auto-complete logic will only set the fields that are not set, so implementing `auto_complete_config` should be optional. Will modify the comment.

README.md
krishung5: I was wondering that since we put all the auto-complete logic and checking in c++, do we still need to pass the `model_config` argument to this function?

examples/add_sub/model.py
krishung5: May I ask is it because we want to pass the JSON with a wrapper as well?

README.md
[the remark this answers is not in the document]
krishung5: Update: I tried on my side and I don't think the link would work if the name is using underscore while the link is using dash.

src/resources/triton_python_backend_utils.py
Tabrizian: @krishung5 Please ignore this comment for now. We can add this later when there is a specific request for it.

examples/add_sub/model.py
Tabrizian: It is not because of that. I think we should be consistent in using the abstractions and that is why it might be better to wrap it in that object.

README.md
Tabrizian: They still need to receive the original model configuration (if any) that was provided by the server or config.pbtxt file and that is why I think they would still need to access the original model configuration.

README.md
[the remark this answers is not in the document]
Tabrizian: Thanks for trying that. Let's keep it the underscores then.

src/resources/triton_python_backend_utils.py
```diff
@@ -295,6 +295,128 @@ def triton_string_to_numpy(triton_type_string):
...
+    ----------
+    config : AutoCompleteConfig object
+        object containing the model configuration
+    value : int
+        the max_batch_size we want to set for the model
+    """
+    config.set_max_batch_size(value)
+
+
+def set_input(config, new_input):
```
tanmayv25, line 310: Sorry for the late request. But what do you think about renaming `set_<property_name>` to `add_<property_name>`?
Also, new_input -> input
new_output -> output

src/resources/triton_python_backend_utils.py
```diff
@@ -295,6 +295,128 @@ def triton_string_to_numpy(triton_type_string):
...
+    ----------
+    config : AutoCompleteConfig object
+        object containing the model configuration
+    new_input : dict
+        containing the input we want to set for the model
+    Raises
+    ------
+    ValueError
+        If 'new_input' contains property other than 'name',
+        'data_type' and 'dims' or any of the properties is
```
tanmayv25, line 322: .. are not set.

Do we have tests for missing properties?

src/python.cc
```diff
@@ -228,19 +233,72 @@ class ModelState : public BackendModel {
...
+  // Auto-complete the model configuration
+  TRITONSERVER_Error* AutoCompleteConfig();
+  TRITONSERVER_Error* FixIO(
+      triton::common::TritonJson::Value& reference_ios,
+      triton::common::TritonJson::Value* mutable_ios);
+  TRITONSERVER_Error* CheckModelConfigMismatch();
+  TRITONSERVER_Error* ValidateModelConfig();
+
+  // Correct the string to the right json string format
+  void CorrectStringFormat(std::string* str);
```
tanmayv25, line 250: FixStringToJsonFormat

src/python.cc
```diff
@@ -2121,6 +2325,461 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
...
+  *str = std::regex_replace(*str, std::regex("False"), "false");
+  *str = std::regex_replace(*str, std::regex("True"), "true");
+}
+
+TRITONSERVER_Error*
+ModelState::AutoCompleteConfig()
+{
+  RETURN_IF_ERROR(CheckModelConfigMismatch());
+
+  // If the model configuration already specifies inputs and outputs
```
tanmayv25, line 2458: This is not accurate. We need to auto-complete missing information even if the model config already specify input and output.  The model_configuration may have specified only one model input but the model might be expecting 2.
I think you copied this logic from onnx backend. 
See the the enhancement request here: https://github.com/triton-inference-server/onnxruntime_backend/issues/113

src/python.cc
```diff
@@ -2121,6 +2325,461 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
...
+    if (ModelConfig().Find("output", &outputs)) {
+      output_cnt = outputs.ArraySize();
+    }
+  }
+
+  // Determine if the model can potentially support batching. All
+  // input and output tensors must have a variable first dimension.
+  // However, ragged batching is an exception to this rule. A tensor
+  // allowing ragged batch is in itself a batching hint.
+  bool config_batch_hint = false;
```
tanmayv25, line 2483: Much of the logic for checking batching support can be be simplified. TF/ONNX/TRT models provide the exact shape that is expected by the model. For example take a model with input and shape: INPUT0 : [-1, 16]. There is no way to detect whether to enable batching on these model while auto-completing. Hence, we look for the hint in config.pbtxt file. If in the file they have provided INPUT0 as [16]. Then that means the batching needs to be enabled.

For us with the current design it is simpler. The python model if it supports batching will call set_max_batch_size with a non-zero value in the model.py. It will still set INPUT0 as [16] as we are providing the config dim in the model.py. Because the dim we are providing is config dim and not model dim, we can not actually look for such hints here. You will see it yourself if you try the batch_model.py, I proposed above.

The logic here should be simple. If the user has provided non-zero max_batch_size in model.py or in model.pbtxt, then treat it as a batching model. If both are non-zero and there is a mismatch between the two sizes then throw an error.

tanmayv25 (changes requested):

src/resources/triton_python_backend_utils.py
krishung5: I think renaming `set_<property_name>` to `add_<property_name>` is more clear. But for max_batch_size, do we want to do the same?

src/resources/triton_python_backend_utils.py
krishung5: Tests are added.

src/python.cc
krishung5: Thanks for the clarification and the link. The logic is updated. Please review.

examples/auto_complete/batch_model.py
```diff
@@ -0,0 +1,180 @@
...
+        model_config : string
+          A JSON string containing the available model configuration. The users can
+          refer to this configuration when setting the properties for this model.
+
+        Returns
+        -------
+        pb_utils.ModelConfig
+          An object containing the max_batch_size, inputs and outputs properties.
+        """
+        auto_complete_model_config = pb_utils.ModelConfig()
```
Tabrizian, line 66: Update based on the comment above.

src/pb_stub.cc
```diff
@@ -255,6 +255,76 @@ Stub::RunCommand()
...
+        if (error_string_shm != nullptr) {
+          auto_complete_response.data_->response_is_error_set = true;
+          auto_complete_response.data_->response_error =
+              error_string_shm->ShmHandle();
+        }
+
+        return true;  // Terminate the stub process.
+      }
+
+      if (auto_complete_config != "None" || auto_complete_config.size() != 0) {
```
Tabrizian, line 316: Do we have a test for this? I don't think `auto_complete_config != "None"` works properly. It should be `auto_complete_config == py::none()`.

src/python.cc
```diff
@@ -211,6 +212,8 @@ struct BackendState {
   std::unique_ptr<EnvironmentManager> env_manager;
 };
 
+class StubLauncher;
```
Tabrizian, line 215: Can you move the StubLauncher to a separate file named `stub_launcher.cc` and `stub_launcher.h`?

src/python.cc
```diff
@@ -361,8 +421,394 @@ class ModelInstanceState : public BackendModelInstance {
...
+  StubLauncher* Stub() { return instance_stub_; }
 };
 
+class StubLauncher {
+ public:
+  ~StubLauncher();
+
+  // Stub process setup
+  TRITONSERVER_Error* Setup(
+      ModelState* model_state, pid_t* parent_pid_, bool* initialized_,
```
Tabrizian, line 435: Can we make the Setup method as simple as possible and hide any unnecessary detail from the ModelState, ModelInstanceState? We can add accessor methods to the stub class to get the objects that ModelState, ModelInstanceState needs but I don't think we should make all the internal attributes of the stub launcher a pointer to the ModelState/ModelInstanceState attributes.

src/python.cc
```diff
@@ -361,8 +421,394 @@ class ModelInstanceState : public BackendModelInstance {
...
+  (*stub_message_queue_)->ResetSemaphores();
+  (*parent_message_queue_)->ResetSemaphores();
+
+  (*initialized_) = false;
+
+  return nullptr;
+}
+
+TRITONSERVER_Error*
+StubLauncher::StubProcess(
```
Tabrizian, line 627: StubProcess -> Launch

Also, can you make forking the process part of launching the stub process too?

Tabrizian (changes requested):

README.md
```diff
@@ -190,6 +191,44 @@ class TritonPythonModel:
...
+
+        Parameters
+        ----------
+        model_config : string
+          A JSON string contains the model configuration.
+
+        Returns
+        -------
+        pb_utils.ModelConfig
+          An object containing the initial and updated model configuration
```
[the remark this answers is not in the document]
tanmayv25: Just be careful here. TRITONBACKEND_ModelSetConfig supports only inputs, outputs and max_batch_size. 
rest will be ignored.
https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonbackend.h#L976-L992

README.md
```diff
@@ -190,6 +191,46 @@ class TritonPythonModel:
...
+        `pb_utils.add_output`. These properties will allow Triton to load the model
+        with minimal model configuration in absence of a configuration file. This
+        function returns the `pb_utils.ModelConfig` object with these properties.
+
+        Note: All the objects in this function will go out of scope after exiting.
+        Should not store any objects in this function.
+
+        Parameters
+        ----------
+        model_config : string
```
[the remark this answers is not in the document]
tanmayv25: How would they refer to the config then? As we don't have any accessor functions on pb_utils.ModelConfig .

I think this was added so that users can refer to the current config and determine the inputs/outputs and max_batch_size.

examples/auto_complete/batch_config.pbtxt
```diff
@@ -0,0 +1,28 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "batch_auto_complete"
+backend: "python"
```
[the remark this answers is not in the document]
tanmayv25: For now. Yes. but FYI we have a ticket in this release DLIS-3699. Where we plan to use more standard way of naming models.
The backend name will be encoded in the model name itself. `model.<backend_name>`

README.md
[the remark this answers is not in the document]
Tabrizian: How about having a `as_dict` that provides a read-only access to the model configuration? cc @krishung5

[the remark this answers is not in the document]
krishung5: I think using `as_dict` to provide the model configuration sounds good to me.

README.md
```diff
@@ -190,6 +191,46 @@ class TritonPythonModel:
...
+          A JSON string containing the available model configuration. The users can
+          refer to this configuration when setting the properties for this model.
+
+        Returns
+        -------
+        pb_utils.ModelConfig
+          An object containing the max_batch_size, inputs and outputs properties.
+        """
+        auto_complete_model_config = pb_utils.ModelConfig()
```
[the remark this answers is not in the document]
krishung5: I was thinking that the model config in model.py should be built independently and should not rely on the configuration in config.pbtxt. I think this is also more like the model file we have in other backends. May I ask the reason behind that the users should build upon the configuration provided in config.pbtxt?

src/python.cc
krishung5: > ...and hide any unnecessary detail from the ModelState, ModelInstanceState
Could you elaborate more on this? I'm not sure what unnecessary detail here means..

README.md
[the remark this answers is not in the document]
Tabrizian: They are kind of dependent even now. For example, if the user has specified an input in the config.pbtxt file we will be returning an error saying that the input has already been specified. The reason I'm requesting this change is that if we ever (not now) decide to support other fields in the model config, we can more easily explain that the config returned here (if it doesn't conflict with the config provided through other mechanisms) will be used as the final configuration. Right now, it looks like we are merging two separate configurations which can be a bit confusing.

src/python.cc
Tabrizian: I mean for example, I think the stub launcher doesn't need to have `parent_pid`, `model_state`, `initialized` etc as the setup arguments. These are internal fields to stub launcher that can be accessed using accessor methods (e.g. `Initialized()`, etc).

krishung5: I see. Thanks for the explanation. Will modify the Setup method.

README.md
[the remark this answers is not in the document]
krishung5: > This will also centralize all the error checking in the stub side instead of being spread in pb_stub.cc and python.cc
For error checking like [CheckModelConfigMismatch](https://github.com/triton-inference-server/python_backend/blob/krish-python/src/python.cc#L2570), do we want to put this in the stub side? I was thinking that this should be part of auto-complete.

> we can more easily explain that the config returned here (if it doesn't conflict with the config provided through other mechanisms) will be used as the final configuration.
For auto-complete logic, we only query the config from model.py when there is missing information in the config from config.pbtxt. If the configuration in model.py is used as the final configuration, which contains the excisting config and the newly added config, is the auto-complete part still needed?

src/python.cc
```diff
@@ -2121,6 +2328,396 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
...
+{
+  *str = std::regex_replace(*str, std::regex("\'"), "\"");
+  *str = std::regex_replace(*str, std::regex("False"), "false");
+  *str = std::regex_replace(*str, std::regex("True"), "true");
+}
+
+TRITONSERVER_Error*
+ModelState::AutoCompleteConfig()
+{
+  RETURN_IF_ERROR(CheckModelConfigMismatch());
```
tanmayv25, line 2480: With https://github.com/triton-inference-server/python_backend/pull/138/files#r868251707
All this will no longer be here and most of the validation will move to the stub process.
The only checks would be:

1. `set_max_batch_size`: If config.pbtxt specified non-zero value and `set_max_batch_size` tries to set different max_batch_size, throw error.
2.`add_input`:  If config.pbtxt includes input with same name but differs datatype or shape, throw error. 
3. `add_output`: If config.pbtxt includes output with same name but differs datatype or shape, throw error.

Otherwise, it's all good. We don't need any other checks.
CC @Tabrizian

tanmayv25: Just use the config returned from `GetAutoCompleteConfig()` directly.

tanmayv25 (changes requested):

README.md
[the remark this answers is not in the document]
krishung5: Got it! Thank you both for all the suggestion, will modify accordingly.

src/python.cc
Tabrizian: Thanks Kris!

src/python.cc
Tabrizian: Sounds good to me!

Tabrizian (changes requested):
We can discuss the rest of the changes offline.

CMakeLists.txt
```diff
@@ -241,7 +244,6 @@ target_link_libraries(
    Threads::Threads
    triton-backend-utils           # from repo-backend
    pybind11::embed
-   dlpack
```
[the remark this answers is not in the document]
krishung5: I think this line is duplicated: https://github.com/triton-inference-server/python_backend/blob/main/CMakeLists.txt#L240

src/stub_launcher.cc
```diff
@@ -0,0 +1,584 @@
...
+    }
+  }
+
+  if (auto_complete_response->response_has_model_config) {
+    std::unique_ptr<PbString> auto_complete_config =
+        PbString::LoadFromSharedMemory(
+            (*shm_pool), auto_complete_response->response_model_config);
+    std::string auto_complete_config_string = auto_complete_config->String();
+    if (!auto_complete_config_string.empty()) {
+      FixStringToJsonFormat(&auto_complete_config_string);
```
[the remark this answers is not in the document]
krishung5: Because in the JSON string returned from model.py, python is using "True/False", but in c++ side, we need to use "true/false" to make it parsed successfully.

src/python.h
```diff
@@ -0,0 +1,399 @@
...
+  int64_t shm_growth_byte_size;
+  int64_t stub_timeout_seconds;
+  int64_t shm_message_queue_size;
+  std::atomic<int> number_of_instance_inits;
+  std::string shared_memory_region_prefix;
+  int64_t thread_pool_size;
+  std::unique_ptr<EnvironmentManager> env_manager;
+};
+
+class StubLauncher;
```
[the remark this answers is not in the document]
krishung5: I think we still need this line since we need to use forward declaration for StubLauncher class.

src/python.h
```diff
@@ -0,0 +1,399 @@
...
+      TRITONSERVER_Error* rarie_err__ = TRITONSERVER_ErrorNew( \
+          TRITONSERVER_ERROR_INTERNAL, pb_exception.what());   \
+      return rarie_err__;                                      \
+    }                                                          \
+  } while (false)
+
+namespace triton { namespace backend { namespace python {
+
+namespace bi = boost::interprocess;
+#ifndef PYTHON_H
```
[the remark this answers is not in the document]
krishung5: I tried to use `#pragma once` instead but seems like it's not working with the forward declaration of StubLauncher class. But using `#ifndef PYTHON_H` works fine. Maybe I'm missing something here?

src/resources/triton_python_backend_utils.py
```diff
@@ -295,6 +296,184 @@ def triton_string_to_numpy(triton_type_string):
...
+        max_batch_size : int
+            The max_batch_size to be set.
+        Raises
+        ------
+        ValueError
+            If configuration has specified max_batch_size non-zero value which
+            is larger than the max_batch_size to be set for the model.
+        """
+        if self._model_config["max_batch_size"] > max_batch_size:
+            raise ValueError(
```
tanmayv25, line 338: Should reformat the error message similar to others which helps users localize the error to auto-complete-config implementation.  Also include the model name.

krishung5: Sorry I missed modifying this error message. Updated, please review.

tanmayv25 (dismissed):
Looks good to me. I would defer final approval to @Tabrizian.

README.md
```diff
@@ -192,6 +193,84 @@ class TritonPythonModel:
...
+            'data_type': 'TYPE_FP32',
+            'dims': [4]
+        }, {
+            'name': 'OUTPUT1',
+            'data_type': 'TYPE_FP32',
+            'dims': [4]
+        }]
+
+        # Demonstrate the usage of `as_dict`, `add_input`, `add_output`,
+        # and `set_max_batch_size` functions.
```
Tabrizian, line 246: "Store the model configuration as a dictionary"

README.md
```diff
@@ -192,6 +193,84 @@ class TritonPythonModel:
     that is created must have "TritonPythonModel" as the class name.
     """
 
+    @staticmethod
+    def auto_complete_config(auto_complete_model_config):
+        """`auto_complete_config` is called only once when the server is started
```
Tabrizian, line 198: Please keep the number of characters in each line below 80.

README.md
```diff
@@ -252,7 +331,27 @@ class TritonPythonModel:
...
+with [`--strict-model-config=false`](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#auto-generated-model-configuration).
+Implementing this function is optional. A no implementation of `auto_complete_config` will do nothing.
+This function can be used to set 
+[`max_batch_size`](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#maximum-batch-size),
+[`input`](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#inputs-and-outputs) and
+[`output`](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#inputs-and-outputs)
+properties of the model using `set_max_batch_size`, `add_input`, and `add_output`.
+These properties will allow Triton to load the model with
+[minimal model configuration](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#minimal-model-configuration)
+in absence of a configuration file. This function returns the `pb_utils.ModelConfig` object with these properties.
```
Tabrizian, line 348: Keep the number of characters below 80.

examples/auto_complete/batch_model.py
```diff
@@ -0,0 +1,218 @@
...
+import triton_python_backend_utils as pb_utils
+
+
+class TritonPythonModel:
+    """Your Python model must use the same class name. Every Python model
+    that is created must have "TritonPythonModel" as the class name.
+    """
+
+    @staticmethod
+    def auto_complete_config(auto_complete_model_config):
```
Tabrizian, line 42: Keep the number of characters below 80.

src/pb_stub.cc
```diff
@@ -402,7 +470,40 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
...
+
+  py::module python_backend_utils =
+      py::module_::import("triton_python_backend_utils");
+  py::object model_config =
+      python_backend_utils.attr("ModelConfig")(pb_string_shm->String());
+
+  if (py::hasattr(model_, "auto_complete_config")) {
+    model_config = model_.attr("auto_complete_config")(model_config);
+  }
+  if (py::isinstance<py::none>(model_config)) {
```
Tabrizian, line 495: The error checking here needs to be improved to make sure the returned object is an instance of pb_utils.ModelConfig.

src/pb_stub.cc
```diff
@@ -402,7 +470,40 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
...
+
+  return TritonPythonModel();
+}
+
+void
+Stub::AutoCompleteModelConfig(
+    bi::managed_external_buffer::handle_t string_handle,
+    std::string* auto_complete_config)
+{
+  model_ = StubSetup();
```
Tabrizian, line 482: Why are we constructing the instance here? Since the auto_complete_config method is a static method we can call this method without constructing an object.

src/pb_stub.cc
```diff
@@ -402,7 +470,40 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
...
+
+  if (py::hasattr(model_, "auto_complete_config")) {
+    model_config = model_.attr("auto_complete_config")(model_config);
+  }
+  if (py::isinstance<py::none>(model_config)) {
+    throw PythonBackendException(
+        "auto_complete_config function in model '" + name_ +
+        "' must return a valid pb.ModelConfig object.");
+  }
+  (*auto_complete_config) = std::string(py::str(model_config));
```
Tabrizian, line 500: I think the reason that you need the FixStringToJSON in python_be.cc is that you are not really converting the dictionary to a json here. You should be calling json.dumps to convert the dictionary to a JSON: https://docs.python.org/3/library/json.html#json.dumps

src/python_be.cc
```diff
@@ -622,277 +280,26 @@ ModelInstanceState::RespondErrorToAllRequests(
 void
 ModelInstanceState::WaitForBLSRequestsToFinish()
 {
-  futures_.clear();
+  (*Stub()->Futures()).clear();
```
Tabrizian, line 283: The futures should be part of the model instance state not the stub launcher.

src/python_be.cc
```diff
@@ -981,83 +388,24 @@ ModelInstanceState::SaveRequestsToSharedMemory(
 }
 
 TRITONSERVER_Error*
-ModelInstanceState::SetupStubProcess()
+ModelInstanceState::LaunchModelInstanceStubProcess()
```
Tabrizian, line 391: LaunchStubProcess

src/python_be.cc
```diff
@@ -981,83 +388,24 @@ ModelInstanceState::SaveRequestsToSharedMemory(
...
+  Stub() = std::make_unique<StubLauncher>(
+      "MODEL_INSTANCE_STUB", Name(), DeviceId(),
+      TRITONSERVER_InstanceGroupKindString(Kind()));
+  RETURN_IF_ERROR(Stub()->Initialize(model_state));
+  RETURN_IF_ERROR(Stub()->Launch());
 
   if (model_state->IsDecoupled()) {
     decoupled_thread_ = true;
-    decoupled_monitor_ =
+    *Stub()->DecoupledMonitor() =
```
Tabrizian, line 402: I think the decoupled monitor doesn't need to be part of the stub launcher and can remain in model instance state.

src/python_be.cc
```diff
@@ -1420,15 +773,15 @@ ModelInstanceState::DecoupledMessageQueueMonitor()
         ResponseSendDecoupled(response_send_message);
       });
       std::future<void> future =
-          boost::asio::post(*thread_pool_, std::move(task));
-      futures_.emplace_back(std::move(future));
+          boost::asio::post(*Stub()->ThreadPool(), std::move(task));
```
Tabrizian, line 776: thread pool should be part of the instance state too.

src/python_be.cc
```diff
@@ -1616,7 +971,7 @@ ModelInstanceState::ProcessRequests(
 
   // This means that the stub process has exited and Python
   // backend failed to restart the stub process.
-  if (stub_pid_ == 0) {
+  if (*(Stub()->StubPid()) == 0) {
```
Tabrizian, line 974: Why is StubPid returning a pointer? Can we just return the pid itself?

src/python_be.cc
```diff
@@ -2128,6 +1451,40 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
...
+ModelState::LaunchAutoCompleteStubProcess()
+{
+  Stub() = std::make_unique<StubLauncher>("AUTOCOMPLETE_STUB");
+  RETURN_IF_ERROR(Stub()->Initialize(this));
+  try {
+    RETURN_IF_ERROR(Stub()->Launch());
+  }
+  catch (const BackendModelException& ex) {
+    Stub()->TerminateStub();
+    // Need to make sure that the shared memory pool is destructed properly
```
Tabrizian, line 1464: Can you expand on this comment? If we destruct the stub, Would this automatically delete the stub launcher as well?

src/stub_launcher.cc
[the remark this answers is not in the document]
Tabrizian: I left a new comment above on why I think this shouldn't be required. I think the correct json format doesn't accept True/False as valid boolean values.

src/stub_launcher.cc
```diff
@@ -0,0 +1,543 @@
...
+StubLauncher::KillStubProcess()
+{
+  kill(stub_pid_, SIGKILL);
+  int status;
+  waitpid(stub_pid_, &status, 0);
+  stub_pid_ = 0;
+}
+
+void
+StubLauncher::FixStringToJsonFormat(std::string* str)
```
Tabrizian, line 536: This shouldn't be required based on comment above.

Tabrizian (changes requested):
I think we are getting much closer. Thanks Kris!

src/python.h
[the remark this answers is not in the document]
Tabrizian: Looks like you have circular dependency stub_launcher.h <-> python_be.h. Do you need to include python_be.h in the stub_launcher.h class?

src/python.h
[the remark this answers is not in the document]
Tabrizian: This comment can probably explain why: https://github.com/triton-inference-server/python_backend/pull/138#discussion_r878566059

CMakeLists.txt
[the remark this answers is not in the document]
Tabrizian: Good catch.

src/python_be.cc
krishung5: I found that when destructing the stub here, `shm_pool` would not be reset correctly and would cause shm page not cleaned if we don't reset it here.

krishung5:
@Tabrizian Addressed and left some comments, please review. Thanks Iman!

src/resources/triton_python_backend_utils.py
```diff
@@ -310,7 +310,7 @@ def __init__(self, model_config):
         self._model_config = json.loads(model_config)
 
     def __str__(self):
```
Tabrizian, line 312: Can we remove this function?

src/pb_stub.cc
Tabrizian: Since the `auto_complete_config` is a static method we don't need to instantiate the object. I think we can just use the `py::hasattr(TritonPythonModel, "auto_complete_config")`.

src/pb_stub.cc
Tabrizian: Call json.dumps here.

src/python_be.cc
```diff
@@ -1961,64 +1319,7 @@ ModelInstanceState::ProcessRequests(
...
-    waitpid(stub_pid_, &status, 0);
-  }
-
-  // First destroy the IPCControl. This makes sure that IPCControl is
-  // destroyed before the shared memory manager goes out of scope.
-  ipc_control_.reset();
-  stub_message_queue_.reset();
-  parent_message_queue_.reset();
-  memory_manager_.reset();
+  Stub()->TerminateStub(&decoupled_monitor_, &futures_, &thread_pool_);
```
Tabrizian, line 1322: Instead of passing the arguments here, can we make sure that decoupled_monitor, futures, and thread_pool have exited properly here:

```
ModelState* model_state = reinterpret_cast<ModelState*>(Model());
      if (model_state->IsDecoupled()) {
        futures_.clear();
        Stub()-ParentMessageQueue()->Push(DUMMY_MESSAGE);
        decoupled_monitor_.join();
      }

      // Wait for all the futures to be finished.
      thread_pool_->wait();
```

examples/auto_complete/batch_config.pbtxt
```diff
@@ -0,0 +1,27 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "batch_auto_complete"
```
Tabrizian, line 27: Remove

examples/auto_complete/nobatch_config.pbtxt
```diff
@@ -0,0 +1,27 @@
+# Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Tabrizian, line 1: Remove

Tabrizian (commented):
We would also need a README.md for examples/auto_complete similar to other examples but it can be addressed in a separate PR.

src/python_be.cc
Tabrizian: What if we reset the stub here (i.e. call Stub().reset())?

src/resources/triton_python_backend_utils.py
krishung5: When returning `auto_complete_model_config`, this function will be called so that the returned `auto_complete_model_config` will have a converted json format. So I added `json.dumps` here and removed `FixStringToJsonFormat` function.
```
def __str__(self):
        return json.dumps(self._model_config)
```

src/pb_stub.cc
krishung5: Please see my comment: https://github.com/triton-inference-server/python_backend/pull/138#discussion_r881901906

krishung5:
Will create another PR for the README.md for examples/auto_complete.

src/python_be.cc
```diff
@@ -1467,9 +1483,9 @@ ModelState::LaunchAutoCompleteStubProcess()
     RETURN_IF_ERROR(Stub()->Launch());
   }
   catch (const BackendModelException& ex) {
-    Stub()->TerminateStub(nullptr, nullptr, nullptr);
-    // Need to make sure that the shared memory pool is destructed properly
-    Stub()->ShmPool().reset();
+    Stub()->UpdateHealth();
+    Stub()->TerminateStub();
```
Tabrizian, line 1487: I was wondering wondering why do we need to update health here?

src/python_be.cc
```diff
@@ -1319,7 +1319,21 @@ ModelInstanceState::ProcessRequests(
 
 ModelInstanceState::~ModelInstanceState()
 {
-  Stub()->TerminateStub(&decoupled_monitor_, &futures_, &thread_pool_);
+  ModelState* model_state = reinterpret_cast<ModelState*>(Model());
+  Stub()->UpdateHealth();
+  if (Stub()->Is_Healthy()) {
```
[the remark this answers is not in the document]
krishung5: Thanks for the catch!

src/python_be.cc
krishung5: Because we have this condition [here](https://github.com/triton-inference-server/python_backend/blob/krish-python/src/stub_launcher.cc#L482) inside `TerminateStub()` that requires the updated health variable.

src/python_be.cc
```diff
@@ -1350,8 +1363,11 @@ ModelState::Create(TRITONBACKEND_Model* triton_model, ModelState** state)
         &message, buf.Base(), buf.Size()));
     RETURN_IF_ERROR(TRITONBACKEND_ModelSetConfig(
         triton_model, 1 /* config_version */, message));
-    RETURN_IF_ERROR(
-        (*state)->ModelConfig().Swap((*state)->Stub()->AutoCompleteConfig()));
+    (*state)->ModelConfig() = std::move((*state)->Stub()->AutoCompleteConfig());
+
+    (*state)->Stub()->UpdateHealth();
```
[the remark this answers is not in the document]
krishung5: Please see my comment: https://github.com/triton-inference-server/python_backend/pull/138#discussion_r886194225

src/pb_stub.cc
```diff
@@ -315,6 +382,68 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
...
+      py::module_::import("triton_python_backend_utils");
+  py::object model_config =
+      python_backend_utils.attr("ModelConfig")(pb_string_shm->String());
+
+  if (py::hasattr(sys.attr("TritonPythonModel"), "auto_complete_config")) {
+    model_config = sys.attr("TritonPythonModel")
+                       .attr("auto_complete_config")(model_config);
+  }
+
+  if (std::string(py::str(model_config.get_type())) !=
```
Tabrizian, line 434: I think it would be better to use `py::isinstance` instead of checking the string of the type.

Tabrizian (changes requested):

Tabrizian:
Only these two minor comments and the PR LGTM! Thank you for all the effort you put into finishing up auto_complete.

tanmayv25 (approved):

PR #141, opened 2022-04-26
Add TYPE_BF16 scaffolding

rmccorm4:
Most backends handle TYPE_BF16 in their `default / else` cases. Simply added a `default` to one of the type conversion helpers, as well as a `TYPE_BF16` case just to have a clear error message since the specific type isn't included in the default case error.

Also removed unnecessary `break`s after `throw`s

### DIFF
--- a/src/pb_stub_utils.cc
+++ b/src/pb_stub_utils.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


rmccorm4:
@Tabrizian 

> Similar changes are required [here](https://github.com/triton-inference-server/python_backend/blob/8547d43a16d67039ac755396cf0932b2c23d7975/src/resources/triton_python_backend_utils.py#L216)

There is no numpy equivalent currently. That may end up related to Katherine's bf16 python client ticket. If the check in this PR throws an exception on BF16, will it make it to this [section](https://github.com/triton-inference-server/python_backend/blob/8547d43a16d67039ac755396cf0932b2c23d7975/src/resources/triton_python_backend_utils.py#L216) of the code? Or is there a valid `else` case we can return/raise exception in these as a catch all if not?

> Is similar scaffolding required for other backends too?

As far as I can tell, TF, TRT, ORT, and PyTorch backends all have a default/else case on these type conversions so they will reject TYPE_BF16 until backend-specific support is implemented in each.

Tabrizian:
1 & 2. On a second thought, I think there is no chance required in this file. We are currently returning None for both of the cases which is fine.

Tabrizian (approved):

Tabrizian:
Please just remove the `python_backend:` from the title of the PR before merging.

tanmayv25 (approved):

PR #147, opened 2022-05-06
Update building stub instructions to include other repo tags

Tabrizian:


rmccorm4 (changes requested):
Can you add a small note/example along the lines of this: (doesn't have to be this exact wording)

For example, if building based on a specific release branch such as `r22.04` , you should set `GIT_BRANCH=r22.04` for each dependent repo tag similarly:
```
# example
...
```

Tabrizian:
It is mentioned in the line above this line:

>  (replace <GIT_BRANCH_NAME> with the branch name that you want to use, for release branches it should be r<xx.yy>):

rmccorm4 (approved):

krishung5 (approved):

PR #154, opened 2022-05-11
Add GPU tensor support to decoupled API

Tabrizian:
- [x] Add support for GPU outputs
- [x] Add support for GPU inputs
- [x] Add backend utility to collect input buffers into a single buffer
- [x] Add testing

### DIFF
--- a/src/infer_response.cc
+++ b/src/infer_response.cc
@@ -177,7 +177,10 @@ InferResponse::Send(
       std::make_shared<TRITONSERVER_Error*>();
   std::unique_ptr<ScopedDefer> response_error_handling;
   requires_deferred_callback = false;
-  bool destruct_resposne_factor = (response == nullptr);
+
+  // Should only destruct the response factory whenever a response factory is
+  // being created.
+  bool destruct_response_factor = (response == nullptr);
 
   if (response == nullptr) {
     SET_ERROR_AND_RETURN(
@@ -187,12 +190,13 @@ InferResponse::Send(
 
   response_error_handling = std::make_unique<ScopedDefer>(
       [response, response_error, flags, response_factory,
-       destruct_resposne_factor] {
+       destruct_response_factor] {
         if (response != nullptr) {
           LOG_IF_ERROR(
               TRITONBACKEND_ResponseSend(response, flags, *response_error),
               "failed to send the response.");
-          if (flags == TRITONSERVER_RESPONSE_COMPLETE_FINAL) {
+          if (flags == TRITONSERVER_RESPONSE_COMPLETE_FINAL &&
+              destruct_response_factor) {
             std::unique_ptr<
                 TRITONBACKEND_ResponseFactory, backend::ResponseFactoryDeleter>
             response_factory_ptr(
--- a/src/infer_response.cc
+++ b/src/infer_response.cc
@@ -164,7 +164,7 @@ InferResponse::Error()
 }
 
 #ifndef TRITON_PB_STUB
-TRITONSERVER_Error*
+std::shared_ptr<TRITONSERVER_Error*>
 InferResponse::Send(
     TRITONBACKEND_ResponseFactory* response_factory, void* cuda_stream,
     bool& requires_deferred_callback, const uint32_t flags,
@@ -174,7 +174,7 @@ InferResponse::Send(
     TRITONBACKEND_Response* response)
 {
   std::shared_ptr<TRITONSERVER_Error*> response_error =
-      std::make_shared<TRITONSERVER_Error*>();
+      WrapTritonErrorInSharedPtr(nullptr);
   std::unique_ptr<ScopedDefer> response_error_handling;
   requires_deferred_callback = false;
 
@@ -188,6 +188,9 @@ InferResponse::Send(
         TRITONBACKEND_ResponseNewFromFactory(&response, response_factory));
   }
 
+  // This lambda expression will be called when this function exits, if the
+  // inference response doesn't have any GPU tensors. Otherwise, it will be
+  // called when the object is destructed or DeferredSendCallback is called.
   response_error_handling = std::make_unique<ScopedDefer>(
       [response, response_error, flags, response_factory,
        destruct_response_factor] {
@@ -206,6 +209,8 @@ InferResponse::Send(
         }
       });
 
+  // Moves the response sending callback so that it is not called until the stub
+  // process fills in the GPU buffers.
   ScopedDefer deferred_task(
       [this, &requires_deferred_callback, &response_error_handling] {
         if (requires_deferred_callback) {
@@ -315,7 +320,7 @@ InferResponse::Send(
   }
 #endif  // TRITON_ENABLE_GPU
 
-  return *response_error;
+  return response_error;
 }
 #endif
 
--- a/src/infer_response.cc
+++ b/src/infer_response.cc
@@ -167,29 +167,49 @@ InferResponse::Error()
 TRITONSERVER_Error*
 InferResponse::Send(
     TRITONBACKEND_ResponseFactory* response_factory, void* cuda_stream,
-    const uint32_t flags)
+    bool& requires_deferred_callback, const uint32_t flags,
+    std::unique_ptr<SharedMemoryManager>& shm_pool,
+    std::vector<std::unique_ptr<PbMemory>>& output_buffers,
+    const std::set<std::string>& requested_output_names,
+    TRITONBACKEND_Response* response)
 {
-  // [FIXME] Use this code to send responses in non-decoupled mode.
-  TRITONBACKEND_Response* response = nullptr;
   TRITONSERVER_Error* response_error = nullptr;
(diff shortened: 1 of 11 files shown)


src/response_sender.cc
```diff
@@ -129,6 +118,57 @@ ResponseSender::Send(
...
+            gpu_buffers_handle.data_.get() + sizeof(uint64_t));
+    uint64_t* gpu_buffer_count =
+        reinterpret_cast<uint64_t*>(gpu_buffers_handle.data_.get());
+    if (gpu_tensors.size() != *gpu_buffer_count) {
+      LOG_INFO
+          << (std::string(
+                  "GPU buffers size does not match the provided buffers: ") +
+              std::to_string(gpu_tensors.size()) +
+              " != " + std::to_string(*gpu_buffer_count));
+      return;
```
GuanLuo, line 147: Send error?

src/python.cc
```diff
@@ -1162,54 +1152,81 @@ ModelInstanceState::GetInputTensor(
...
-    // that the cuda shared memory was used and the input is in a single buffer.
-    // [FIXME] for the case where the input is in cuda shared memory and uses
-    // multiple input buffers this needs to be changed.
-    TRITONSERVER_BufferAttributes* buffer_attributes;
+      // If the tensor is using the cuda shared memory, we need to extract the
+      // handle that was used to create the device pointer. This is because of a
+      // limitation in the legacy CUDA IPC API that doesn't allow getting the
+      // handle of an exported pointer. If the cuda handle exists, it indicates
+      // that the cuda shared memory was used and the input is in a single
+      // buffer. [FIXME] for the case where the input is in cuda shared memory
```
GuanLuo, line 1167: Newline for [FIXME]

src/python.cc
```diff
@@ -1443,7 +1460,7 @@ ModelInstanceState::ResponseSendDecoupled(
   ResponseSendMessage* send_message_payload =
       reinterpret_cast<ResponseSendMessage*>(send_message.data_.get());
   std::unique_ptr<PbString> error_message;
-  ScopedDefer _([send_message_payload] {
+  ScopedDefer _([&send_message_payload] {
```
GuanLuo, line 1463: Why take reference of a pointer?

src/pb_stub.h
```diff
@@ -39,14 +44,74 @@
...
+
+  ~LogMessage() { gLogger_.Log(stream_.str()); }
+
+  std::stringstream& stream() { return stream_; }
+
+ private:
+  std::stringstream stream_;
+};
+
+#define LOG_INFO_FL(FN, LN) LogMessage((char*)(FN), LN).stream()
```
GuanLuo, line 113: Why not use the log function in Triton In-process API.

src/python.cc
Tabrizian: No particular reason. I'll revert this change.

tanmayv25 (changes requested):

src/infer_response.cc
```diff
@@ -167,50 +167,65 @@ InferResponse::Error()
...
-          "failed to send the response.");
-      if (flags == TRITONSERVER_RESPONSE_COMPLETE_FINAL) {
-        std::unique_ptr<
-            TRITONBACKEND_ResponseFactory, backend::ResponseFactoryDeleter>
-        response_factory_ptr(
-            reinterpret_cast<TRITONBACKEND_ResponseFactory*>(response_factory));
-      }
-    }
-  });
+  std::shared_ptr<TRITONSERVER_Error*> response_error =
```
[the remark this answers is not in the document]
GuanLuo: This doesn't sound right, when will "if an error is ever set" happen? This is near the end of inference pipeline and why we can't expect error has been set, if any, by this point?

src/pb_stub.h
GuanLuo: I think using backend API requires linking to libtritonserver.so already

Tabrizian: Python backend has a shared backend library and a stub process. The shared library is linked against libtritonserver.so but the stub process is not. This logging is only added in the stub process.

src/infer_response.cc
```diff
@@ -167,50 +167,65 @@ InferResponse::Error()
...
+      std::make_shared<TRITONSERVER_Error*>();
+  std::unique_ptr<ScopedDefer> response_error_handling;
+  requires_deferred_callback = false;
+  bool destruct_resposne_factor = (response == nullptr);
```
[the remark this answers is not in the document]
Tabrizian: Updated.

src/infer_response.cc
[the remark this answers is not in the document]
Tabrizian: Can you clarify on the question? The error can happen any time. This for handling the decoupled responses when the model has returned from the execute function.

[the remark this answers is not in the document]
tanmayv25: To my original question, wouldn't this need a custom deleter so that the shared_ptr abstraction actually destructs the TRITONSERVER_Error, instead of just a pointer to the pointer to the object? If there are indeed errors that can happen in the deferred task.

src/infer_response.cc
```diff
@@ -167,50 +167,69 @@ InferResponse::Error()
...
     TRITONSERVER_MemoryType actual_memory_type = src_memory_type;
     int64_t actual_memory_type_id = src_memory_type_id;
 
-    // [FIXME] GPU tensors are not supported in the decoupled API mode.
     if (actual_memory_type == TRITONSERVER_MEMORY_GPU) {
-      response_error = TRITONSERVER_ErrorNew(
-          TRITONSERVER_ERROR_INTERNAL,
-          "GPU tensors are not supported in decoupled API.");
-      return response_error;
+      requires_deferred_callback = true;
```
rmccorm4, line 232: I think a comment here explaining why GPU tensors require a deferred callback would be great for the reader.

Similarly, comments in a few spots that you added the use of `ScopedDefer` may also be helpful.

rmccorm4 (changes requested):
Using Go-style `defer` is cool to see 🙂

src/infer_response.cc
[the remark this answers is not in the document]
GuanLuo: Note that if the error does send `TRITONBACKEND_ResponseSend`, must make sure the error in shared pointer is set to nullptr as the error ownership is transferred.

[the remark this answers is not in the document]
Tabrizian: @GuanLuo Is it transferred though? As far as I can tell, the error object is not deleted by the Triton core and a copy is created. I think it is the responsibility of the caller to delete the object. https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L1029-L1048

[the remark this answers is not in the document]
GuanLuo: My bad, that is indeed the case

src/infer_response.cc
Tabrizian: Added

tanmayv25 (approved):

tanmayv25 (approved):

src/python.cc
```diff
@@ -1468,13 +1488,76 @@ ModelInstanceState::ResponseSendDecoupled(
...
+      send_message_payload->gpu_buffers_handle = gpu_buffers_handle.handle_;
+
+      size_t index = 0;
+      for (auto& output_buffer_pair : gpu_output_buffers) {
+        std::unique_ptr<PbMemory>& pb_memory = output_buffer_pair.first;
+        gpu_buffers_handle_shm[index] = pb_memory->ShmHandle();
+        ++index;
+      }
+
+      // Additional round trip so that the stub can fill the GPU output buffers.
```
GuanLuo, line 1520: Why there is another round trip? the backend process signals back that there is CUDA IPC handle and wait for stub process to copy the data to within the CUDA IPC handle?

Do I summarize the workflow correctly:

1. stub requests for output buffer (and passing buffer via shared memory at the same time if CPU tensor)
2. backend acquires output buffer, copy if CPU tensor, otherwise, passing the CUDA IPC handle back to stub.
3. (only for GPU tensor) stub then copy GPU tensor into the CUDA IPC handle

Tabrizian: Exactly. The third round is what requires the round trip.

src/infer_response.cc
```diff
@@ -167,50 +167,69 @@ InferResponse::Error()
...
+      [response, response_error, flags, response_factory,
+       destruct_response_factor] {
+        if (response != nullptr) {
+          LOG_IF_ERROR(
+              TRITONBACKEND_ResponseSend(response, flags, *response_error),
+              "failed to send the response.");
+          if (flags == TRITONSERVER_RESPONSE_COMPLETE_FINAL &&
+              destruct_response_factor) {
+            std::unique_ptr<
+                TRITONBACKEND_ResponseFactory, backend::ResponseFactoryDeleter>
```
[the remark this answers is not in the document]
Tabrizian: I'd rather keep it the way it is since they are almost equal.

rmccorm4 (approved):

PR #161, opened 2022-06-01
Add README for auto-complete example

krishung5:


examples/auto_complete/README.md
```diff
@@ -0,0 +1,110 @@
...
+[`auto_complete_config`](https://github.com/triton-inference-server/python_backend/#auto_complete_config)
+function in Python backend to provide
+[`max_batch_size`](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#maximum-batch-size),
+[`input`](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#inputs-and-outputs)
+and [`output`](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#inputs-and-outputs)
+properties. These properties will allow Triton to load the Python model with
+[MinimalModel Configuration](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#minimal-model-configuration)
+in absence of a configuration file. The
+[model repository](https://github.com/triton-inference-server/server/blob/main/docs/model_repository.md)
+should contain [nobatch_auto_complete](./nobatch_model.py), and
```
tanmayv25, line 41: Just before the model repository line, describe the distinction between batch and nobatch models in a line.

tanmayv25 (changes requested):

examples/auto_complete/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```console
+$ mkdir -p models/nobatch_auto_complete/1/
+$ mkdir -p models/batch_auto_complete/1/
+
+# Copy the Python models
+$ cp examples/auto_complete/nobatch_model.py models/nobatch_auto_complete/1/model.py
+$ cp examples/auto_complete/batch_model.py models/batch_auto_complete/1/model.py
+```
+(Note that we don't need a model configuration file since Triton will use the
```
tanmayv25, line 65: Better to highlight this statement in bold format.

tanmayv25 (approved):

examples/auto_complete/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+```console
+$ mkdir -p models/nobatch_auto_complete/1/
+$ mkdir -p models/batch_auto_complete/1/
+
+# Copy the Python models
+$ cp examples/auto_complete/nobatch_model.py models/nobatch_auto_complete/1/model.py
+$ cp examples/auto_complete/batch_model.py models/batch_auto_complete/1/model.py
+```
+**(Note that we don't need a model configuration file since Triton will use the
```
Tabrizian, line 69: Remove parenthesis in the beginning and end.

Tabrizian (commented):
Do you think it makes sense to add an additional input to `batch_auto_complete` and `nobatch_auto_complete` models that returns the model configuration received in the initialize function? We will verify that the configuration that the model returns matches with what expect in the client.

krishung5:
> Do you think it makes sense to add an additional input to `batch_auto_complete` and `nobatch_auto_complete` models that returns the JSON of the model configuration received in the initialize function? We will verify that the configuration that the model returns matches with what expect in the client.

Please correct me if I'm wrong, I assume this is for verifying if the model configuration is registered properly? I was wondering what is the difference between the model config received in the initialize function and the one retrieved using [get_model_config](https://github.com/triton-inference-server/python_backend/blob/main/examples/auto_complete/client.py#L80) from client?

examples/auto_complete/README.md
krishung5: Updated, please review.

Tabrizian:
I think they should be the same. What we have already in the client demonstrates that so please ignore my earlier comment. I guess what we have to do is only remove the inference performed on the model as it is not relevant to this example and instead print `{model_name} configuration matches the expected auto complete configuration`.

Tabrizian (approved):

PR #164, opened 2022-06-06
Fix deferred callback when there is an error

Tabrizian:


krishung5 (approved):

GuanLuo (approved):
Should we add the same "fail infer" test in L0_backend_python?

Tabrizian:
We do have a fail infer case in L0_backend_python here:

https://github.com/triton-inference-server/server/blob/main/qa/python_models/execute_error/model.py#L45-L46

The reason the test didn't fail is the second request in the batch fails and when a vector contains a single element the capacity is 64 elements which doesn't segfault the test.

GuanLuo:
We should improve the existing test or add new test then, as the current test doesn't capture this particular case?

PR #165, opened 2022-06-06
Autocomplete sets dynamic batcher if no scheduler set.

nv-kmcgill53:
Autocomplete checks if the max_batch_size is greater than 1 and if so, attempts to set dynamic_batching as the scheduler if none set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,9 @@ any C++ code.
     - [`initialize`](#initialize)
     - [`execute`](#execute)
       - [Default Mode](#default-mode)
-      - [Decoupled Mode](#decoupled-mode)
+      - [Decoupled mode](#decoupled-mode)
+        - [Use Cases](#use-cases)
+        - [Known Issues](#known-issues)
     - [`finalize`](#finalize)
   - [Model Config File](#model-config-file)
   - [Using Custom Python Execution Environments](#using-custom-python-execution-environments)
@@ -66,8 +68,9 @@ any C++ code.
 - [Examples](#examples)
   - [AddSub in NumPy](#addsub-in-numpy)
   - [AddSubNet in PyTorch](#addsubnet-in-pytorch)
-  - [Business Logic Scripting](#business-logic-scripting)
+  - [Business Logic Scripting](#business-logic-scripting-1)
   - [Preprocessing](#preprocessing)
+  - [Decoupled Models](#decoupled-models)
 - [Running with Inferentia](#running-with-inferentia)
 - [Reporting problems, asking questions](#reporting-problems-asking-questions)
 
@@ -243,7 +246,7 @@ class TritonPythonModel:
         }]
 
         # Demonstrate the usage of `as_dict`, `add_input`, `add_output`,
-        # and `set_max_batch_size` functions.
+        # `set_max_batch_size`, and `set_dynamic_batching` functions.
         # Store the model configuration as a dictionary.
         config = auto_complete_model_config.as_dict()
         input_names = []
@@ -269,6 +272,12 @@ class TritonPythonModel:
                 auto_complete_model_config.add_output(output)
 
         auto_complete_model_config.set_max_batch_size(0)
+        
+        # To enable a dynamic batcher with default settings, you can use 
+        # auto_complete_model_config set_dynamic_batching() function. It is 
+        # commented in this example because the max_batch_size is zero.
+        #
+        # auto_complete_model_config.set_dynamic_batching()
 
         return auto_complete_model_config
 
@@ -343,11 +352,13 @@ Implementing this function is optional. A no implementation of
 `auto_complete_config` will do nothing. This function can be used to set 
 [`max_batch_size`](
   https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#maximum-batch-size),
+[dynamic_batching](
+  https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#dynamic-batcher),
 [`input`](
   https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#inputs-and-outputs) and
 [`output`](
   https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#inputs-and-outputs)
-properties of the model using `set_max_batch_size`, `add_input`, and
+properties of the model using `set_max_batch_size`, `set_dynamic_batching`, `add_input`, and
 `add_output`. These properties will allow Triton to load the model with
 [minimal model configuration](
   https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#minimal-model-configuration)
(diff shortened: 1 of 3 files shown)


src/resources/triton_python_backend_utils.py
```diff
@@ -340,6 +342,28 @@ def set_max_batch_size(self, max_batch_size):
...
+
+    def set_dynamic_batching(self):
+        """Set dynamic_batching as the scheduler for the model if no scheduler is set.
+           If dynamic_batching is already set, then no action is taken and return success.
+        Raises
+        ------
+        ValueError
+            If the configuration has a scheduler set which is not dynamic_batching.
+        """
+        if "sequence_batching" in self._model_config.keys():
```
tanmayv25, line 356: Should we return an error if config had any other scheduling policy? I think we must just not attempt to add dynamic batching in such cases like in rest of the backends.

tanmayv25: https://github.com/triton-inference-server/tensorflow_backend/blob/main/src/tensorflow.cc#L1316-L1331

nv-kmcgill53: fixed, along with the tests.

src/resources/triton_python_backend_utils.py
```diff
@@ -340,6 +342,23 @@ def set_max_batch_size(self, max_batch_size):
                 str(max_batch_size))
         else:
             self._model_config["max_batch_size"] = max_batch_size
+            if max_batch_size > 1:
+                self.set_dynamic_batching()
```
Tabrizian, line 346: I think it we should not implicitly set the dynamic batching setting to ON when max_batch_size is greater than zero. Since currently there is not way for the user to set max_batch_size to greater than one and not turn on dynamic batching.

src/resources/triton_python_backend_utils.py
```diff
@@ -340,6 +342,23 @@ def set_max_batch_size(self, max_batch_size):
...
+
+    def set_dynamic_batching(self):
+        """Set dynamic_batching as the scheduler for the model if no scheduler is set.
+           If dynamic_batching is already set, then no action is taken and return success.
```
Tabrizian, line 350: Have you run the formatter? The number of characters should be below 80.

Tabrizian, line 350: Have you run the formatter? The number of characters should be below 80.

src/resources/triton_python_backend_utils.py
```diff
@@ -340,6 +342,23 @@ def set_max_batch_size(self, max_batch_size):
...
+        Raises
+        ------
+        ValueError
+            If the configuration has a scheduler set which is not dynamic_batching.
```
Tabrizian, line 354: I don't see ValueError being raised anywhere. Maybe we can raise a pb_utils.TritonModelException to be consistent with other APIs when sequence_batching/ensemble_scheduling/dynamic_batching exists in the configuration?

Tabrizian, line 354: I don't see ValueError being raised anywhere. Maybe we can raise a pb_utils.TritonModelException to be consistent with other APIs when sequence_batching/ensemble_scheduling/dynamic_batching exists in the configuration?

Tabrizian (changes requested):

src/resources/triton_python_backend_utils.py
nv-kmcgill53: Weird. I did, but the formatter didn't correct these lines. Manually done.

src/resources/triton_python_backend_utils.py
nv-kmcgill53: This is an out of date comment, and I have removed it as such. See the comment from @tanmayv25 above for more context.

src/resources/triton_python_backend_utils.py
Tabrizian: I think Python backend is a bit different from other backends in that it is more a Python API for the Triton's C-API. Because of this I think the API should be complete and have maximum flexiblity.

src/resources/triton_python_backend_utils.py
```diff
@@ -341,6 +341,18 @@ def set_max_batch_size(self, max_batch_size):
         else:
             self._model_config["max_batch_size"] = max_batch_size
 
+    def set_dynamic_batching(self):
+        """Set dynamic_batching as the scheduler for the model if no scheduler 
+        is set. If dynamic_batching is already set, then no action is taken and
+        return success.
+        """
+        if "sequence_batching" in self._model_config.keys(
+        ) or "ensemble_scheduling" in self._model_config.keys():
```
Tabrizian, line 350: We should raise ValueError here. Also, the condition can be updated to include dynamic batching.

README.md
```diff
@@ -269,6 +272,9 @@ class TritonPythonModel:
                 auto_complete_model_config.add_output(output)
 
         auto_complete_model_config.set_max_batch_size(0)
+        
+        # Uncomment this line when setting max_batch_size > 1
```
Tabrizian, line 276: "To enable a dynamic batcher with default settings, you can use `auto_complete_model_config.set_dynamic_batching()` function. It is commented in this example because the max_batch_size is zero"

src/resources/triton_python_backend_utils.py
nv-kmcgill53: Is there a better way in python to search through the same list multiple times? I had trouble finding anything on the python docs and general search.

Tabrizian: I think removing the `keys()` should probably make it more performant. Performance is not a huge concern here since we only auto-complete during model loading.

README.md
nv-kmcgill53: Updated.

src/resources/triton_python_backend_utils.py
nv-kmcgill53: Attempting to set dynamic_batching as the scheduler when one is already set (including `dynamic_batching`) will raise an error. [Please see this thread for more context](https://github.com/triton-inference-server/python_backend/pull/165#discussion_r891619472)

Tabrizian (approved):

README.md
Tabrizian: Thanks Kyle!

krishung5 (dismissed):

src/resources/triton_python_backend_utils.py
tanmayv25: @Tabrizian  @nv-kmcgill53 Why should we throw an error if user config has already provided dynamic_batching? 
A user may want to provide their own specific dynamic_batching settings via config.pbtxt. Things like max_queue_wait_time etc. 
We can just use whatever config the user has provided in this case.

nv-kmcgill53: I agree with @tanmayv25. If the user has provided us with dynamic_batching as the scheduler and we wish to auto-complete the scheduler to dynamic_batching, then the logic is consistent and no action should be taken. The user provided config should pass through.

Tabrizian: I agree with the no-op if the config has dynamic batching settings and the user is trying to use dynamic batching. We can remove the error for this case.

PR #171, opened 2022-07-06
Move tensorflow default version from 1 to 2

krishung5:


Tabrizian:
I think according to Tanmay's comment [here ](https://github.com/triton-inference-server/server/pull/4601#discussion_r916404857) this should be reverted too?

krishung5:
> I think according to Tanmay's comment [here ](https://github.com/triton-inference-server/server/pull/4601#discussion_r916404857) this should be reverted too?

Ah right.. sorry I missed that.  Thanks Iman, closing this PR.

PR #173, opened 2022-07-12
Fix requested output names

Tabrizian:


src/pb_stub.cc
```diff
@@ -741,13 +741,17 @@ Stub::ProcessRequests(RequestBatch* request_batch_shm_ptr)
...
 
+
     response_batch_shm_ptr->batch_size = response_size;
 
     std::vector<std::shared_ptr<PbTensor>> gpu_tensors;
 
     size_t i = 0;
-    for (auto& response : responses) {
-      InferResponse* infer_response = response.cast<InferResponse*>();
+    for (; i < batch_size; i++) {
```
jbkyang-nvi, line 750: confused why we don't just do
```suggestion
    for (size_t i = 0; i < batch_size; i++) {
```

fpetrini15 (approved):
Other than Katherine's comment, LGTM.

src/pb_stub.cc
Tabrizian: No, particular reason. Fixed.

krishung5 (dismissed):
Just out of curiosity, was the issue related to output names confliction?

Tabrizian:
@krishung5 The problem was that for decoupled models the `requested_output_names` is always empty since the request object is released before and it was the reason why the test was failing.

fpetrini15 (commented):
Is requested_output_names supposed to be a set? Or intentionally made a vector of strings in this case?

src/pb_stub.cc
```diff
@@ -846,11 +846,19 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
       module, "InferenceRequest")
       .def(
-          py::init<
-              const std::string&, uint64_t,
-              const std::vector<std::shared_ptr<PbTensor>>&,
-              const std::set<std::string>&, const std::string&, const int64_t,
-              const uint32_t>(),
+          py::init([](const std::string& request_id, uint64_t correlation_id,
+                      const std::vector<std::shared_ptr<PbTensor>>& inputs,
+                      const std::vector<std::string>& requested_output_names,
```
fpetrini15, line 851: Is requested_output_names supposed to be a set? Or intentionally made a vector of strings in this case?

src/pb_stub.cc
```diff
@@ -845,11 +846,19 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
+                      const std::string& model_name,
+                      const int64_t model_version, const uint32_t flags) {
+            std::set<std::string> requested_outputs;
+            for (auto& requested_output_name : requested_output_names) {
+              requested_outputs.emplace(requested_output_name);
+            }
+            return std::make_shared<InferRequest>(
+                request_id, correlation_id, inputs, requested_outputs,
+                model_name, model_version, flags);
+          }),
```
jbkyang-nvi, line 861: nit: `requested_output_names`

jbkyang-nvi: Also, why not use `std::set`? Is it because pybind doesn't support?

Tabrizian: @jbkyang-nvi requested_output_names will create a conflict because a variable with the same name already exists.

>why not use std::set? Is it because pybind doesn't support?

The reason is that it will be a breaking change to the user's Python model. Currently they are expected to pass a list. Changing it to a std::set will required them providing a Python set.

src/pb_stub.cc
Tabrizian: > Is requested_output_names supposed to be a set?

Yes, we intentionally made it a vector so that it works with Python lists.

fpetrini15 (dismissed):

src/pb_stub.cc
jbkyang-nvi: I would recommend changing the name to `requested_output_names_non_unique` instead. Do we have a plan of when we want to have the users change their model? If so, we should add a ticket to remind us of this workaround

tanmayv25 (approved):

src/pb_stub.cc
Tabrizian: I think the name is too long. If you have shorter variable name suggestions I'd be happy to change it. We neither have any plans or want users to change their Python model. This is only an internal change.

krishung5 (approved):

PR #178, opened 2022-07-22
Extend Logging to the python backend

fpetrini15:
Triton server logging functionality can now be invoked from python models. Here are a few example formats:
```python
logger = pb_utils.Logger
logger.log("Specific Msg!", logger.INFO)
logger.log_info("Info Msg!")
logger.log_warn("Warning Msg!")
logger.log_error("Error Msg!")
logger.log_verbose("Verbose Msg!")
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ any C++ code.
   - [Preprocessing](#preprocessing)
   - [Decoupled Models](#decoupled-models)
 - [Running with Inferentia](#running-with-inferentia)
+- [Logging to Triton](#logging-to-triton)
 - [Reporting problems, asking questions](#reporting-problems-asking-questions)
 
 ## Quick Start
@@ -1052,6 +1053,25 @@ You can find the complete example instructions in [examples/decoupled](examples/
 # Running with Inferentia
 
 Please see the [README.md](https://github.com/triton-inference-server/python_backend/tree/main/inferentia/README.md) located in the python_backend/inferentia sub folder.
+
+# Logging to Triton
+
+Model files that have imported ```triton_python_backend_utils``` can send log messages to the Triton server using any of the following access functions:
+```python
+import triton_python_backend_utils as pb_utils
+logger = pb_utils.Logger
+logger.log_info("Info Msg!")
+logger.log_warn("Warning Msg!")
+logger.log_error("Error Msg!")
+logger.log_verbose("Verbose Msg!")
+```
+Log messages can also be sent with their log-level type explcitiy specified:
+```python
+# log-level options: INFO, WARNING, ERROR, VERBOSE
+logger.log("Specific Msg!", logger.INFO)
+```
+
+Note that the Triton server's settings determine which log messages appear within the server log. For example, if a model attempts to log a verbose-level message, but Triton is not set to log verbose-level messages, it will not appear in the server log. For more information on Triton's log settings and how to adjust them dynamically, please see Triton's [logging extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_logging.md) document.
 
 # Reporting problems, asking questions
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ any C++ code.
   - [Preprocessing](#preprocessing)
   - [Decoupled Models](#decoupled-models)
 - [Running with Inferentia](#running-with-inferentia)
-- [Logging to Triton](#logging-to-triton)
+- [Logging](#logging)
 - [Reporting problems, asking questions](#reporting-problems-asking-questions)
 
 ## Quick Start
@@ -1054,24 +1054,33 @@ You can find the complete example instructions in [examples/decoupled](examples/
 
 Please see the [README.md](https://github.com/triton-inference-server/python_backend/tree/main/inferentia/README.md) located in the python_backend/inferentia sub folder.
 
-# Logging to Triton
+# Logging
 
-Model files that have imported ```triton_python_backend_utils``` can send log messages to the Triton server using any of the following access functions:
+Your Python model can log information using the following methods:
+
 ```python
 import triton_python_backend_utils as pb_utils
+...
 logger = pb_utils.Logger
 logger.log_info("Info Msg!")
 logger.log_warn("Warning Msg!")
 logger.log_error("Error Msg!")
 logger.log_verbose("Verbose Msg!")
 ```
-Log messages can also be sent with their log-level type explcitiy specified:
+Log messages can also be sent with their log-level explcitiy specified:
 ```python
 # log-level options: INFO, WARNING, ERROR, VERBOSE
 logger.log("Specific Msg!", logger.INFO)
 ```
+If no log-level is specified, this method will log INFO level messages.
 
-Note that the Triton server's settings determine which log messages appear within the server log. For example, if a model attempts to log a verbose-level message, but Triton is not set to log verbose-level messages, it will not appear in the server log. For more information on Triton's log settings and how to adjust them dynamically, please see Triton's [logging extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_logging.md) document.
+Note that the Triton server's settings determine which log messages appear
+within the server log. For example, if a model attempts to log a verbose-level
+message, but Triton is not set to log verbose-level messages, it will not
+appear in the server log. For more information on Triton's log settings and
+how to adjust them dynamically, please see Triton's 
+[logging extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_logging.md)
+documentation.
 
 # Reporting problems, asking questions
 
--- a/src/ipc_message.h
+++ b/src/ipc_message.h
@@ -49,7 +49,6 @@ typedef enum PYTHONSTUB_commandtype_enum {
   PYTHONSTUB_ResponseClose,
   PYTHONSTUB_AutoCompleteRequest,
   PYTHONSTUB_AutoCompleteResponse,
-  PYTHONSTUB_LogRequest,
 } PYTHONSTUB_CommandType;
 
 ///
--- a/src/ipc_message.h
+++ b/src/ipc_message.h
@@ -48,7 +48,8 @@ typedef enum PYTHONSTUB_commandtype_enum {
   PYTHONSTUB_ResponseSend,
   PYTHONSTUB_ResponseClose,
   PYTHONSTUB_AutoCompleteRequest,
-  PYTHONSTUB_AutoCompleteResponse
+  PYTHONSTUB_AutoCompleteResponse,
+  PYTHONSTUB_LogRequest,
 } PYTHONSTUB_CommandType;
 
 ///
--- a/src/ipc_message.h
+++ b/src/ipc_message.h
@@ -48,7 +48,7 @@ typedef enum PYTHONSTUB_commandtype_enum {
   PYTHONSTUB_ResponseSend,
   PYTHONSTUB_ResponseClose,
   PYTHONSTUB_AutoCompleteRequest,
-  PYTHONSTUB_AutoCompleteResponse,
+  PYTHONSTUB_AutoCompleteResponse
 } PYTHONSTUB_CommandType;
 
 ///
--- a/src/pb_log.cc
+++ b/src/pb_log.cc
@@ -119,4 +119,3 @@ PbLogShm::LogMessage()
 }
 
 }}}  // namespace triton::backend::python
-
--- a/src/pb_log.cc
+++ b/src/pb_log.cc
@@ -118,5 +118,4 @@ PbLogShm::LogMessage()
   return log_container_shm_ptr_;
 }
 
-
 }}}  // namespace triton::backend::python
--- a/src/pb_log.cc
+++ b/src/pb_log.cc
@@ -1,0 +1,122 @@
+// Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of NVIDIA CORPORATION nor the names of its
+//    contributors may be used to endorse or promote products derived
+//    from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "pb_log.h"
+
+namespace triton { namespace backend { namespace python {
+
+PbLog::PbLog(
+    const std::string& filename, uint32_t line, const std::string& message,
+    LogLevel level)
+    : filename_(filename), line_(line), message_(message), level_(level)
+{
+}
+
+const std::string&
+PbLog::Filename()
+{
+  return filename_;
+}
+const std::string&
+PbLog::Message()
+{
+  return message_;
+}
+const LogLevel&
+PbLog::Level()
+{
+  return level_;
+}
+const uint32_t&
+PbLog::Line()
+{
+  return line_;
+}
+
+PbLogShm::PbLogShm(
+    AllocatedSharedMemory<LogSendMessage>& log_container_shm,
+    std::unique_ptr<PbString>& filename, std::unique_ptr<PbString>& message)
+    : log_container_shm_(std::move(log_container_shm)),
+      filename_pb_string_(std::move(filename)),
+      message_pb_string_(std::move(message))
+{
+  log_container_shm_ptr_ = log_container_shm_.data_.get();
+  log_container_shm_ptr_->filename = filename_pb_string_->ShmHandle();
+  log_container_shm_ptr_->log_message = message_pb_string_->ShmHandle();
+}
+
+std::unique_ptr<PbLogShm>
+PbLogShm::Create(
+    std::unique_ptr<SharedMemoryManager>& shm_pool, const std::string& filename,
+    const uint32_t& line, const std::string& message, const LogLevel& level)
+{
+  std::unique_ptr<PbString> file_name = PbString::Create(shm_pool, filename);
+  std::unique_ptr<PbString> log_message = PbString::Create(shm_pool, message);
+  AllocatedSharedMemory<LogSendMessage> log_send_message =
+      shm_pool->Construct<LogSendMessage>();
+
+  LogSendMessage* send_message_payload = log_send_message.data_.get();
+  new (&(send_message_payload->log_mu)) bi::interprocess_mutex;
+  new (&(send_message_payload->log_cv)) bi::interprocess_condition;
+  send_message_payload->line = line;
+  send_message_payload->level = level;
+
+  return std::unique_ptr<PbLogShm>(
+      new PbLogShm(log_send_message, file_name, log_message));
+}
+
+std::unique_ptr<PbLog>
+PbLogShm::LoadFromSharedMemory(
+    std::unique_ptr<SharedMemoryManager>& shm_pool,
+    bi::managed_external_buffer::handle_t handle)
+{
+  AllocatedSharedMemory<LogSendMessage> log_container_shm =
+      shm_pool->Load<LogSendMessage>(handle);
+  std::unique_ptr<PbString> pb_string_filename = PbString::LoadFromSharedMemory(
+      shm_pool, log_container_shm.data_->filename);
+  const std::string& filename = pb_string_filename->String();
+  uint32_t line = log_container_shm.data_->line;
+  std::unique_ptr<PbString> pb_string_msg = PbString::LoadFromSharedMemory(
+      shm_pool, log_container_shm.data_->log_message);
+  const std::string& message = pb_string_msg->String();
+  LogLevel level = log_container_shm.data_->level;
+  return std::unique_ptr<PbLog>(new PbLog(filename, line, message, level));
+}
+
+bi::managed_external_buffer::handle_t
+PbLogShm::ShmHandle()
+{
+  return log_container_shm_.handle_;
+}
+
+LogSendMessage*
+PbLogShm::LogMessage()
+{
+  return log_container_shm_ptr_;
+}
+
+
+}}}  // namespace triton::backend::python
--- a/src/pb_log.cc
+++ b/src/pb_log.cc
@@ -119,3 +119,4 @@ PbLogShm::LogMessage()
 }
 
 }}}  // namespace triton::backend::python
+
--- a/src/pb_log.h
+++ b/src/pb_log.h
@@ -35,15 +35,13 @@ class PbLog {
  public:
   PbLog(
       const std::string& filename, uint32_t line, const std::string& message,
-      LogLevel level, uint32_t verbosity)
-      : filename_(filename), line_(line), message_(message), level_(level),
-        verbosity_(verbosity)
+      LogLevel level)
+      : filename_(filename), line_(line), message_(message), level_(level)
   {
   }
   const std::string& Filename() { return filename_; };
   const std::string& Message() { return message_; };
   const LogLevel& Level() { return level_; };
-  const uint32_t& Verbosity() { return verbosity_; };
   const uint32_t& Line() { return line_; };
 
  private:
@@ -51,14 +49,12 @@ class PbLog {
   uint32_t line_;
   std::string message_;
   LogLevel level_;
-  uint32_t verbosity_;
 };
 struct LogMsgShm {
   bi::managed_external_buffer::handle_t filename;
   uint32_t line;
   bi::managed_external_buffer::handle_t logMsg;
   LogLevel level;
-  uint32_t verbosity;
   bool is_stub_turn;
   bi::managed_external_buffer::handle_t response_mutex;
   bi::managed_external_buffer::handle_t response_cond;
--- a/src/pb_log.h
+++ b/src/pb_log.h
@@ -59,8 +59,7 @@ class PbLogShm {
       bi::managed_external_buffer::handle_t handle);
   bi::managed_external_buffer::handle_t ShmHandle();
   LogSendMessage* LogMessage();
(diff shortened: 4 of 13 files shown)


rmccorm4:
Is `__line__()` a thing in python? Or do users have to implement something to get the line number for each call?

If users have to implement something for each call, I think this could be done in some kind of function decorator or wrapper to do that for them for a better user experience.

src/pb_stub.cc
```diff
@@ -837,6 +862,130 @@ Stub::GetOrCreateInstance()
...
+    bi::scoped_lock<bi::interprocess_mutex> guard{send_message_payload->log_mu};
+    SendIPCLogMessage(log_request_msg);
+    while (!send_message_payload->waiting_on_stub) {
+      send_message_payload->log_cv.wait(guard);
+    }
+  }
+}
+
+void
+Logger::log(
```
rmccorm4, line 946: Is `__line__()` a thing in python? Or do users have to implement something to get the line number for each call?

If users have to implement something for each call, I think this could be done in some kind of function decorator or wrapper to do that for them for a better user experience.

---

I don't know how robust this is, but from quick googling, an example decorator might look like this. It can probably be done more cleanly if looked into further:
```python
import sys 
from inspect import currentframe, getframeinfo

def log_decorator(func):
  def wrapper(*args, **kwargs):
      caller_frame = currentframe().f_back
      caller_info = getframeinfo(caller_frame)
      caller_line_number = caller_info.lineno
      caller_filename = caller_info.filename
      print(f"{caller_filename}::{caller_line_number}", end=" ")
      return func(*args, **kwargs)

  return wrapper 

def log_full(filename, line, message, level, verbosity):
  print(f"{filename}::{line} [{level}] - {message}")

@log_decorator
def log_wrapped(message, level, verbosity):
  print(f"[{level}] - {message}")
  
if __name__ == "__main__":
  msg = "Hello"
  level = "INFO"
  verbosity = 0
  lineno = 28

  log_full(__file__, lineno, msg, level, verbosity) # depends on user to get their line number everytime / define it themself
  log_wrapped(msg, level, verbosity) # line number retrieved automatically without user intervention
```

Note these two functions achieve the same output in this example:
```
rmccormick@rmccormick-dt:~$ python3 test.py 
test.py::28 [INFO] - Hello
test.py::29 [INFO] - Hello
```

fpetrini15: I took a look into decorator implementation, but found that a simple interface function in triton_python_backend_utils should suffice:

`class level(Enum):
    INFO = 0,
    WARNINGS = 1,
    ERRORS = 2,
    VERBOSE = 3

def log(message, log_level=level.INFO, verbosity=0):
    caller_frame = currentframe().f_back
    caller_info = getframeinfo(caller_frame)
    caller_line_number = caller_info.lineno
    caller_filename = caller_info.filename
    if verbosity != 0 and log_level != level.VERBOSE:
        print(str(verbosity) + " " + str(log_level))
        print("Warning at " + caller_filename + " " + str(caller_line_number) + ": Verbosity set but log level is not VERBOSE")
        return
    logger = Logger()
    if(log_level == level.INFO):
        logger.log_info(caller_filename, caller_line_number, message)
    elif(log_level == level.WARNINGS):
        logger.log_warn(caller_filename, caller_line_number, message)
    elif(log_level == level.ERRORS):
        logger.log_error(caller_filename, caller_line_number, message)
    elif(log_level == level.VERBOSE):
        logger.log_verbose(caller_filename, caller_line_number, message, verbosity)`

With this interface function we can also a achieve a greater level of abstraction. Now instead of specifying one of multiple log functions and instantiating and object, the user simply has a single log interface function and we handle the rest internally: 

`pb_utils.log("Hello!", pb_utils.level.INFO, verbosity=0) # Okay
pb_utils.log(message="Verbose Msg!", verbosity=1) # Error, verbosity but not verbose level
pb_utils.log(message="Verbose Msg Lvl 2!", log_level=pb_utils.level.VERBOSE, verbosity=2) # okay
pb_utils.log("Hello!") # okay --> will default to INFO log
pb_utils.log(message="Hello!", log_level=pb_utils.level.ERRORS) #okay
pb_utils.log("Info Msg!", pb_utils.level.INFO) #okay
pb_utils.log("Warn Msg!", pb_utils.level.WARNINGS) #okay
pb_utils.log("Error Msg!", pb_utils.level.ERRORS) #okay`

src/python_be.h
```diff
@@ -261,6 +261,8 @@ class ModelInstanceState : public BackendModelInstance {
   std::vector<intptr_t> closed_requests_;
   std::mutex closed_requests_mutex_;
 
+  std::thread log_monitor_;
+  bool log_thread_;
```
GuanLuo, line 265: @Tabrizian can this handling be moved to backend level? All Python stubs will send logs to the same queue and avoid spanning additional thread per instance.

src/resources/triton_python_backend_utils.py
```diff
@@ -497,6 +500,26 @@ def add_output(self, output):
...
+    caller_filename = caller_info.filename
+    logger = Logger()
+    if(log_level == level.INFO):
+        logger.log_info(caller_filename, caller_line_number, message)
+    elif(log_level == level.WARNINGS):
+        logger.log_warn(caller_filename, caller_line_number, message)
+    elif(log_level == level.ERRORS):
+        logger.log_error(caller_filename, caller_line_number, message)
+    elif(log_level == level.VERBOSE):
+        logger.log_verbose(caller_filename, caller_line_number, message)
```
GuanLuo, line 522: Why not just call `log()`

src/pb_stub.cc
GuanLuo: Can you attach some example logs

rmccorm4: Where'd `verbosity` go? Are we only supporting a single verbosity level for python backend?

fpetrini15: The verbosity option was removed. It will be added back if there is a request for it:
https://github.com/triton-inference-server/core/pull/101

fpetrini15: **Example Logs:**
These lines of code:
```
pb_utils.log("Info Msg!", pb_utils.Logger.INFO) 
pb_utils.log(message="Verbose Msg!", level=pb_utils.Logger.VERBOSE)
pb_utils.log(message="Error Msg!", level=pb_utils.Logger.ERRORS) 
pb_utils.log_info("Info Msg!") 
pb_utils.log_warn("Warn Msg!") 
pb_utils.log_error("Error Msg!")
pb_utils.log_verbose =("Verbose Msg!")
```
will appear in the logs like this (verbose appears too if set):
```
I0725 23:02:15.542552 167 model.py:54] Info Msg!
E0725 23:02:15.542828 167 model.py:56] Error Msg!
I0725 23:02:15.542989 167 model.py:57] Info Msg!
W0725 23:02:15.543126 167 model.py:58] Warn Msg!
E0725 23:02:15.543398 167 model.py:59] Error Msg!
```

Tabrizian (changes requested):

src/pb_stub.cc
```diff
@@ -837,6 +862,96 @@ Stub::GetOrCreateInstance()
...
+Stub::ServiceLogRequests()
+{
+  while (log_thread_) {
+    if (log_request_buffer.size() != 0) {
+      PbLog* log_request = log_request_buffer.front();
+      log_request_buffer.pop();
+      SendLogMessage(
+          log_request->Filename(), log_request->Line(), log_request->Message(),
+          log_request->Level());
+      delete log_request;
```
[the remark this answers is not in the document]
fpetrini15: Used shared_ptr since we are using it in the queue. However, unique_ptr is used for the allocated memory that is later pushed onto the LogQueue.

src/ipc_message.h
```diff
@@ -48,7 +48,7 @@ typedef enum PYTHONSTUB_commandtype_enum {
   PYTHONSTUB_ResponseSend,
   PYTHONSTUB_ResponseClose,
   PYTHONSTUB_AutoCompleteRequest,
-  PYTHONSTUB_AutoCompleteResponse
+  PYTHONSTUB_AutoCompleteResponse,
```
Tabrizian, line 51: Remove comma

src/pb_log.cc
```diff
@@ -0,0 +1,122 @@
...
+}
+
+LogSendMessage*
+PbLogShm::LogMessage()
+{
+  return log_container_shm_ptr_;
+}
+
+
+}}}  // namespace triton::backend::python
```
Tabrizian, line 122: Add newline in the end.

src/pb_log.h
```diff
@@ -0,0 +1,75 @@
...
+  static std::unique_ptr<PbLogShm> Create(
+      std::unique_ptr<SharedMemoryManager>& shm_pool,
+      const std::string& filename, const uint32_t& line,
+      const std::string& message, const LogLevel& level);
+  static std::unique_ptr<PbLog> LoadFromSharedMemory(
+      std::unique_ptr<SharedMemoryManager>& shm_pool,
+      bi::managed_external_buffer::handle_t handle);
+  bi::managed_external_buffer::handle_t ShmHandle();
+  LogSendMessage* LogMessage();
+  AllocatedSharedMemory<LogSendMessage> LogMessageShm();
```
Tabrizian, line 62: You don't need to expose the underlying LogMessageShm object. You can instead add getters to expose all the fields that you need.

src/pb_stub.cc
```diff
@@ -934,6 +1026,18 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
           "send", &ResponseSender::Send, py::arg("response") = nullptr,
           py::arg("flags") = 0);
 
+  py::class_<Logger> logger(module, "Logger");
+  py::enum_<LogLevel>(logger, "LogLevel")
+      .value("INFO", LogLevel::INFO)
+      .value("WARNINGS", LogLevel::WARNINGS)
```
Tabrizian, line 1032: WARNINGS -> WARNING
ERRORS -> ERROR

src/resources/triton_python_backend_utils.py
```diff
@@ -498,6 +500,36 @@ def add_output(self, output):
         self._model_config["output"].append(output)
 
 
+def log(message, level, nested_call=False):
```
Tabrizian, line 503: Now that there is a way to extract the line number and file information, can we move this logic to the C++ side?

src/pb_stub.cc
```diff
@@ -837,6 +857,78 @@ Stub::GetOrCreateInstance()
...
+Stub::TerminateLogRequestThread()
+{
+  log_thread_ = false;
+  log_monitor_.join();
+}
+
+void
+Stub::EnqueueLogRequest(std::shared_ptr<PbLog> log_ptr)
+{
+  log_request_buffer.push(log_ptr);
```
Tabrizian, line 877: You need a synchronization logic to protect pushing messages to this queue. Otherwise, if multiple threads are pushing to this queue simultaneously there will be a race condition.

src/pb_stub.cc
```diff
@@ -837,6 +857,78 @@ Stub::GetOrCreateInstance()
...
+}
+
+void
+Stub::ServiceLogRequests()
+{
+  while (log_thread_) {
+    if (log_request_buffer.size() != 0) {
```
Tabrizian, line 884: You need the synchronization logic here too.

src/pb_stub.cc
```diff
@@ -837,6 +857,78 @@ Stub::GetOrCreateInstance()
...
+
+void
+Stub::TerminateLogRequestThread()
+{
+  log_thread_ = false;
+  log_monitor_.join();
+}
+
+void
+Stub::EnqueueLogRequest(std::shared_ptr<PbLog> log_ptr)
```
Tabrizian, line 875: You don't need to use a shared_ptr here. You should use unique_ptr instead.

src/resources/triton_python_backend_utils.py
```diff
@@ -498,6 +500,27 @@ def add_output(self, output):
...
+def log_warn(message):
+    log(message, Logger().WARNING, True)
+
+
+def log_error(message):
+    log(message, Logger().ERROR, True)
+
+
+def log_verbose(message):
+    log(message, Logger().VERBOSE, True)
```
rmccorm4, line 521: It looks like we're instantiating a new logger for every single log and log().LEVEL here. Can there be some kind of global/singleton instead instantiated once?

src/pb_stub.cc
```diff
@@ -837,6 +857,92 @@ Stub::GetOrCreateInstance()
...
+Stub::EnqueueLogRequest(std::unique_ptr<PbLog>& log_ptr)
+{
+  std::unique_lock<std::mutex> guard{log_message_mutex_};
+  log_request_buffer.push(std::move(log_ptr));
+}
+
+void
+Stub::ServiceLogRequests()
+{
+  while (log_thread_) {
```
Tabrizian, line 884: This is a busy thread loop which significantly increases the CPU utilization. You need to either use semaphore or a condition variable to avoid busy waiting. If there are no logs, this is a while loop that is always true and just keeps checking the condition.

src/pb_stub.cc
Tabrizian: You should only check the size of the array when you have acquired the mutex.

src/pb_stub.h
```diff
@@ -74,6 +77,10 @@ class Logger {
   // Log a message.
   void Log(const std::string& msg) { std::cerr << msg << std::endl; }
```
Tabrizian, line 78: I think we probably need to redirect all the existing logging in Python backend to the server too right? Otherwise, the logging info in the stub process doesn't obey the server's logging settings. You should also make all the `log_<level>` function in the triton_python_backend_utils part of this class.

src/resources/triton_python_backend_utils.py
```diff
@@ -497,6 +498,27 @@ def add_output(self, output):
 
         self._model_config["output"].append(output)
 
+logger = Logger()
+
+def log(message, level, nested_call=False):
+    logger.log(message, level, nested_call)
+
+
+def log_info(message):
```
Tabrizian, line 507: Move all this functions to C++.

src/pb_stub.cc
```diff
@@ -934,6 +1040,18 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
           py::arg("flags") = 0);
 
+  py::class_<Logger> logger(module, "Logger");
+  py::enum_<LogLevel>(logger, "LogLevel")
+      .value("INFO", LogLevel::INFO)
+      .value("WARNING", LogLevel::WARNING)
+      .value("ERROR", LogLevel::ERROR)
+      .value("VERBOSE", LogLevel::VERBOSE)
+      .export_values();
+  logger.def(py::init<>());
```
Tabrizian, line 1050: Don't expose the constructor to the Python side. Instead, try to make the logger class in C++ a singleton and only expose the singleton instance to C++. The discussion here might be helpful: 
https://github.com/pybind/pybind11/issues/1418

src/pb_stub.h
fpetrini15: Migrated existing LOG_INFO macro to send messages through the new infrastructure. I also created LOG_WARN, LOG_ERROR, and LOG_VERBOSE macros so we can send future logs with setting-level precision.

src/resources/triton_python_backend_utils.py
fpetrini15: The python client now has access to a singleton from C++, but no ownership. The client can now call log functions from their models in the following fashion:
```
logger = pb_utils.Logger.instance()
logger.log("Hello!", logger.INFO)
logger.log_info("Info Msg!")
logger.log_warn("Warning Msg!")
logger.log_error("Error Msg!")
logger.log_verbose("Verbose Msg!")
```

src/pb_stub.h
Tabrizian: Thank you!

src/resources/triton_python_backend_utils.py
Tabrizian: I think the API needs some improvements. How about not exposing `instance()` to the user? We set the pb_utils.Logger to the instance.

Then the API would look like:

```
logger = pb_utils.Logger
logger.log("Specific Msg!", logger.INFO)
logger.log_info("Info Msg!")
logger.log_warn("Warning Msg!")
logger.log_error("Error Msg!")
logger.log_verbose("Verbose Msg!")
```

src/pb_stub.cc
```diff
@@ -812,16 +820,28 @@ Stub::SendIPCMessage(std::unique_ptr<IPCMessage>& ipc_message)
...
+
 Stub::~Stub()
 {
   {
     py::gil_scoped_acquire acquire;
     model_instance_ = py::none();
   }
-
+  while (log_request_buffer.size() != 0) {
+    log_request_buffer.pop();
```
Tabrizian, line 839: Why do we need to pop requests here? Shouldn't the Triton process be responsible for popping messages?

src/pb_stub.cc
```diff
@@ -837,6 +857,135 @@ Stub::GetOrCreateInstance()
...
+
+void
+Stub::TerminateLogRequestThread()
+{
+  log_thread_ = false;
+  {
+    bi::scoped_lock<bi::interprocess_mutex> guard{log_message_mutex_};
+    log_request_buffer.push(DUMMY_MESSAGE);
+  }
+  log_message_cv_.notify_all();
```
Tabrizian, line 875: The notification must be sent when the mutex is acquired.

src/pb_stub.cc
```diff
@@ -837,6 +857,135 @@ Stub::GetOrCreateInstance()
...
+  log_monitor_.join();
+}
+void
+Stub::EnqueueLogRequest(std::unique_ptr<PbLog>& log_ptr)
+{
+  {
+    bi::scoped_lock<bi::interprocess_mutex> guard{log_message_mutex_};
+    log_request_buffer.push(std::move(log_ptr));
+  }
+  log_message_cv_.notify_one();
```
Tabrizian, line 885: Same

src/pb_stub.cc
```diff
@@ -837,6 +857,135 @@ Stub::GetOrCreateInstance()
...
+}
+
+void
+Stub::ServiceLogRequests()
+{
+  while (log_thread_) {
+    bi::scoped_lock<bi::interprocess_mutex> guard{log_message_mutex_};
```
Tabrizian, line 892: Since this mutex is not interprocess, you should use the std library mutex and condition variable. https://en.cppreference.com/w/cpp/thread/mutex

Tabrizian (changes requested):

src/pb_stub.cc
rmccorm4: Is this true? ref:https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all

> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock.

fpetrini15: I was following the convention in the above link. I attempted to place the notification within the scope of the lock and it ends up stalling the program.

Tabrizian: Thanks for the link Ryan. Looks like it is not required. Resolving this thread.

Tabrizian (commented):
@fpetrini15 You also need to a documentation section for this PR too but it can be part of a follow up PR.

src/pb_stub.cc
fpetrini15: This queue is separate from the inter-process queue. This buffer only exists on the stub side and allows for a high volume of log requests to be stored without impacting performance. The thread I created reads from this buffer, processes the info and then places it on the Triton queue. Without it, the stub process would have to wait for the message synchronization to take place during the transfer.

src/resources/triton_python_backend_utils.py
fpetrini15: It's been tricky attempting to bind a property with the same name as the class. Is it appropriate to tie the instance to a non-function property of the class (e.g. `pb_utils.Logger.Log`)? Bound like so: `def_property_readonly_static("Log", [](py::object) { return &Logger::Instance(); });`

src/python_be.h
```diff
@@ -360,5 +360,11 @@ class ModelInstanceState : public BackendModelInstance {
 
   // Model instance stub
   std::unique_ptr<StubLauncher>& Stub() { return model_instance_stub_; }
+
+  // Start the log monitor thread
+  void TerminateLogMonitor();
+
+  // Stop the log monitor thread
+  void StartLogMonitor();
```
rmccorm4, line 368: Descriptions for start/stop are reversed here

fpetrini15:
Latest changes pass all L0_backend_python tests. I am planning a new section to the README.md file of this repo so I can include it in the same ticket and we can avoid creating a new one.

README.md
```diff
@@ -1053,6 +1054,25 @@ You can find the complete example instructions in [examples/decoupled](examples/
 
 Please see the [README.md](https://github.com/triton-inference-server/python_backend/tree/main/inferentia/README.md) located in the python_backend/inferentia sub folder.
 
+# Logging to Triton
```
Tabrizian, line 1057: Remove "to Triton"

README.md
```diff
@@ -1053,6 +1054,25 @@ You can find the complete example instructions in [examples/decoupled](examples/
...
+
+Model files that have imported ```triton_python_backend_utils``` can send log messages to the Triton server using any of the following access functions:
```
Tabrizian, line 1059: Keep the number of characters in each line below 80.

Tabrizian, line 1059: "Your Python model can log information using the following methods"

README.md
```diff
@@ -1053,6 +1054,25 @@ You can find the complete example instructions in [examples/decoupled](examples/
...
+```python
+import triton_python_backend_utils as pb_utils
```
Tabrizian, line 1061: Add "..." here.

README.md
```diff
@@ -1053,6 +1054,25 @@ You can find the complete example instructions in [examples/decoupled](examples/
...
+logger = pb_utils.Logger
+logger.log_info("Info Msg!")
+logger.log_warn("Warning Msg!")
+logger.log_error("Error Msg!")
+logger.log_verbose("Verbose Msg!")
+```
+Log messages can also be sent with their log-level type explcitiy specified:
```
Tabrizian, line 1068: Remove "type". Also, mention the default level if no log-level is specified.

README.md
```diff
@@ -1053,6 +1054,25 @@ You can find the complete example instructions in [examples/decoupled](examples/
...
+```python
+# log-level options: INFO, WARNING, ERROR, VERBOSE
+logger.log("Specific Msg!", logger.INFO)
+```
+
+Note that the Triton server's settings determine which log messages appear within the server log. For example, if a model attempts to log a verbose-level message, but Triton is not set to log verbose-level messages, it will not appear in the server log. For more information on Triton's log settings and how to adjust them dynamically, please see Triton's [logging extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_logging.md) document.
```
Tabrizian, line 1074: Keep the number of characters below 80.

Tabrizian, line 1074: document -> documentation

src/pb_log.cc
```diff
@@ -0,0 +1,122 @@
...
+}
+
+LogSendMessage*
+PbLogShm::LogMessage()
+{
+  return log_container_shm_ptr_;
+}
+
+}}}  // namespace triton::backend::python
+
```
Tabrizian, line 122: Remove extra new line.

src/pb_stub.cc
```diff
@@ -464,6 +471,8 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
   py::setattr(
       python_backend_utils, "InferenceResponse",
       c_python_backend_utils.attr("InferenceResponse"));
+  py::setattr(
+      python_backend_utils, "Logger", c_python_backend_utils.attr("Logger"));
```
Tabrizian, line 475: I think you can remove all these lines here. They seem to be already set in `StubSetup`.

src/pb_stub.cc
```diff
@@ -837,6 +864,183 @@ Stub::GetOrCreateInstance()
...
+// Called internally (.e.g. LOG_ERROR << "Error"; )
+void
+Logger::Log(
+    const std::string& filename, uint32_t lineno, LogLevel level,
+    const std::string& message)
+{
+  std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
+  // If the log monitor service is not active yet, format
+  // and pass messages to cerr
+  if (!stub->LogServiceActive()) {
```
Tabrizian, line 981: I was wondering why LogService is not active. Is there anyway to make it active before we call the user's `Initialize` function.

src/pb_stub.h
```diff
@@ -66,51 +69,84 @@ namespace triton { namespace backend { namespace python {
...
+  {
+    static Logger instance;
+    return instance;
+  }
+
+  /// Should not be cloneable
+  Logger(Logger const&) = delete;
+
+  /// Should not be assignable
+  void operator=(Logger const&) = delete;
```
Tabrizian, line 115: There is `DISALLOW_COPY_AND_ASSIGN` macro that you can use.

src/python_be.cc
```diff
@@ -1818,10 +1898,13 @@ TRITONBACKEND_ModelInstanceExecute(
...
           TRITONSERVER_LOG_ERROR,
           "Stub process is unhealthy and it will be restarted.");
+      instance_state->TerminateLogMonitor();
       instance_state->Stub()->KillStubProcess();
-      LOG_IF_ERROR(
-          instance_state->Stub()->Launch(),
-          "Failed to restart the stub process.");
+      TRITONSERVER_Error* err = instance_state->Stub()->Launch();
+      if(err == nullptr) {
+        instance_state->StartLogMonitor();
```
Tabrizian, line 1905: Can this be part of `instance_state->Stub()->Launch()`?

src/python_be.cc
```diff
@@ -397,7 +398,7 @@ ModelInstanceState::LaunchStubProcess()
       TRITONSERVER_InstanceGroupKindString(Kind()));
   RETURN_IF_ERROR(Stub()->Initialize(model_state));
   RETURN_IF_ERROR(Stub()->Launch());
-
+  StartLogMonitor();
```
Tabrizian, line 401: Make this part of `Stub()->Launch()`.

src/pb_stub.cc
fpetrini15: There are 2 critical periods where log service is not active, startup and shutdown. During startup, we can only start the logging threads after the stub launcher has allocated the log message queue. The reason I wait until the stub is initialized before activating the log thread is because activating it before assumes that the launch will be successful and that the backend will begin listening for messages from the stub. This is not a safe guarantee and will result in a deadlock if a message is pushed onto the queue but the stub fails to launch. This is the same reason I terminate the stub log thread before any stub termination messages are sent. The Python backend must terminate its log monitor prior to terminating the stub because the stub termination process will deallocate the log message queue. If we terminate the stub log thread last, we will also encounter a deadlock for the same reason. 

I figure the safest handling would be to have the condition variables do a wait_for and check their predicates on timeout. But I was worried this would degrade performance, especially if the user does not post any log messages.

src/python_be.cc
fpetrini15: Will this be confusing later on that we are terminating/restarting the log thread in one file but starting it in another?

src/pb_stub.cc
fpetrini15: Confirmed to still be working after removal.

src/pb_stub.cc
Tabrizian: I see. Thanks for the detailed response!

Tabrizian: Just wanted to make sure that the user can use logging in `initialize` and `finalize` parts of their Python model.

src/python_be.cc
Tabrizian: Maybe we can combine both `Terminate` and `Start` in the in `instance_state->Stub()->Launch()`? Whenever `Stub()->Lanuch()` is called we will terminate the log monitor first and then Start it.

src/pb_stub.cc
fpetrini15: The user can access the pybind functions with no issues when logging during initialize and finalize. However, the messages will be logged using cerr as opposed to triton logging as the threads will not have started up yet/ will have already terminated.

src/python_be.cc
Tabrizian: Discussed offline. Resolving.

src/pb_stub.cc
fpetrini15: Resolving, this is now outdated. User can log messages to Triton within initialize and finalize as well now.

src/pb_stub.cc
```diff
@@ -1253,11 +1247,11 @@ main(int argc, char** argv)
   bool finalize = false;
   while (true) {
     if (finalize) {
+      stub->Finalize();
       // Need check or may recieve not joinable error
```
rmccorm4, line 1251:
```suggestion
      // Need check or may receive not joinable error
```

Tabrizian (approved):

rmccorm4 (approved):
Very nice PR :rocket:

PR #181, opened 2022-08-16
Implement GetBackendAttribute to provide instance kind hint

GuanLuo:
core: https://github.com/triton-inference-server/core/pull/109
example:
if model has below config and build with ENABLE_GPU=OFF:
```
name: "add_sub"
backend: "python"

input [
...
]
output [
...
]
# not providing instance group
#instance_group [{ kind: KIND_CPU }]
```
Before this change, Triton will always fill KIND_GPU on GPU system and result in:
```
...
I0816 00:00:09.814568 21489 python_be.cc:1767] TRITONBACKEND_ModelInstanceInitialize: add_sub (GPU device 0)
E0816 00:00:09.814760 21489 model_lifecycle.cc:596] failed to load 'add_sub' version 1: Internal: GPU instances not supported
...
```
After:
```
...
I0815 23:57:16.036978 21355 python_be.cc:1767] TRITONBACKEND_ModelInstanceInitialize: add_sub (CPU device 0)
I0815 23:57:16.209357 21355 model_lifecycle.cc:693] successfully loaded 'add_sub' version 1
...
````
And config API will return:
`... "instance_group":[{"name":"add_sub","kind":"KIND_CPU","count":1,"gpus":[],"secondary_devices":[],"profile":[],"passive":false,"host_policy":""} ...`

### DIFF
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -1908,5 +1908,24 @@ TRITONBACKEND_ModelInstanceFinalize(TRITONBACKEND_ModelInstance* instance)
   return nullptr;
 }
 
+TRITONSERVER_Error*
+TRITONBACKEND_GetBackendAttribute(
+    TRITONBACKEND_Backend* backend,
+    TRITONBACKEND_BackendAttribute* backend_attributes)
+{
+  LOG_MESSAGE(
+      TRITONSERVER_LOG_VERBOSE,
+      "TRITONBACKEND_GetBackendAttribute: setting attributes");
+#ifdef TRITON_ENABLE_GPU
+  RETURN_IF_ERROR(TRITONBACKEND_BackendAttributeAddPreferredInstanceGroup(backend_attributes,
+      TRITONSERVER_INSTANCEGROUPKIND_GPU, 0, nullptr, 0));
+#else
+  RETURN_IF_ERROR(TRITONBACKEND_BackendAttributeAddPreferredInstanceGroup(backend_attributes,
+      TRITONSERVER_INSTANCEGROUPKIND_CPU, 0, nullptr, 0));
+#endif
+
+  return nullptr;
+}
+
 }  // extern "C"
 }}}  // namespace triton::backend::python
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/python_be.cc [hunk 1]
```diff
@@ -1908,5 +1908,24 @@ TRITONBACKEND_ModelInstanceFinalize(TRITONBACKEND_ModelInstance* instance)
...
+TRITONBACKEND_GetBackendAttribute(
+    TRITONBACKEND_Backend* backend,
+    TRITONBACKEND_BackendAttribute* backend_attributes)
+{
+  LOG_MESSAGE(
+      TRITONSERVER_LOG_VERBOSE,
+      "TRITONBACKEND_GetBackendAttribute: setting attributes");
+#ifdef TRITON_ENABLE_GPU
+  RETURN_IF_ERROR(TRITONBACKEND_BackendAttributeAddPreferredInstanceGroup(backend_attributes,
+      TRITONSERVER_INSTANCEGROUPKIND_GPU, 0, nullptr, 0));
```
Tabrizian, line 1921: Can you add inline comments for these arguments?

tanmayv25:
I was wondering shouldn't this be default attributes for other backends as well? I mean if a backend is built without GPU support, shouldn't it let the core know that it needs CPU instance?

src/python_be.cc
```diff
@@ -1908,5 +1908,28 @@ TRITONBACKEND_ModelInstanceFinalize(TRITONBACKEND_ModelInstance* instance)
...
 
+TRITONSERVER_Error*
+TRITONBACKEND_GetBackendAttribute(
+    TRITONBACKEND_Backend* backend,
+    TRITONBACKEND_BackendAttribute* backend_attributes)
+{
+  LOG_MESSAGE(
+      TRITONSERVER_LOG_VERBOSE,
+      "TRITONBACKEND_GetBackendAttribute: setting attributes");
+  // Specify different preferred instance kind baseds on backend compatibility,
```
tanmayv25, line 1919: *based on

GuanLuo:
> I was wondering shouldn't this be default attributes for other backends as well? I mean if a backend is built without GPU support, shouldn't it let the core know that it needs CPU instance?

Can you create a ticket to track this. I don't want to span too many repos and it can be done along with moving the default instance count logic to specific backends.

tanmayv25 (approved):

PR #184, opened 2022-09-08
PYBE Logging Seg Fault Fix

fpetrini15:
A segmentation fault would occur in the python stub when Log messages were sent prior to the stub being initialized. This was due to improper logger clean up and the fact that Logger a logger class method was instantiating an instance of Stub before it was initialized. Stub launch was checked with valgrind and resulted in 0 errors. Further, after running tests, no core dumps were produced.

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -858,13 +858,13 @@ Stub::LaunchLogRequestThread()
 {
   log_thread_ = true;
   log_monitor_ = std::thread(&Stub::ServiceLogRequests, this);
-  logger->SetBackendLoggingActive(true);
+  Logger::GetOrCreateInstance()->SetBackendLoggingActive(true);
 }
 
 void
 Stub::TerminateLogRequestThread()
 {
-  logger->SetBackendLoggingActive(false);
+  Logger::GetOrCreateInstance()->SetBackendLoggingActive(false);
   {
     std::lock_guard<std::mutex> guard{log_message_mutex_};
     log_request_buffer_.push(DUMMY_MESSAGE);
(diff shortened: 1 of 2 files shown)


src/pb_stub.cc
```diff
@@ -1241,6 +1256,7 @@ main(int argc, char** argv)
             // messages to the backend ASAP.
             if (stub->LogServiceActive()) {
               stub->TerminateLogRequestThread();
+              logger->SetBackendLoggingActive(false);
```
Tabrizian, line 1259: Can this be part of `TerminateLogRequestThread`?

fpetrini15: Part of the reason the reason this seg fault existed is that the Logger and Stub objects were interleaved and were being instantiated at unintended times. Normally, the way we handle it would be fine, but the GIL lock makes this process a bit sensitive. Calling the stub in the internal Log() method is unavoidable, but this was a conscious effort to keep these classes as independent as possible. I can quickly add these methods of the launch/terminate thread methods, however, and you can choose which you prefer.

krishung5 (approved):
Personally I like putting `logger->SetBackendLoggingActive(false);` inside `TerminateLogRequestThread` more. I'll defer this decision to Iman, otherwise LGTM.

src/pb_stub.cc
Tabrizian: I'm just saying would it be possible to make `Logger` completely part of the `Stub` process to make sure that it doesn't outlive the Stub object which may lead to these issues.

fpetrini15: Yes, that should be possible. Though it may have to be "marked for improvement", as I think this PR should make into the release if the logging API is going to be released during this cycle.

Tabrizian (approved):

PR #187, opened 2022-09-23
Fix the handling for empty GPU tensor

krishung5:
Closes https://github.com/triton-inference-server/server/issues/4865

### DIFF
--- a/src/pb_memory.cc
+++ b/src/pb_memory.cc
@@ -297,9 +297,13 @@ PbMemory::GetGPUStartAddress()
     CUDAHandler& cuda_api = CUDAHandler::getInstance();
     CUdeviceptr start_address;
 
-    cuda_api.PointerGetAttribute(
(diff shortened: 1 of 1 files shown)


src/pb_memory.cc
```diff
@@ -297,9 +297,13 @@ PbMemory::GetGPUStartAddress()
     CUDAHandler& cuda_api = CUDAHandler::getInstance();
     CUdeviceptr start_address;
```
tanmayv25, line 298: Shouldn't we initialize `start_address` to some value before returning it if data_ptr_ is null?

tanmayv25 (changes requested):

src/pb_memory.cc
krishung5: Fixed.

kthui:
Looks good to me.

tanmayv25 (approved):

PR #188, opened 2022-09-24
Fix Python Backend shm-default-byte-size out_of_range error

wangchengdng:
When using the python backend, if the shm-default-byte-size or shm-growth-byte-size is set larger than 2 * 1024 * 1024 * 1024 bytes, the tritonserver will fail to start
Start command
```
./bin/tritonserver --grpc-port=8001 --model-repository=phash_zoo/ --repository-poll-secs=60 --strict-model-config=false --log-verbose=true --backend-config=python,shm-default-byte-size=2147483648
```
Error info
```
I0924 12:45:02.620398 2315 python.cc:630] Starting Python backend stub:  exec /opt/tritonserver/backends/python/triton_python_backend_stub phash_zoo/phash_compute/1/model.py triton_python_backend_shm_region_1 2147483648 67108864 2153 /opt/tritonserver/backends/python 336 phash_compute_0
terminate called after throwing an instance of 'std::out_of_range'
  what():  stoi
```
I found that shm-size is stored in int64_t, but std::stoi returns int, which will cause a out_of_range error, so I changed std::stoi to std::stol, which is consistent with the use in another file:https://github.com/triton-inference-server/python_backend/blob/4fc9cf6efd1df2b6341b8b92c2b9a9068400f4a6/src/python_be.cc#L1660

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -1197,7 +1197,7 @@ main(int argc, char** argv)
   // Path to model.py
   std::string model_path = argv[1];
   std::string shm_region_name = argv[2];
-  int64_t shm_default_size = std::stoi(argv[3]);
+  int64_t shm_default_size = std::stol(argv[3]);
 
   std::vector<std::string> model_path_tokens;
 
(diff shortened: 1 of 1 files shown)


Tabrizian:
@wangchengdng Thanks for your pull request. Could you please sign the CLA according to the instructions here:

https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla

Tabrizian:
Closing due to lack of activity.

PR #191, opened 2022-10-08
Update EXECUTION_ENV_PATH documentation

kthui:


README.md
```diff
@@ -718,7 +718,16 @@ default version of the stub is Python 3.8.
...
 `config.pbtxt` of all the models that want to use the execution environment.
 
-4. If you need to compile the Python backend stub, it is recommended that you
+4. The `EXECUTION_ENV_PATH` **should not** point to anywhere outside the
+`$$TRITON_MODEL_DIRECTORY`, the behavior of accessing anywhere outside the
+`$$TRITON_MODEL_DIRECTORY` is **undefined**.
+
+5. An *experimental* feature has been added to allow the `EXECUTION_ENV_PATH`
+to access one level outside the `$$TRITON_MODEL_DIRECTORY` for local and S3
+storages. The *experimental* feature can be modified or removed without any
```
Tabrizian, line 727: "The experimental.." -> "This experimental..."

Tabrizian: Add this sentence in the end. "Note that this experimental feature does NOT work for Azure Cloud Storage (ACS) or Google Cloud Storage (GCS)."

GuanLuo: Just not just don't mention this at all?

kthui: Yes. I think not mentioning this at all can simplify things, because undefined behavior can cover it all.
This sentence is removed.
@Tabrizian please comment if you think otherwise.

README.md
```diff
@@ -718,7 +718,16 @@ default version of the stub is Python 3.8.
 provide the path to the tar file in the `EXECUTION_ENV_PATH` in the
 `config.pbtxt` of all the models that want to use the execution environment.
 
-4. If you need to compile the Python backend stub, it is recommended that you
+4. The `EXECUTION_ENV_PATH` **should not** point to anywhere outside the
```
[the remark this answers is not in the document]
kthui: It shouldn't matter whether `$$TRITON_MODEL_DIRECTORY` is used or not, because
For this file directory:
```
model_repo
|
`---model_dir
|
`---env_file.tar.gz
```
The user can reference `$$TRITON_MODEL_DIRECTORY/../env_file.tar.gz` or `s3://model_repo/env_file.tar.gz`, which are the same, and both will not work without the patch. Then, the added if statement doesn't cover the latter case, which is inaccurate.

Tabrizian (approved):

README.md
Tabrizian: Sounds good to me too.

README.md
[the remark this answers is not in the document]
GuanLuo: This is for eliminating ambiguity, for users that don't know implementation detail, they may think that `$$TRITON_MODEL_DIRECTORY` is just `s3://model_repo/model_dir`, which is exactly the confusion we have seen in the GitHub issue. Under this notion, does that mean `EXECUTION_ENV_PATH: "s3://model_repo/model_dir/../env_file.tar.gz"` is not accepted? This is okay and thus I suggested to add the clause above.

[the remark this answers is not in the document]
kthui: Do you mean we divide the responsibility in the following way, for the future:
1. The core only provides access guarantee to the contents **inside** `$$TRITON_MODEL_DIRECTORY`, so paths using `$$TRITON_MODEL_DIRECTORY` should not escape outside the model directory.
2. Any paths that do not include `$$TRITON_MODEL_DIRECTORY` is the responsibility of the backend to fetch the content, whether the content is stored locally or from the cloud.

Then, it means `$$TRITON_MODEL_DIRECTORY/../env_file.tar.gz` path is invalid, as it requires the core to access outside the model directory. `s3://model_repo/env_file.tar.gz` and `s3://model_repo/model_dir/../env_file.tar.gz` are valid from the core's perspective, but it is up to the backend to ultimately decide whether it is valid or not, and the backend has to fetch the file.

All three paths are covered by this patch, so I think we can add the "If $$TRITON_MODEL_DIRECTORY is used in EXECUTION_ENV_PATH, ..." statement, for the future.

kthui:
Added more documentation on the behavior of `$$TRITON_MODEL_DIRECTORY`.

README.md
```diff
@@ -718,7 +718,16 @@ default version of the stub is Python 3.8.
...
 
-4. If you need to compile the Python backend stub, it is recommended that you
+4. If `$$TRITON_MODEL_DIRECTORY` is used in the `EXECUTION_ENV_PATH`, the final
+`EXECUTION_ENV_PATH` **must not** escape from the `$$TRITON_MODEL_DIRECTORY`,
+as the behavior of accessing anywhere outside the `$$TRITON_MODEL_DIRECTORY` is
+**undefined**.
+
+5. If a non-`$$TRITON_MODEL_DIRECTORY` `EXECUTION_ENV_PATH` is used, only local
+file system paths are currently supported. The behavior of using cloud paths is
+**undefined**.
```
GuanLuo, line 728: This will be fixed by long term solution, right?

GuanLuo (approved):

README.md
kthui: Yes, that is the ultimate goal.
Any S3 paths will work after this patch, providing the core can access it.

PR #195, opened 2022-11-04
Logger release info and commands updated

hamidriasat:
README.md updated with following changes

- Logger release info added for better understanding
- Unnecessary `$` sign removed from commands to support better copy/past using clipboard button


Tabrizian (approved):

Tabrizian:
@hamidriasat Thanks for your contribution!

PR #197, opened 2022-11-18
Link tcmalloc to stub process

krishung5:


CMakeLists.txt
```diff
@@ -248,6 +254,7 @@ target_link_libraries(
    pybind11::embed
    -lrt                           # shared memory 
    -larchive                      # libarchive
+   tcmalloc
```
Tabrizian, line 257: Can you elaborate why this is required for the stub process? I was expecting that it might be required for the backend shared library but not the stub process.

krishung5: Sorry for the late reply. I remembered I saw linking tcmalloc to executable was advised somewhere in the documentation. But since the stub process is run separately and only interacts with the parent process, maybe it would make more sense to just link tcmalloc to the python backend shared library.

PR #198, opened 2022-11-24
Acquire GIL lock when releasing DLPack tensors

Tabrizian:
There has been some changes in PyTorch and it no longer has GIL lock acquired when releasing the tensors (See this [PR ](https://github.com/pytorch/pytorch/pull/83623/)and this [issue ](https://github.com/pytorch/pytorch/issues/88082)for more details).

We need to acquire this lock when in the deleter.

Closes: https://github.com/triton-inference-server/server/issues/5073
Closes: https://github.com/triton-inference-server/server/issues/5055

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,7 @@ FetchContent_MakeAvailable(repo-common repo-core repo-backend)
 FetchContent_Declare(
   pybind11
   GIT_REPOSITORY "https://github.com/pybind/pybind11"
-  GIT_TAG "v2.6"
+  GIT_TAG "v2.10"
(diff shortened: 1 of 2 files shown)


GuanLuo (approved):

GuanLuo (commented):
Do we not need to acquire lock elsewhere? i.e. when creating dlpack tensor?

Tabrizian:
The GIL lock needs to be only acquired when interacting with the Python interpreter. Since in the destructor we are calling `dec_ref/inc_ref`, we need to acquire the lock since they are calling Python C-API methods. Also, we don't release the lock unless we are executing the BLS requests.

PR #199, opened 2022-12-13
Fix GPU buffers when dynamic batching is enabled

Tabrizian:
As a result of using unordered_map the order of GPU buffers returned to the stub process could change resulting in the bug.

Testing: https://github.com/triton-inference-server/server/pull/5166

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -25,6 +25,7 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "pb_stub.h"
+
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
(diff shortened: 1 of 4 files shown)


jbkyang-nvi (approved):

krishung5 (approved):

src/python_be.cc
```diff
@@ -1389,6 +1390,7 @@ ModelInstanceState::ProcessRequests(
         gpu_buffers_handle_shm[index] = pb_memory->ShmHandle();
         ++index;
       }
+      response_index++;
```
GuanLuo, line 1393: What is the use of `response_index`?

src/pb_stub.cc
```diff
@@ -748,22 +749,14 @@ Stub::ProcessRequests(RequestBatch* request_batch_shm_ptr)
...
       InferResponse* infer_response = responses[i].cast<InferResponse*>();
       InferRequest* infer_request = py_request_list[i].cast<InferRequest*>();
       infer_response->PruneOutputTensors(infer_request->RequestedOutputNames());
 
       ProcessResponse(infer_response);
-      for (auto output_tensor : infer_response->OutputTensors()) {
-        if (!output_tensor->IsCPU()) {
-          gpu_tensors.push_back(output_tensor);
-        }
-      }
```
GuanLuo, line 766: This is cleanup outdated code?

Tabrizian: Yes

src/python_be.cc
Tabrizian: It is used in Line 1382 for error reporting.

PR #200, opened 2022-12-21
Make sure to use correct shared memory mapping before deallocation

Tabrizian:
The problem was that on deallocate, we didn't update the shared map. As a result if shared memory growth happened and the object was using an old shared memory map, the shared memory data became corrupted.

### DIFF
--- a/src/shm_manager.h
+++ b/src/shm_manager.h
@@ -179,6 +179,7 @@ class SharedMemoryManager {
                                        shm_ownership_data](T* memory) {
       bool destroy = false;
       bi::scoped_lock<bi::interprocess_mutex> gaurd{*shm_mutex_};
+      GrowIfNeeded(0);
       shm_ownership_data->ref_count_ -= 1;
       if (shm_ownership_data->ref_count_ == 0) {
         destroy = true;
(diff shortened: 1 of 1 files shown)


src/shm_manager.h
```diff
@@ -179,6 +179,7 @@ class SharedMemoryManager {
                                        shm_ownership_data](T* memory) {
       bool destroy = false;
       bi::scoped_lock<bi::interprocess_mutex> gaurd{*shm_mutex_};
+      GrowIfNeeded(0);
```
GuanLuo, line 182: Can you add comment on this? By the name it seems like you are relying on the side-affect of the function?

Tabrizian: Added a comment. Yes, basically, both sides (stub and main process) need to make sure to use the updated shared memory map for allocation/deallocation.

GuanLuo: Actually, why not using the safe version of `Deallocate`? To avoid mutex dead lock, you can add scope:
```
      ...
      {
        bi::scoped_lock<bi::interprocess_mutex> gaurd{*shm_mutex_};
        shm_ownership_data->ref_count_ -= 1;
        if (shm_ownership_data->ref_count_ == 0) {
          destroy = true;
        }
      }
      if (destroy) {
        Deallocate(handle);
      }
      ...
```

Tabrizian: It is to avoid locking the mutex two times.

GuanLuo (approved):

krishung5 (approved):

tanmayv25 (approved):

PR #202, opened 2023-01-10
Remove LOG_MESSAGE after fork in child process

kthui:
The LOG_MESSAGE has underlying mutexes that could be forked when it is locked by other thread(s). If that is the case, calling LOG_MESSAGE will hang the process.


kthui:
#204 is a better fix.

PR #203, opened 2023-01-11
Add decoupled support for BLS

krishung5:
In this PR, ~~two APIs, `InferenceRequest.stream_exec()` and `InferenceRequest.async_stream_exec()` are added for BLS decoupled support.~~ two arguments, `decoupled` and `execution_timeout` are added to the original `exec()` and `async_exec()` funstion for BLS decoupled support. [Here ](https://docs.google.com/document/d/1_bb574D4_XPR-_CzztJjHL9LUYcNSI5bdbuFF22I1oU/edit)is the design doc for reference.

Under the hood, chained futures were used for retrieving responses from decoupled models (please refer to the design [here](https://docs.google.com/presentation/d/1aIDFLy0KPe1r-Vv88H4bRt1t9BMdSsjOn5At0hWFUhQ/edit#slide=id.p)). ~~hence, instead of using the [generator](https://docs.python.org/3/glossary.html#term-generator), the futures will gather the responses and the responses will be returned as a list~~. A generator that contains all the responses will be returned.

**Update: Currently, all the responses will be retrieved first and then transfer it to the user's Python model. For the next release, we would fix this issue and send the responses as they are being received.**

For the timeout parameter mentioned in the design doc, the timeout in between two different responses from the generator will not be needed since the implementation of using chained futures handles the possible missing `TRITONSERVER_RESPONSE_COMPLETE_FINAL` flag leading to infinite loop situation, 

Testing: https://github.com/triton-inference-server/server/pull/5245

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,6 +194,8 @@ set(
   src/response_sender.h
   src/pb_stub.h
   src/pb_stub.cc
+  src/pb_generator.h
+  src/pb_generator.cc
 )
 
 list(APPEND
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,6 +174,8 @@ set(
   src/request_executor.h
   src/stub_launcher.h
   src/stub_launcher.cc
+  src/infer_payload.h
+  src/infer_payload.cc
 )
 
 list(APPEND
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,7 +43,7 @@ set(TRITON_COMMON_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/
 set(TRITON_CORE_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/core repo")
 
 if(NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE Debug)
+  set(CMAKE_BUILD_TYPE Release)
 endif()
 
 #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,10 @@ set(TRITON_BACKEND_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server
 set(TRITON_COMMON_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/common repo")
 set(TRITON_CORE_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/core repo")
 
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Debug)
+endif()
+
 #
 # Dependencies
 #
--- a/README.md
+++ b/README.md
@@ -858,13 +858,13 @@ class TritonPythonModel:
           requested_output_names=['REQUESTED_OUTPUT_1', 'REQUESTED_OUTPUT_2'],
           inputs=[<pb_utils.Tensor object>])
 
-      # `pb_utils.InferenceRequest` supports request_id, correlation_id, and
-      # model version in addition to the arguments described above. These
-      # arguments are optional. An example containing all the arguments:
+      # `pb_utils.InferenceRequest` supports request_id, correlation_id,
+      # model version and timeout in addition to the arguments described above.
+      # These arguments are optional. An example containing all the arguments:
       # inference_request = pb_utils.InferenceRequest(model_name='model_name',
       #   requested_output_names=['REQUESTED_OUTPUT_1', 'REQUESTED_OUTPUT_2'],
       #   inputs=[<list of pb_utils.Tensor objects>],
-      #   request_id="1", correlation_id=4, model_version=1, flags=0)
+      #   request_id="1", correlation_id=4, model_version=1, flags=0, timeout=5)
 
       # Execute the inference_request and wait for the response
       inference_response = inference_request.exec()
@@ -951,10 +951,13 @@ models in both [default mode](#default-mode) and
 [generator](https://docs.python.org/3/glossary.html#term-generator) of
 inference responses returned by a decouple model. If the `decoupled` parameter
 is set to `False`, the `exec` and `async_exec` function will return a single
-response as shown in the example above. Besides, you can set the timeout via
-the parameter 'timeout' in microseconds. If the request times out, the request
-will respond with an error. The default of 'timeout' is 0 which indicates
-that the request has no timeout. Example below shows how to use this feature:
+response as shown in the example above.
+
+Besides, you can set the timeout via the parameter 'timeout' in microseconds
+within the constructor of `InferenceRequest`. If the request times out, the
+request will respond with an error. The default of 'timeout' is 0 which
+indicates that the request has no timeout. Example below shows how to use this
+feature:
 
 ```python
 import triton_python_backend_utils as pb_utils
@@ -974,18 +977,18 @@ class TritonPythonModel:
           requested_output_names=['REQUESTED_OUTPUT_1', 'REQUESTED_OUTPUT_2'],
           inputs=[<pb_utils.Tensor object>])
 
-      # `pb_utils.InferenceRequest` supports request_id, correlation_id, and
-      # model version in addition to the arguments described above. These
-      # arguments are optional. An example containing all the arguments:
+      # `pb_utils.InferenceRequest` supports request_id, correlation_id,
+      # model version and timeout in addition to the arguments described above.
+      # These arguments are optional. An example containing all the arguments:
       # inference_request = pb_utils.InferenceRequest(model_name='model_name',
       #   requested_output_names=['REQUESTED_OUTPUT_1', 'REQUESTED_OUTPUT_2'],
       #   inputs=[<list of pb_utils.Tensor objects>],
-      #   request_id="1", correlation_id=4, model_version=1, flags=0)
+      #   request_id="1", correlation_id=4, model_version=1, flags=0, timeout=5)
 
       # Execute the inference_request and wait for the response. Here we are
       # running a BLS request on a decoupled model, hence setting the parameter
       # 'decoupled' to 'True'.
-      inference_responses = inference_request.exec(decoupled=True, timeout=10)
+      inference_responses = inference_request.exec(decoupled=True)
 
       for inference_response in inference_responses:
         # Check if the inference response has an error
@@ -1041,7 +1044,7 @@ class TritonPythonModel:
         # [Awaitable](https://docs.python.org/3/library/asyncio-task.html#awaitables)
         # object.
         infer_response_awaits.append(
-          inference_request.async_exec(decoupled=True, timeout=10))
+          inference_request.async_exec(decoupled=True))
 
       # Wait for all of the inference requests to complete.
       async_responses = await asyncio.gather(*infer_response_awaits)
--- a/README.md
+++ b/README.md
@@ -200,28 +200,31 @@ class TritonPythonModel:
 
     @staticmethod
     def auto_complete_config(auto_complete_model_config):
-        """`auto_complete_config` is called only once when loading the model assuming
-        the server was not started with `--disable-auto-complete-config`. Implementing
-        this function is optional. No implementation of `auto_complete_config` will
-        do nothing. This function can be used to set `max_batch_size`, `input` and
-        `output` properties of the model using `set_max_batch_size`, `add_input`, and
-        `add_output`. These properties will allow Triton to load the model with minimal
-        model configuration in absence of a configuration file. This function returns
-        the `pb_utils.ModelConfig` object with these properties. You can use the `as_dict`
-        function to gain read-only access to the `pb_utils.ModelConfig` object.
-        The `pb_utils.ModelConfig` object being returned from here will be used as
-        the final configuration for the model.
+        """`auto_complete_config` is called only once when loading the model
+        assuming the server was not started with
+        `--disable-auto-complete-config`. Implementing this function is
+        optional. No implementation of `auto_complete_config` will do nothing.
+        This function can be used to set `max_batch_size`, `input` and `output`
+        properties of the model using `set_max_batch_size`, `add_input`, and
+        `add_output`. These properties will allow Triton to load the model with
+        minimal model configuration in absence of a configuration file. This
+        function returns the `pb_utils.ModelConfig` object with these
+        properties. You can use the `as_dict` function to gain read-only access
+        to the `pb_utils.ModelConfig` object. The `pb_utils.ModelConfig` object
+        being returned from here will be used as the final configuration for
+        the model.
 
-        Note: The Python interpreter used to invoke this function will be destroyed
-        upon returning from this function and as a result none of the objects created
-        here will be available in the `initialize`, `execute`, or `finalize` functions.
+        Note: The Python interpreter used to invoke this function will be
+        destroyed upon returning from this function and as a result none of the
+        objects created here will be available in the `initialize`, `execute`,
+        or `finalize` functions.
 
         Parameters
         ----------
         auto_complete_model_config : pb_utils.ModelConfig
-          An object containing the existing model configuration. You can build upon
-          the configuration given by this object when setting the properties for
-          this model.
+          An object containing the existing model configuration. You can build
+          upon the configuration given by this object when setting the
+          properties for this model.
 
         Returns
         -------
@@ -294,7 +297,8 @@ class TritonPythonModel:
           Both keys and values are strings. The dictionary keys and values are:
           * model_config: A JSON string containing the model configuration
           * model_instance_kind: A string containing model instance kind
-          * model_instance_device_id: A string containing model instance device ID
+          * model_instance_device_id: A string containing model instance device
+            ID
           * model_repository: Model repository path
           * model_version: Model version
           * model_name: Model name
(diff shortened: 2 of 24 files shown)


src/pb_stub.cc
```diff
@@ -1101,13 +1106,76 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
-              return response;
+              auto responses = infer_request->Exec(false /* is_stream*/);
+              return responses[0];
             };
             py::object future =
                 loop.attr("run_in_executor")(py::none(), callback);
             return future;
           })
+      .def(
+          "stream_exec",
```
krishung5, line 1117: On a second thought, for the execution timeout, it is possible that the backend side cannot properly clean up the objects if the process is terminated when exceeding the execution timeout. I was wondering if the execution timeout should be added to the API or if there is some WAR for this.

src/pb_stub.cc
```diff
@@ -1101,13 +1106,76 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
+          [](std::shared_ptr<InferRequest>& infer_request,
+             const int32_t execution_timeout) {
```
krishung5, line 1119: On a second thought, for the execution timeout, it is possible that the backend side cannot properly clean up the objects if the process is terminated when exceeding the execution timeout. I was wondering if the execution timeout should be added to the API or if there is some WAR for this.

CMakeLists.txt
```diff
@@ -42,6 +42,10 @@ set(TRITON_BACKEND_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server
 set(TRITON_COMMON_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/common repo")
 set(TRITON_CORE_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/core repo")
 
+if(NOT CMAKE_BUILD_TYPE)
```
Tabrizian, line 45: Why is this required?

src/python_be.cc
```diff
@@ -713,30 +732,58 @@ ModelInstanceState::ExecuteBLSRequest(std::shared_ptr<IPCMessage> ipc_message)
       }
 
       if (pb_exception.what() != nullptr) {
-        infer_response =
-            request_executor->Infer(infer_request, &inference_response);
+        auto response_future =
+            request_executor->Infer(infer_request, is_stream);
+        GetBLSResponses(infer_responses, std::move(response_future));
```
Tabrizian, line 737: We don't want to wait for all the responses to be received and them send them to the stub process. We want to send the responses as we are receiving them from the model.

Tabrizian (changes requested):

CMakeLists.txt
krishung5: Adding this just because I'm seeing we set CMAKE_BUILD_TYPE for other backends but not for Python backend. I can remove it if you think this is not needed.

src/infer_request.h
```diff
@@ -69,8 +73,14 @@ class InferRequest {
   const std::set<std::string>& RequestedOutputNames();
   bi::managed_external_buffer::handle_t ShmHandle();
 
+  void SetPrevPromise(std::promise<std::unique_ptr<InferResponse>>** promise);
+  void SetValueForPrevPromise(std::unique_ptr<InferResponse> infer_response);
```
Tabrizian, line 77: use const& for `infe_response`

src/request_executor.cc
```diff
@@ -56,12 +56,110 @@ void
...
+
+        response = nullptr;
+      }
+      std::shared_ptr<PbError> pb_error =
+          std::make_shared<PbError>(pb_exception.what());
+      infer_response = std::make_unique<InferResponse>(
+          std::vector<std::shared_ptr<PbTensor>>{}, pb_error);
+    }
+
+    infer_response->SetCompletedResponse(response);
```
Tabrizian, line 132: Why do you need to set response for `infer_response`?

src/request_executor.cc
```diff
@@ -56,12 +56,110 @@ void
...
+  } else if (
+      (*p)->IsDecoupled() &&
+      (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) != 0) {
+    // An empty response may be the last reponse for decoupled models.
+    (*p)->SetValueForPrevPromise(std::unique_ptr<InferResponse>{});
+    (*p)->ResetPrevPromise();
+  } else {
+    (*p)->SetValueForPrevPromise(std::unique_ptr<InferResponse>{});
+    (*p)->ResetPrevPromise();
+    throw PythonBackendException("Unexpected empty response.");
```
Tabrizian, line 162: I don't think you can throw an exception here.

src/pb_stub.cc
Tabrizian: I think for timeout you can use the server API to set the timeout: https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonserver.h#L1088 If the request times out, the request will respond with an error.

src/pb_stub.cc
```diff
@@ -1101,13 +1106,76 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
+            std::mutex timeout_mu;
+            std::condition_variable timeout_cv;
+            std::vector<std::shared_ptr<InferResponse>> responses;
+            std::thread t([&timeout_cv, &responses, &infer_request]() {
```
Tabrizian, line 1123: Spawning a thread for each request is a lot of overhead. I think it would be best to just use the server timeout.

src/infer_response.cc
```diff
@@ -179,6 +179,47 @@ InferResponse::Error()
...
+    std::promise<std::unique_ptr<InferResponse>>* promise)
+{
+  next_response_future_ =
+      std::make_unique<std::future<std::unique_ptr<InferResponse>>>(
+          promise->get_future());
+  ;
+}
+
+void
+InferResponse::ResetNextResponseFuture()
```
Tabrizian, line 193: Instead of manually setting these futures would it be possible to pass a `flags` argument in the constructor which automatically sets the proper futures? For example, if the flags says it is the final request the future would be null, otherwise it would be an empty promise?

src/infer_request.cc
```diff
@@ -133,6 +133,38 @@ InferRequest::ShmHandle()
   return shm_handle_;
 }
 
+void
+InferRequest::SetPrevPromise(
```
Tabrizian, line 137: Why do we need promises for the inference request?

src/python_be.cc
krishung5: Synced offline that we will be addressing this comment in the next release. Created a [ticket](https://jirasw.nvidia.com/browse/DLIS-4529) for tracking.

README.md
```diff
@@ -917,6 +917,110 @@ class TritonPythonModel:
 A complete example for sync and async BLS in Python backend is included in the
 [Examples](#examples) section.
 
+Starting from 23.02 release, you can execute inference requests on decoupled
+models in both [default mode](#default-mode) and
+[decoupled mode](#decoupled-mode). By using function `stream_exec`, you can get
```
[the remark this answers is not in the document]
krishung5: Great idea, I think this will provide a neater interface to users. I can make the change if Tanmay has no objection to it.

README.md
```diff
@@ -917,6 +917,110 @@ class TritonPythonModel:
...
+      # Execute the inference_request and wait for the response
+      inference_responses = inference_request.stream_exec()
+
+      for inference_response in inference_responses:
+        # Check if the inference response has an error
+        if inference_response.has_error():
+            raise pb_utils.TritonModelException(inference_response.error().message())
+        else:
+            # Extract the output tensors from the inference response.
+            output1 = pb_utils.get_output_tensor_by_name(inference_response, 'REQUESTED_OUTPUT_1')
```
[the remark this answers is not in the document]
krishung5: Will fix the character count. Would we want to do the same for the code section? I think the code section is easier to read as it is. Let me know if you think otherwise,

src/pb_stub.cc
krishung5: Great suggestion! Updated.

README.md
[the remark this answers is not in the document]
Tabrizian: Yes, it would be great to keep the character count below 80 in the code too.

src/infer_request.h
krishung5: Here we are calling `set_value` to set the promise with the `infer_response` in this function. The ownership of `infer_response` is transferring, I might be wrong here but I think due to that we cannot use `const&` for `infer_response`.

README.md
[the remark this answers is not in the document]
krishung5: Updated.

src/request_executor.cc
krishung5: Great catch. Updated to make the error being returned as a part of  `InferResponse` object rather than throwing exception here.

src/infer_response.cc
krishung5: Yes. Now we are setting the future in the constructor.

src/infer_request.cc
krishung5: We would need to store the previous promise so that we can call `set_value` later with the current response.

README.md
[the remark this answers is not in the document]
krishung5: Updated

krishung5:
@Tabrizian The above comments are addressed, please review. For the shm memory leak that I mentioned before, I'm still looking into it and will update here once it's fixed. Thanks!

README.md
```diff
@@ -904,27 +928,154 @@ class TritonPythonModel:
...
+            # Decide the next steps for model execution based on the received
+            # output tensors.
 ```
 
 A complete example for sync and async BLS in Python backend is included in the
 [Examples](#examples) section.
 
+Starting from 23.02 release, you can execute inference requests on decoupled
+models in both [default mode](#default-mode) and
+[decoupled mode](#decoupled-mode). By setting the parameter `decoupled` to
```
Tabrizian, line 949: "By setting the parameter `decoupled` to `True`" -> "By setting the `decoupled parameter to `True`,

src/infer_response.cc
Tabrizian: Can we delete this function now?

src/infer_response.cc
```diff
@@ -179,6 +198,34 @@ InferResponse::Error()
   return error_;
 }
 
+void
+InferResponse::SetNextResponseFuture(
```
Tabrizian, line 202: Can we delete this function too? It is not used anywhere.

src/request_executor.cc
```diff
@@ -245,98 +346,29 @@ RequestExecutor::Infer(
     }
 
     {
-      auto p = new std::promise<TRITONSERVER_InferenceResponse*>();
-      std::future<TRITONSERVER_InferenceResponse*> completed = p->get_future();
+      auto p = new std::promise<std::unique_ptr<InferResponse>>();
+      response_future = p->get_future();
+      infer_request->SetPrevPromise(std::move(&p));
```
Tabrizian, line 351: Why do we need to attach the promise to the inference request? How about just passing the response_future as the `userp` in the callback?

src/pb_stub.cc
```diff
@@ -1149,6 +1178,18 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
+      module, "ResponseGenerator")
+      .def(py::init<const std::vector<std::shared_ptr<InferResponse>>&>())
+      .def(
+          "__iter__",
+          [](ResponseGenerator& self) {
+            return py::make_iterator(self.Begin(), self.End());
+          },
+          py::keep_alive<0, 1>())
+      .def("__next__", &ResponseGenerator::Next)
+      .def("__len__", &ResponseGenerator::Length);
```
Tabrizian, line 1191: Let's remove `__len__` since it cannot be supported when we change the response delivery in future releases.

krishung5: Removed.

src/infer_response.cc
krishung5: This got slipped from my previous changes. Deleted the unused functions and variable.

src/infer_response.h
```diff
@@ -110,5 +123,29 @@ class InferResponse {
...
+  int Length();
+  std::vector<std::shared_ptr<InferResponse>>::iterator Begin();
+  std::vector<std::shared_ptr<InferResponse>>::iterator End();
+
+ private:
+  std::vector<std::shared_ptr<InferResponse>> responses_;
+  size_t index_;
 };
+#endif
+
```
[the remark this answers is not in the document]
krishung5: Created new files `pb_generator.h` and `pb_generator.cc`.

README.md
```diff
@@ -954,13 +1106,16 @@ flags = pb_utils.TRITONSERVER_REQUEST_FLAG_SEQUENCE_START | pb_utils.TRITONSERVE
...
-on itself and there are no more model instances ready to execute the inference request, the
-model will block on the inference execution forever.
+- You need to make sure that the inference requests performed as a part of your
+model do not create a circular dependency. For example, if model A performs an
+inference request on itself and there are no more model instances ready to
+execute the inference request, the model will block on the inference execution
+forever.
 
-- Currently, BLS can not run inference on a decoupled model.
+- BLS can not run inference on a decoupled model using functions
```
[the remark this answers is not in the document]
krishung5: Just to clarify that the limitations which should be removed do not include the one `BLS can not run inference on a decoupled model in *async* decoupled mode`, is this correct?

README.md
```diff
@@ -904,27 +928,154 @@ class TritonPythonModel:
...
 [Examples](#examples) section.
 
+Starting from 23.02 release, you can execute inference requests on decoupled
+models in both [default mode](#default-mode) and
+[decoupled mode](#decoupled-mode). By setting the parameter `decoupled` to
+`True` in function `exec`, you can get a
+[generator](https://docs.python.org/3/glossary.html#term-generator) of
+responses returned by a decouple model. (By default the parameter `decoupled`
+is set to `False`, which makes the function return a single response as the
+example above.) Besides, you can set the exectuion timeout via the parameter
```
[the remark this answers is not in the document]
krishung5: Sure, renamed.

README.md
```diff
@@ -904,27 +928,154 @@ class TritonPythonModel:
...
 
 A complete example for sync and async BLS in Python backend is included in the
 [Examples](#examples) section.
 
+Starting from 23.02 release, you can execute inference requests on decoupled
+models in both [default mode](#default-mode) and
+[decoupled mode](#decoupled-mode). By setting the parameter `decoupled` to
+`True` in function `exec`, you can get a
+[generator](https://docs.python.org/3/glossary.html#term-generator) of
+responses returned by a decouple model. (By default the parameter `decoupled`
```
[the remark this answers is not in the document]
krishung5: Updated the wording here.

README.md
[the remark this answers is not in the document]
krishung5: Updated.

src/infer_response.cc
krishung5: Yes, deleted.

src/request_executor.cc
krishung5: Sorry I'm not sure what you mean here. The `response_future` here lets us to retrieve the response later when we need to.  Would it be possible if we return the `response_future` and pass it to the callback function as well?
We need to pass the promise to the callback function so that we can set the current `InferResponse` to the promise. We also need to check if the request is for decoupled model or not, and the information can be retrieved by `infer_request->IsDecoupled()`. The promise is stored inside the `inferRequest` object so that the lifetime of the promise would be long enough for  calling `set_value` later with the next response.

Tabrizian: Discussed offline.

README.md
```diff
@@ -904,27 +928,155 @@ class TritonPythonModel:
...
+      ...
+      # Create an InferenceRequest object. `model_name`,
+      # `requested_output_names`, and `inputs` are the required arguments and
+      # must be provided when constructing an InferenceRequest object. Make
+      # sure to replace `inputs` argument with a list of `pb_utils.Tensor`
+      # objects.
+      inference_request = pb_utils.InferenceRequest(
+          model_name='model_name',
+          requested_output_names=['REQUESTED_OUTPUT_1', 'REQUESTED_OUTPUT_2'],
+          inputs=[<pb_utils.Tensor object>])
```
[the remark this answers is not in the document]
krishung5: `timeout` was added below as it seems like this section only demonstrates the required arguments.

README.md
```diff
@@ -904,27 +928,155 @@ class TritonPythonModel:
...
+
+      # `pb_utils.InferenceRequest` supports request_id, correlation_id,
+      # model version and timeout in addition to the arguments described above.
+      # These arguments are optional. An example containing all the arguments:
+      # inference_request = pb_utils.InferenceRequest(model_name='model_name',
+      #   requested_output_names=['REQUESTED_OUTPUT_1', 'REQUESTED_OUTPUT_2'],
+      #   inputs=[<list of pb_utils.Tensor objects>],
+      #   request_id="1", correlation_id=4, model_version=1, flags=0, timeout=5)
```
krishung5, line 986: `timeout` was added here along with other optional arguments.

src/infer_request.cc
krishung5: This should not be a problem anymore by using `infer_payload` object.

Tabrizian (approved):

README.md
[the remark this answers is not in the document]
Tabrizian: Sorry, we just need to remove the current bullet point. Perhaps we can reword the second bullet point to "Async BLS is not supported in decoupled mode in Python backend".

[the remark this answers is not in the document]
krishung5: Updated the limitation.

Tabrizian (approved):

PR #204, opened 2023-01-11
Move child process code to before fork

kthui:
The only process that must be done after fork within the child process is the `execvp()`, and all other processes can be done before `fork()`. When running `execvp()` in child, the only issue is printing the error message if `execvp()` would fail, but it is handled by printing the error directly to `stderr` and exit immediately if an error occurs.

Closes https://github.com/triton-inference-server/server/issues/5211.

### DIFF
--- a/src/stub_launcher.cc
+++ b/src/stub_launcher.cc
@@ -197,12 +197,82 @@ StubLauncher::Setup()
 TRITONSERVER_Error*
 StubLauncher::Launch()
 {
-
   std::string stub_name;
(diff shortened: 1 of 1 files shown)


Tabrizian (approved):
Thanks for getting to the bottom of this issue. Could you also link to the Github issue that this closes?

krishung5 (approved):

kthui:
Resolves https://github.com/triton-inference-server/server/issues/5211.

PR #205, opened 2023-01-13
Atomically increase/read stub process count and fix random boost truncate fail

kthui:
This PR involves two changes:

[Atomically increase and read stub process count](https://github.com/triton-inference-server/python_backend/pull/205/commits/626349b691e405be53306bb9754a8bbffa502463)
The stub process count needs to be increased and read atomically. Otherwise, during parallel model loading, the combination: 
```
thread 1:           count = 1
thread 1: increase; count = 2
thread 2: increase; count = 3
thread 1: read;     count = 3
thread 2: read;     count = 3
```
could happen resulting in shared memory name collision, which two or more stubs could be using the same region.

[Upgrade boost to 1.79.0](https://github.com/triton-inference-server/python_backend/pull/205/commits/ebca0c101d48cbbe15ce0d931318065471665d2a)
The current version of boost (1.76.0) has two bugs in the `shared_memory_object.truncate()`. The first bug is it does not correctly return the error from `posix_fallocate()` when it encounters error, as the function only returns the code but does not set `errno`, while boost checks for `errno` for the error. The second bug is if `posix_fallocate()` encounters `EINTR` error, it is safe to retry and should not throw the error back to the caller. The upgraded version 1.79.0 fixed both bugs.

### DIFF
--- a/src/shm_manager.cc
+++ b/src/shm_manager.cc
@@ -1,4 +1,4 @@
-// Copyright 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -43,10 +43,27 @@ SharedMemoryManager::SharedMemoryManager(
 
   try {
     if (create) {
-      shm_obj_ = std::make_unique<bi::shared_memory_object>(
-          bi::open_or_create, shm_region_name.c_str(), bi::read_write);
-      shm_obj_->truncate(shm_size);
+      // Remove (if any) and create the region. Truncate the shm region could
+      // occasionally fail, allow up to max_retry for stability.
+      int max_retry = 3;
+      while (true) {
+        try {
+          bi::shared_memory_object::remove(shm_region_name.c_str());
+          shm_obj_ = std::make_unique<bi::shared_memory_object>(
+              bi::create_only, shm_region_name.c_str(), bi::read_write);
+          shm_obj_->truncate(shm_size);
+          break;
+        }
+        catch (bi::interprocess_exception& ex) {
+          if (--max_retry > 0) {
+            sleep(1);
+          } else {
+            throw ex;
+          }
+        }
+      }
     } else {
+      // Open the existing region.
       shm_obj_ = std::make_unique<bi::shared_memory_object>(
           bi::open_only, shm_region_name.c_str(), bi::read_write);
     }
--- a/src/shm_manager.cc
+++ b/src/shm_manager.cc
@@ -43,25 +43,11 @@ SharedMemoryManager::SharedMemoryManager(
 
   try {
     if (create) {
-      // Remove (if any) and create the region. Truncate the shm region could
-      // occasionally fail, allow up to max_retry for stability.
-      int max_retry = 3;
-      while (true) {
-        try {
-          bi::shared_memory_object::remove(shm_region_name.c_str());
-          shm_obj_ = std::make_unique<bi::shared_memory_object>(
-              bi::create_only, shm_region_name.c_str(), bi::read_write);
-          shm_obj_->truncate(shm_size);
-          break;
-        }
-        catch (bi::interprocess_exception& ex) {
-          if (--max_retry > 0) {
-            sleep(1);
-          } else {
-            throw ex;
-          }
-        }
-      }
+      // Remove (if any) and create the region.
+      bi::shared_memory_object::remove(shm_region_name.c_str());
+      shm_obj_ = std::make_unique<bi::shared_memory_object>(
+          bi::create_only, shm_region_name.c_str(), bi::read_write);
+      shm_obj_->truncate(shm_size);
     } else {
       // Open the existing region.
       shm_obj_ = std::make_unique<bi::shared_memory_object>(
--- a/src/stub_launcher.cc
+++ b/src/stub_launcher.cc
@@ -1,4 +1,4 @@
-// Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2022-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
--- a/src/stub_launcher.cc
+++ b/src/stub_launcher.cc
@@ -62,12 +62,12 @@ StubLauncher::Initialize(ModelState* model_state)
   is_decoupled_ = model_state->IsDecoupled();
   model_repository_path_ = model_state->RepositoryPath();
 
-  // Increase the stub process count to avoid shared memory region name
-  // collision
-  model_state->StateForBackend()->number_of_instance_inits++;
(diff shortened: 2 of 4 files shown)


src/stub_launcher.cc
```diff
@@ -62,12 +62,12 @@ StubLauncher::Initialize(ModelState* model_state)
...
-  // Increase the stub process count to avoid shared memory region name
-  // collision
-  model_state->StateForBackend()->number_of_instance_inits++;
+  // Atomically increase and read the stub process count to avoid shared memory
+  // region name collision
+  int num_init = ++model_state->StateForBackend()->number_of_instance_inits;
   shm_region_name_ =
       model_state->StateForBackend()->shared_memory_region_prefix +
-      std::to_string(model_state->StateForBackend()->number_of_instance_inits);
+      std::to_string(num_init);
```
rmccorm4, line 70: If this fixes a potential race that existed, do we still need the retries above?

kthui: The retry is still needed, because the potential race still happens after fixing the atomic increment and read.

rmccorm4: > The 1000+ iteration L0_infer--cpu-only test, with parallel model loading, passed after this fix.
and
> The retry is still needed, because the potential race still happens after fixing the atomic increment and read.

Does that mean retries occurred during the 1000 iteration test?

kthui: > Does that mean retries occurred during the 1000 iteration test?
Yes.

src/shm_manager.cc
```diff
@@ -43,10 +43,27 @@ SharedMemoryManager::SharedMemoryManager(
...
-      shm_obj_->truncate(shm_size);
+      // Remove (if any) and create the region. Truncate the shm region could
+      // occasionally fail, allow up to max_retry for stability.
+      int max_retry = 3;
+      while (true) {
+        try {
+          bi::shared_memory_object::remove(shm_region_name.c_str());
+          shm_obj_ = std::make_unique<bi::shared_memory_object>(
+              bi::create_only, shm_region_name.c_str(), bi::read_write);
+          shm_obj_->truncate(shm_size);
```
GuanLuo, line 54: > When creating a new shm region, the write (truncate) call could occasionally fail due to the created shm region having "No such file or directory".

Doesn't that mean the `shared_memory_object` is not properly constructed? There is no exception being raised from the line above this?

kthui: It could be, the exception is always raised from line 54 `shm_obj_->truncate(shm_size);`.

rmccorm4: I don't see anything immediately wrong with the code, but if `shm_obj_->truncate(shm_size)` is throwing an exception that the file doesn't exist, and the construction of the `shared_memory_object` is not throwing an exception, I would imagine the shm region is beging concurrently deleted/removed unexpectedly.
```
template<class CharT>
inline bool shared_memory_object::priv_open_or_create
   (ipcdetail::create_enum_t type, const CharT *filename, mode_t mode, const permissions &perm)
{
...
      case ipcdetail::DoCreate:
         m_handle = ipcdetail::create_new_file(shmfile.c_str(), mode, perm, true);
      break;
...
   //Check for error
   if(m_handle == ipcdetail::invalid_file()){
      error_info err = system_error_code();
      this->priv_close();
      throw interprocess_exception(err);
   }
```

Maybe you can add some printing/logging around the deletion/removal of a shared memory region to catch some info during the exception cases?

Tabrizian: Do we know why `truncate` fails?

kthui: I am going to try @rmccorm4 suggestion earlier to run
```
bi::shared_memory_object::remove(shm_region_name.c_str());
shm_obj_ = std::make_unique<bi::shared_memory_object>(bi::create_only, shm_region_name.c_str(), bi::read_write);
shm_obj_->truncate(shm_size);
```
in a loop with multiple threads, and see if that would break too.

kthui: Further testing with some printout statements in boost show the shm region was created successfully, which the `::fchmod(m_handle, unix_perm);` returned 0 and the constructor did not throw any error. The error is from the `posix_fallocate()` inside boost `truncate()`:
```
inline void shared_memory_object::truncate(offset_t length)
{
   #ifdef BOOST_INTERPROCESS_POSIX_FALLOCATE
   int ret = posix_fallocate(m_handle, 0, length);

   if (ret && ret != EOPNOTSUPP){
      error_info err(system_error_code());
      throw interprocess_exception(err);
   }
   //ftruncate fallback
   #endif //BOOST_INTERPROCESS_POSIX_FALLOCATE
   if(0 != ftruncate(m_handle, length)){
      error_info err(system_error_code());
      throw interprocess_exception(err);
   }
}
```
Ref: https://www.boost.org/doc/libs/1_76_0/boost/interprocess/shared_memory_object.hpp

Notice the error message "No such file or directory" from boost regarding `posix_fallocate()` is incorrect, because `posix_fallocate()` does not set `errno` https://man7.org/linux/man-pages/man3/posix_fallocate.3.html#RETURN_VALUE, but boost checks the error from `system_error_code()` which actually checks the `errno`:
```
inline int system_error_code() // artifact of POSIX and WINDOWS error reporting
{
   #if defined (BOOST_INTERPROCESS_WINDOWS)
   return winapi::get_last_error();
   #else
   return errno; // GCC 3.1 won't accept ::errno
   #endif
}
```
Ref: https://www.boost.org/doc/libs/1_76_0/boost/interprocess/errors.hpp

The actual error is within the return value which is 4, and according to the code printout from the build environment: 
```
EBADF: 9
EFBIG: 27
EINTR: 4
EINVAL: 22
ENODEV: 19
ENOSPC: 28
EOPNOTSUPP: 95
ESPIPE: 29
```
that 4 matches EINTR, and according to https://man7.org/linux/man-pages/man3/posix_fallocate.3.html#ERRORS, the actual error is "A signal was caught during execution"

src/shm_manager.cc
```diff
@@ -43,10 +43,13 @@ SharedMemoryManager::SharedMemoryManager(
 
   try {
     if (create) {
+      // Remove (if any) and create the region.
+      bi::shared_memory_object::remove(shm_region_name.c_str());
```
rmccorm4, line 47: @Tabrizian do we have any behavior/tests that expect a duplicate create() on an existing region to fail? Not sure if removing before every create would introduce a new/unexpected issue.

Tabrizian (approved):

src/shm_manager.cc
Tabrizian: We don't expect that behavior. I think its fine to remove the region before creating. But I'm not sure why @kthui has added this line. Do you see any failures if you remove this line?

kthui: It is actually from a boost example: https://www.boost.org/doc/libs/1_49_0/doc/html/interprocess/sharedmemorybetweenprocesses.html#interprocess.sharedmemorybetweenprocesses.sharedmemory.shared_memory_a_simple_example

They remove the region when creating the `struct shm_remove` and remove again when it goes out of scope.

rmccorm4 (approved):

src/shm_manager.cc
Tabrizian: I think they are just trying to be safe but either way its ok with me.

GuanLuo (approved):

PR #207, opened 2023-02-15
Typo in class variable declaration on preprocessing example

isaacperez:
`model_config` is declared twice: once as a class variable (`self.model_config`) and once as a local variable (`model_config`) in the same statement. I think it is better to declare it only once as class variable.

I have updated the only line where it is used with the new name.


Tabrizian:
It is intentional to avoid typing `self.` every time `model_config` is used.

PR #208, opened 2023-02-23
Enhancement for BLS decoupled support

krishung5:


krishung5:
Refactoring the current log message queue only requires some name changes, so instead of creating another PR, it's included in the commit [Rename the logging-specific objects](https://github.com/triton-inference-server/python_backend/pull/208/commits/fbc4cbbebdeac123d11ebdeeaa0c23c8c27b2473).

src/python_be.cc
```diff
@@ -1468,6 +1494,80 @@ ModelInstanceState::ProcessRequests(
...
+  {
+    bi::scoped_lock<bi::interprocess_mutex> lock{
+        *(ipc_message->ResponseMutex())};
+    bool success = false;
+    while (!success) {
+      Stub()->BLSResponseQueue()->Push(ipc_message->ShmHandle(), 1000, success);
+    }
+    ipc_message->ResponseCondition()->wait(lock);
+  }
+}
```
Tabrizian, line 1569: This looks similar to `ExecuteBLSRequest` portion of the code. Do you think we can refactor this into a common function?

src/pb_stub.cc
```diff
@@ -909,29 +930,186 @@ Stub::SendLogMessage(std::unique_ptr<PbLog>& log_send_message)
...
+          infer_response = std::make_unique<InferResponse>(
+              std::vector<std::shared_ptr<PbTensor>>{},
+              std::make_shared<PbError>(
+                  "An error occurred while performing BLS request."));
+        }
+      }
+
+      if (responses_is_set) {
+        // Need to acquire the GIL to avoid hangs.
+        py::gil_scoped_acquire acquire;
```
Tabrizian, line 1058: Can we move the GIL lock to: https://github.com/triton-inference-server/python_backend/blob/0aef2c4cbee83984ce9302d4563a3c40f351d992/src/pb_tensor.cc#L470

You need to add `#ifdef TRITON_PB_STUB` around it so that it is only activated in the Python backend stub.

src/python_be.cc
```diff
@@ -735,57 +717,36 @@ ModelInstanceState::ExecuteBLSRequest(
...
-          response_batch->batch_size = 1;
-          response_batch->has_error = false;
-          response_batch->is_error_set = false;
-          response_batch->cleanup = false;
-          response_batch->response_size = response_length;
-        }
+        // Need to manage the lifetime of InferPayload and RequestExecutor
+        // objects for bls decoupled responses.
+        infer_payload_[reinterpret_cast<void*>(&infer_payload)] = infer_payload;
+        request_executor_[reinterpret_cast<void*>(&infer_payload)] =
```
Tabrizian, line 730: Why do you need to keep the request executor alive?

src/python_be.cc
```diff
@@ -1468,6 +1494,80 @@ ModelInstanceState::ProcessRequests(
...
+  }
+
+  {
+    bi::scoped_lock<bi::interprocess_mutex> lock{
+        *(ipc_message->ResponseMutex())};
+    bool success = false;
+    while (!success) {
+      Stub()->BLSResponseQueue()->Push(ipc_message->ShmHandle(), 1000, success);
+    }
+    ipc_message->ResponseCondition()->wait(lock);
```
Tabrizian, line 1567: This also needs to be wrapped around in a while loop.

src/pb_stub.cc
```diff
@@ -909,29 +930,186 @@ Stub::SendLogMessage(std::unique_ptr<PbLog>& log_send_message)
...
+{
+  if (bls_response_thread_) {
+    bls_response_thread_ = false;
+    bls_response_queue_->Push(DUMMY_MESSAGE);
+    bls_response_monitor_.join();
+  }
+}
+
+void
+Stub::BLSResponseQueueMonitor()
```
Tabrizian, line 1015: This can be a part of a future refactor but I was wondering whether we can create a class that encapsulates the listener threads in the parent and stub process side along with a message queue. Then, we can extend from this class to have specific handling for Logging/BLS resposnes.

src/python_be.cc
```diff
@@ -860,83 +821,148 @@ ModelInstanceState::DecoupledMessageQueueMonitor()
...
+  {
+    bi::scoped_lock<bi::interprocess_mutex> lock{*(message->ResponseMutex())};
+    message->ResponseCondition()->notify_all();
+    message->ResponseCondition()->wait(lock);
+  }
 }
 
 void
-ModelInstanceState::TerminateLogMonitor()
+ModelInstanceState::BLSResponseQueueMonitor()
```
Tabrizian, line 919: It looks like the `BLSResponseQueueMonitor` is not really required. Why not just call `SendBLSDecoupledResponse` directly in the callback function?

src/pb_generator.cc
```diff
@@ -24,39 +24,103 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#include <chrono>
+
```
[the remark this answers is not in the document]
krishung5: Removed.

src/pb_stub.cc
```diff
@@ -103,8 +104,11 @@ Stub::Instantiate(
         MessageQueue<bi::managed_external_buffer::handle_t>::
             LoadFromSharedMemory(shm_pool_, ipc_control_->parent_message_queue);
 
-    log_message_queue_ = MessageQueue<bi::managed_external_buffer::handle_t>::
-        LoadFromSharedMemory(shm_pool_, ipc_control_->log_message_queue);
+    utils_message_queue_ = MessageQueue<bi::managed_external_buffer::handle_t>::
```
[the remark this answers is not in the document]
krishung5: Renamed here and everywhere.

src/infer_payload.h
```diff
@@ -26,24 +26,31 @@
...
+#include <queue>
 #include "infer_response.h"
 
 namespace triton { namespace backend { namespace python {
 
 class InferPayload {
  public:
-  InferPayload(const bool is_decoupled);
+  InferPayload(
+      const bool is_decoupled, std::mutex& mu, std::condition_variable& cv,
```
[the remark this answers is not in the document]
krishung5: Good idea. Used a callback function instead.

src/python_be.cc
krishung5: Break down this section to two helper functions, `PrepareResponseBatch` and `PrepareResponseHandle`.

src/pb_stub.cc
krishung5: I tried to add the GIL inside `pb_tensor.cc`, but somehow it failed for async cases. Need to look into it deeper. I moved the GIL to [here](https://github.com/triton-inference-server/python_backend/pull/208/files#diff-32938d76ce399e799a75fbec5050c99c8546fcc08fb75ab526ea331342731932R159) right before `PbTensor::LoadFromSharedMemory` is called. This works for async cases.

src/python_be.cc
krishung5: I think we would need to make sure the `TRITONSERVER_ResponseAllocator` object inside request executor is alive when it allocates for decoupled responses.

src/python_be.cc
krishung5: Added.

src/pb_stub.cc
krishung5: Great idea! Right now we don't actually need this `BLSResponseQueueMonitor` as the comment below. I think we can refactor this part maybe in the python custom metrics feature. Does it make sense?

krishung5:
Observing some flaky failures in some of the test cases. I'm looking into it.

src/pb_stub.cc
Tabrizian: Sounds good.

src/python_be.cc
```diff
@@ -1468,6 +1441,112 @@ ModelInstanceState::ProcessRequests(
...
+      ipc_message->ResponseCondition()->notify_all();
+    }
+  });
+
+  {
+    bi::scoped_lock<bi::interprocess_mutex> lock{
+        *(ipc_message->ResponseMutex())};
+    bool success = false;
+    while (!success) {
+      Stub()->ParentToStubMessageQueue()->Push(
```
Tabrizian, line 1541: Can we use the `Push` without timeout so that we can remove the while loop?

krishung5: Removed timeout and the while loop.

src/pb_generator.cc
```diff
@@ -25,38 +25,110 @@
...
+  while (!done) {
+    try {
+      responses_.push_back(Next());
+    }
+    catch (const py::stop_iteration& exception) {
+      done = true;
+    }
+  }
+
+  return py::make_iterator(responses_.begin(), responses_.end());
```
Tabrizian, line 99: I think we should return a new response generator here instead of fetching all the responses at once. This would allow the loop to make progress while the responses are being fetched from the backend.

src/pb_generator.cc
```diff
@@ -25,38 +25,110 @@
...
 std::shared_ptr<InferResponse>
 ResponseGenerator::Next()
 {
-  if (index_ == responses_.size()) {
+  if (is_finished_) {
+    if (!is_cleared_) {
+      Clear();
+    }
     throw py::stop_iteration("Iteration is done for the responses.");
   }
```
Tabrizian, line 59: I was wondering whether we can loop through the responses if we have looped through them once. For example, does the code below work successfully:

```
infer_responses = infer_request.exec(decoupled=True)

for infer_response in infer_responses:
           pass
  
for infer_response in infer_responses:
           pass
```

From the code, it looks like we can only iterate once.

src/python_be.cc
krishung5: The request executor is now stored as a member variable.

src/pb_generator.cc
krishung5: Returned self here.

src/pb_generator.cc
krishung5: Made the iterator to be able to iterate multiple times.

src/pb_stub.cc
```diff
@@ -909,29 +932,196 @@ Stub::SendLogMessage(std::unique_ptr<PbLog>& log_send_message)
...
+  ipc_message->Args() = cleanup_request_message.handle_;
+
+  ScopedDefer _([&ipc_message, cleanup_message_ptr] {
+    {
+      bi::scoped_lock<bi::interprocess_mutex> lock{
+          *(ipc_message->ResponseMutex())};
+      cleanup_message_ptr->waiting_on_stub = false;
+      ipc_message->ResponseCondition()->notify_all();
+    }
+  });
```
Tabrizian, line 977: I was wondering why this additional cleanup step is required?

src/pb_stub.cc
```diff
@@ -1092,9 +1083,14 @@ Stub::ParentToStubMQMonitor()
     }
 
     {
-      std::lock_guard<std::mutex> lock(response_generator_map_mu_);
-      response_generator_map_[infer_response->Id()]->EnqueueResponse(
-          std::move(infer_response));
+      std::lock_guard<std::mutex> lock(response_iterator_map_mu_);
+      // No need to enqueue the response if the response iterator has been
+      // cleaned up and removed from the map.
+      if (response_iterator_map_.find(infer_response->Id()) !=
```
krishung5, line 1089: I think maybe we don't have to wait for retrieving the remaining responses in ResponseIterator destructor if we don't enqueue the responses after clean-up.

src/pb_stub.cc
krishung5: Removed the additional round of communication.

src/python_be.cc
```diff
@@ -734,58 +706,24 @@ ModelInstanceState::ExecuteBLSRequest(
...
-          response_batch->has_error = false;
-          response_batch->is_error_set = false;
-          response_batch->cleanup = false;
-          response_batch->response_size = response_length;
+        auto response_future =
+            request_executor_->Infer(infer_request, infer_payload);
+        infer_response = response_future.get();
+
+        if (is_decoupled && (infer_response->Id() != nullptr)) {
+          // Need to manage the lifetime of InferPayload and RequestExecutor
```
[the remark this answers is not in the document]
krishung5: Removed.

README.md
```diff
@@ -969,7 +969,7 @@ Starting from 23.03 release, you can execute inference requests on decoupled
 models in both [default mode](#default-mode) and
 [decoupled mode](#decoupled-mode). By setting the `decoupled` parameter to
 `True`, the `exec` and `async_exec` function will return a
```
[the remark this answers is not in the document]
krishung5: Fixed.

src/pb_stub.cc
krishung5: On second thought, I think I should revert this change as it's possible that there is a new response with the same id just got stored inside the map.

src/pb_response_iterator.cc
```diff
@@ -0,0 +1,153 @@
...
+      {
+        std::unique_lock<std::mutex> lock{mu_};
+        while (response_buffer_.empty()) {
+          py::gil_scoped_release release;
+          cv_.wait(lock);
+        }
+        response = response_buffer_.front();
+        response_buffer_.pop();
+        is_finished_ = response->IsLastResponse();
+        if (!response->IsEmptyResponse()) {
```
Tabrizian, line 89: I think we shouldn't swallow the empty responses. We should return it to the user. The user can can ignore it.

src/pb_stub.cc
```diff
@@ -848,52 +854,69 @@ Stub::GetOrCreateInstance()
...
+    if (!bls_response_cleanup_buffer_.empty()) {
+      void* id = std::move(bls_response_cleanup_buffer_.front());
+      if (id == DUMMY_MESSAGE) {
+        bls_response_cleanup_buffer_.pop();
+        break;
+      } else {
+        bls_response_cleanup_buffer_.pop();
+        SendCleanupId(id);
+        response_iterator_map_.erase(id);
+      }
```
Tabrizian, line 919: I think we should merge these queues into a single queue and have `command` to distinguish between them. This doesn't need to be fixed right now but we probably need to fix it later.

krishung5: Sounds good! Let me make this change in another PR.

Tabrizian (approved):

PR #209, opened 2023-02-28
Model Instance Kind Example for python backend

oandreeva-nv:
This PR adds example, showing how to integrate `KIND_GPU/KIND_CPU` with the Python model.
 
It adds 5 files:
model.py and config.pdtxt <- files required to set up a ResNet50 model on a triton server.
client.py - a simple http client that takes an image url (if none is provided, it uses a default url) and send an inference request to a server. It also has a `verbose` argument to request and print model's configuration on the server. This is used as a part of the example to check `instance_group: kind` parameter, all steps are described in the provided README.
resnet50_labels.txt - self-explanatory, is used by the client to convert prediction into an actual label.
README contains information for all necessary steps on how to run this example, as well as sample outputs.


examples/instance_kind/README.md
```diff
@@ -0,0 +1,241 @@
...
+        {
+            "name": "resnet50_0",
+            "kind": "KIND_CPU",
+            "count": 1,
+            "gpus": [],
+            "secondary_devices": [],
+            "profile": [],
+            "passive": false,
+            "host_policy": ""
+        }
```
Tabrizian, line 195: Perhaps only highlight this section of the config and remove the rest?

examples/instance_kind/model.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def initialize(self, args):
+        """
+        This function initializes pre-trained ResNet50 model,
+        depending on the value specified by an `instance_group` parameter
+        in `config.pbtxt`. 
+
+        Depending on what `instance_group` was specified in 
+        the config.pbtxt file (KIND_CPU, KIND_GPU, or both), the model instance
+        will be initialised on a cpu, a gpu, or both.
```
Tabrizian, line 43: I think it would be good to mention that without manually loading your model to GPU all the models will be loaded in the default device of the framework.

examples/instance_kind/README.md
```diff
@@ -0,0 +1,241 @@
...
+-->
+
+# Model Instance Kind Example
+
+Triton allows an execution of a model to be performed on both CPU, and GPU. 
+In Python Backend this can be achieved by specifying an 
+[`instance_group`](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_configuration.md#instance-groups) 
+parameter in a config file.
+
+In this section, we demonstrate how to specify a `device kind` for your python model.
```
Tabrizian, line 36: Rename `device kind` to `instance kind`.

Tabrizian (commented):
Minor comments, otherwise looks great! You also need to add a link to this example here: https://github.com/triton-inference-server/python_backend#examples

Also, please add a test for this example here: https://github.com/triton-inference-server/server/tree/main/qa/L0_backend_python/examples

examples/instance_kind/README.md
```diff
@@ -0,0 +1,190 @@
...
+To leave the container for the next step, press: `CTRL + P + Q`.
+
+## Start the Triton SDK Container and Test Inference
+
+To start the sdk container, run the following command:
+```
+docker run --gpus all --network=host --pid=host --ipc=host -v `pwd`:/instance_kind -ti nvcr.io/nvidia/tritonserver:<yy.mm>-py3-sdk /bin/bash
+```
+
+The `client.py` requires the following packages to be installed: `torch`, `torchvision`, `pillow` and `validators`.
```
Tabrizian, line 108: I was wondering where is `validators` used?

examples/instance_kind/README.md
```diff
@@ -0,0 +1,190 @@
...
+Results is class: TABBY
+PASS: ResNet50 instance kind
+```
+It is also possible to load your python model to both CPU and GPU if neccessary.
+To do this, make sure your config file has the following entry:
+```
+instance_group [{ kind: KIND_CPU }, { kind: KIND_GPU}]
+```
+
+For more information on possible model configurations, check out the Triton Server documentation [here](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_configuration.md#model-configuration)
```
Tabrizian, line 190: Keep the number of characters below 80 in each line.

examples/instance_kind/README.md
oandreeva-nv: Client uses utils, loaded from `torch.hub` to preprocess image from url. The exact `utils` is described here with all imports, including `validators` 

https://github.com/NVIDIA/DeepLearningExamples/blob/a8c42a142ceb67513892a303b48185d9c52864c0/PyTorch/Classification/ConvNets/image_classification/models/entrypoints.py#L43

examples/instance_kind/README.md
```diff
@@ -0,0 +1,196 @@
...
+## Pull the Triton Docker images
+
+We need to install Docker and NVIDIA Container Toolkit before proceeding, refer
+to the
+[installation steps](https://github.com/triton-inference-server/server/tree/main/docs#installation).
+
+To pull the latest containers, run the following commands:
+```
+$ docker pull nvcr.io/nvidia/tritonserver:<yy.mm>-py3
+$ docker pull nvcr.io/nvidia/tritonserver:<yy.mm>-py3-sdk
```
Tabrizian, line 65: Remove `$` from the beginning of the commands so that they can be copied more easily.

Tabrizian (commented):
One minor comment otherwise looks good.

README.md
```diff
@@ -1261,6 +1261,12 @@ The examples of decoupled models shows how to develop and serve
 You can find the complete example instructions in
 [examples/decoupled](examples/decoupled/README.md).
 
+## Model Instance Kind
+
+Triton allows an execution of a model to be performed on both CPU, and GPU.
```
tanmayv25, line 1266: Triton model configuration allows users to provide kind to instance group settings<link:https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_configuration.md#instance-groups>. A python backend model can be written to respect the kind setting to control the execution of a model instance either on CPU or GPU.

In the ... we demonstrate how this can be achieved for your python model.

examples/instance_kind/README.md
```diff
@@ -0,0 +1,196 @@
...
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# Model Instance Kind Example
+
+Triton allows an execution of a model to be performed on both CPU, and GPU.
```
tanmayv25, line 31: Should clarify here as well.

examples/instance_kind/model.py
```diff
@@ -0,0 +1,70 @@
...
+class TritonPythonModel:
+
+    def initialize(self, args):
+        """
+        This function initializes pre-trained ResNet50 model,
+        depending on the value specified by an `instance_group` parameter
+        in `config.pbtxt`. 
+
+        Depending on what `instance_group` was specified in 
+        the config.pbtxt file (KIND_CPU, KIND_GPU, or both), the model instance
```
tanmayv25, line 42: KIND_CPU or KIND_GPU. remove `or both`.
This initialize function is called per model instance. One model instance will only carry KIND_CPU, KIND_GPU or KIND_MODEL. See here: https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto#L145-L182

tanmayv25: Also in line 47 we are only setting the device to "cuda" if the kind is GPU. otherwise to cpu. Hence, it is not both.

examples/instance_kind/README.md
```diff
@@ -0,0 +1,196 @@
...
+            "passive": false,
+            "host_policy": ""
+        }
+    ],
+    ...
+}
+Results is class: TABBY
+PASS: ResNet50 instance kind
+```
+It is also possible to load your python model to both CPU and GPU if neccessary.
```
tanmayv25, line 189: It is also possible to load multiple model instances on CPU and GPU if neccessary.

Below instance group setting will create two model instances, one on CPU and other on GPU.
```
instance_group [{ kind: KIND_CPU }, { kind: KIND_GPU}]
```

examples/instance_kind/model.py
oandreeva-nv: I was referring to the option of loading model instances, one on CPU and other on GPU. 
If 
```
instance_group [{ kind: KIND_CPU }, { kind: KIND_GPU}]
```
is present in the config, then one instance will be loaded to cpu, another instance to gpu. I'll remove `or both` to avoid confusion.

tanmayv25: The initialize function is under scope of the model instance. Hence, or both will confuse users.

oandreeva-nv: Thanks for rising this concern, agree with this

tanmayv25 (approved):

PR #211, opened 2023-03-08
Improve error message when BLS is used in 'initialize' or 'finalize' function

krishung5:
Need to also capture the BLS call in `finalize` function to avoid stub hanging during model unloading.

Added testing: https://github.com/triton-inference-server/server/pull/5469


src/infer_request.cc
```diff
@@ -373,9 +373,15 @@ InferRequest::GetResponseSender()
 std::shared_ptr<InferResponse>
 InferRequest::Exec(const bool is_decoupled)
 {
+  // BLS should not be used in "initialize" or "finalize" function.
+  std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
+  if (!stub->IsInitialized() || stub->IsFinalizing()) {
+    throw PythonBackendException(
+        "BLS is only supported during 'execute' function.");
```
Tabrizian, line 380: during "the" 'execute' function.

Tabrizian (approved):

PR #215, opened 2023-03-16
Pass InferPayload pointer to callback function to fix reinterpret_cast issue on shared_ptr

krishung5:
This PR fixed the value corrupted issue with the`std::shared_ptr<InferPayload>*` object.  `shared_ptr` is only thread safe when used by value (copied/moved). It's not thread safe when calling by reference.

### DIFF
--- a/src/request_executor.cc
+++ b/src/request_executor.cc
@@ -56,7 +56,7 @@ void
 InferResponseComplete(
     TRITONSERVER_InferenceResponse* response, const uint32_t flags, void* userp)
 {
-  auto p = reinterpret_cast<std::shared_ptr<InferPayload>*>(userp);
+  auto p = reinterpret_cast<InferPayload*>(userp);
   std::unique_ptr<InferResponse> infer_response;
   std::vector<std::shared_ptr<PbTensor>> output_tensors;
   std::shared_ptr<PbError> pb_error;
@@ -125,7 +125,7 @@ InferResponseComplete(
       output_tensors.clear();
     }
 
-    if (!(*p)->IsDecoupled()) {
+    if (!p->IsDecoupled()) {
       infer_response = std::make_unique<InferResponse>(
           output_tensors, pb_error, true /* is_last_response */);
     } else {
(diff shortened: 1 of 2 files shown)


src/request_executor.cc
```diff
@@ -56,7 +56,7 @@ void
 InferResponseComplete(
     TRITONSERVER_InferenceResponse* response, const uint32_t flags, void* userp)
 {
-  auto p = reinterpret_cast<std::shared_ptr<InferPayload>*>(userp);
+  auto p = reinterpret_cast<InferPayload*>(userp);
```
krishung5, line 59: When running on Jetson, the `std::shared_ptr<InferPayload>*` object after casting has different value when calling  `IsDecoupled()` and `IsPromiseSet()`. The corrupted value made the condition below calling `SetValueForPrevPromise` twice after the promise had been reset, which caused Segmentation fault. It works fine when building with Debug symbol but not in Release type, so I assume that there might be some race condition issue. Changing the cast type to simply `InferPayload*` help with the case.

Tabrizian: I think `shared_ptr` is thread-safe according to the C++ documentation. So I'm not sure why the value would be corrupted.
https://en.cppreference.com/w/cpp/memory/shared_ptr
> To satisfy thread safety requirements, the reference counters are typically incremented using an equivalent of [std::atomic::fetch_add](https://en.cppreference.com/w/cpp/atomic/atomic/fetch_add) with [std::memory_order_relaxed](https://en.cppreference.com/w/cpp/atomic/memory_order) (decrementing requires stronger ordering to safely destroy the control block).

dyastremsky: Please include this and any other detailed information about the PR in its description. That should act as the central repository of critical information about what this PR does and how.

krishung5: Synced with Iman offline. @dyastremsky Thanks for the reminder. I added the explanation to the description.

dyastremsky: Thanks! Is there any chance you could explain here as well? I'm curious to when shared_ptr is not thread safe for my learning. Is this a case where we're accessing the address directly and deallocating it, so the pointer is not aware that the underlying data is gone?

I don't love the idea of using raw pointers (or addresses) if we don't have to. Is this something where we need to code the reference differently, add mutexes, or use std::weak_ptr instead?

krishung5: Searched from chatgpt, it seems like `std::shared_ptr` is not thread-safe when different instances sharing the same underlying object:
> it's important to note that std::shared_ptr itself is thread-safe only when accessed through separate instances that do not share the same underlying object. If multiple threads access the same instance of std::shared_ptr and try to modify it concurrently, this can lead to data races and undefined behavior.

I think `std::weak_ptr` has the same limitation as well.

PR #218, opened 2023-03-21
Fix L0_backend_python timeout issue

krishung5:
When python backend sending BLS decoupled responses to the stub process, in some cases the stub process gets the following response before getting the first response, causing the `ResponseIterator` stop receiving the responses. For example, if a square model generates two responses, and the last response is placed before the actual first response, the `ResponseIterator` will think all the responses are fetched and stop receiving other responses.

This PR addresses the issue to make sure that the `ResponseIterator` object has the correct order of the first response.

### DIFF
--- a/src/pb_response_iterator.cc
+++ b/src/pb_response_iterator.cc
@@ -114,11 +114,11 @@ ResponseIterator::Iter()
 }
 
 void
-ResponseIterator::EnqueueResponse(std::unique_ptr<InferResponse> infer_response)
+ResponseIterator::EnqueueResponse(std::shared_ptr<InferResponse> infer_response)
 {
   {
     std::lock_guard<std::mutex> lock{mu_};
-    response_buffer_.push(std::move(infer_response));
+    response_buffer_.push(infer_response);
   }
   cv_.notify_one();
 }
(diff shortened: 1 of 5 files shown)


src/pb_response_iterator.cc
```diff
@@ -144,4 +144,19 @@ ResponseIterator::Clear()
   is_cleared_ = true;
 }
 
+std::shared_ptr<InferResponse>
+ResponseIterator::Pop()
```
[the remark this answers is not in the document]
krishung5: I think we couldn't directly use `Pop()` in `Next()`. `Pop()` is just to pop out all the existing responses in the `ResponseIterator` object, and when the response buffer is empty, it throws `py::stop_iteration`. It doesn't expect any coming responses. On the other hand, in `Next()` function, when the response buffer is empty, we call `cv.wait(lock)` and the cv will be notified once there is new request being enqueued to the buffer. It only throws  `py::stop_iteration` when the last response is received.

src/pb_stub.cc
```diff
@@ -1115,13 +1122,36 @@ Stub::ParentToStubServiceActive()
...
-      std::pair<void*, std::shared_ptr<ResponseIterator>>(
-          response_iterator->Id(), response_iterator));
+  if (response_iterator_map_.find(infer_response->Id()) !=
+      response_iterator_map_.end()) {
+    // Need to re-construct the 'ResponseIterator' and update the
+    // 'response_iterator_map_' to make sure the 'ResponseIterator' object has
+    // the correct first response.
+    auto response_iterator = std::make_shared<ResponseIterator>(infer_response);
+    bool done = false;
+    while (!done) {
```
Tabrizian, line 1136: Can we make `done` an argument of `Pop` instead of raising an exception? I think we shouldn't raise Python exceptions for C++ code.

krishung5: Would changing to `PythonBackendException` help? I agree that we shouldn't be throwing `py::` object.

Tabrizian: I think it might be best to add a new boolean to `Pop(done)`. If `done` is set to true, we will no longer call Pop. I think exceptions should be reserved for errors etc.

Tabrizian (approved):

tanmayv25 (approved):

PR #221, opened 2023-03-22
Add healthiness check to avoid hanging during model initialization

krishung5:
This PR adds healthiness check to fix the hanging issue when the stub process gets killed during function `initialize`.
Added test case: https://github.com/triton-inference-server/server/pull/5554
Fixes: https://github.com/triton-inference-server/server/issues/5381

### DIFF
--- a/src/stub_launcher.h
+++ b/src/stub_launcher.h
@@ -145,6 +145,9 @@ class StubLauncher {
   // Kill stub process
   void KillStubProcess();
 
+  // Get a message from the stub process
+  TRITONSERVER_Error* ReceiveMessageFromStub(off_t& message);
+
  private:
   pid_t parent_pid_;
   pid_t stub_pid_;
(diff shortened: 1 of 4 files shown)


src/stub_launcher.h
```diff
@@ -145,6 +145,9 @@ class StubLauncher {
   // Kill stub process
   void KillStubProcess();
 
+  // Get a message from the stub process
+  TRITONSERVER_Error* ReceiveMessageFromStub(off_t& message);
```
Tabrizian, line 149: Use `bi::managed_external_buffer::handle_t` instead of `off_t`.

Tabrizian (commented):
One minor comment, otherwise looks good!

src/stub_launcher.h
krishung5: Good catch. Fixed.

Tabrizian (dismissed):

Tabrizian (approved):

PR #222, opened 2023-03-23
Initialize CUDA driver API before using it

Tabrizian:
Looks like there were some changes in the CUDA driver API that is affecting our GPU tensor support.

### DIFF
--- a/src/pb_tensor.cc
+++ b/src/pb_tensor.cc
@@ -1,4 +1,4 @@
-// Copyright 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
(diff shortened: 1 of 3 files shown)


krishung5 (changes requested):
Only need to update the copyright, otherwise LGTM!

oandreeva-nv:
Same on copyright and I have same question as Tanmay, do we have a test to catch API changes? otherwise LGTM!

Tabrizian:
> do we have a test to catch API changes?

We do have a test for GPU tensors but for this bug it seems like something has changed in the CuPy/CUDA side. We should call `cuInit` before using any driver API.

krishung5:
I think we also need to change the copyright in `src/pb_tensor.cc` as well.

tanmayv25 (approved):

tanmayv25 (approved):

krishung5 (approved):

PR #223, opened 2023-03-24
Update dlpack implementation for PbTensor

oandreeva-nv:
This PR updates DLPack implementation for PbTensor according to Array API and DLPack v0.8 requirements.

Major changes:
 - Added `__dlpack__` and `__dlpack_device__` implementations according to Array API standard
   + [Python Specification for DLPack](https://dmlc.github.io/dlpack/latest/python_spec.html#python-specification-for-dlpack)
   + [Array API specification] (https://data-apis.org/array-api/latest/API_specification/index.html)
 -  Current implementation is backward-compatible, i.e. `PbTensor.from_dlpack()` accepts PyCapsule (as before) and external tensor directly. `PbTensor::FromDLPack` takes care of that. Former `PbTensor::FromDLPack` -> `PbTensor::FromDLPackCapsule`
 -  `PbTensor::DLPack`, which is bind to `__dlpack__(self, Optional[Union[int, Any]] stream)` make sure to synchronize streams, if needed (passed stream is not a default stream, not -1, and not a legacy stream). Since PbTensor does not support async execution, if `DLPack` receives default stream, sync is not needed.
 - I've implemented some refactoring, i.e. separated `switch` into `PbTensor::DeviceType`, because I realized I need this for `DLPackDevice` 
 - `DLPackDevice` has returned type `std::pair<int32_t, int64_t>` with the following in mind. [This](https://github.com/pybind/pybind11/issues/3858) isn't finished yet. When it is done, a more elegant type description can be put instead. For now I used types, which are indicated in the code. Open to any suggestions/alternatives
 - Added bool support

Fixes issue: [3944](https://github.com/triton-inference-server/server/issues/3944)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1223,7 +1223,7 @@ class TritonPythonModel:
     # tensor.
     input0 = pb_utils.Tensor.from_dlpack("INPUT0", to_dlpack(pytorch_tensor))
 ```
-Starting from 23.04 release, Python backend allows tensors implementing
+Python backend allows tensors implementing
 [`__dlpack__`](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.array.__dlpack__.html) 
 and [`__dlpack_device__`](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.array.__dlpack_device__.html) 
 [interface](https://dmlc.github.io/dlpack/latest/python_spec.html) 
--- a/README.md
+++ b/README.md
@@ -1223,6 +1223,15 @@ class TritonPythonModel:
     # tensor.
     input0 = pb_utils.Tensor.from_dlpack("INPUT0", to_dlpack(pytorch_tensor))
 ```
+Starting from 23.04 release, Python backend allows tensors implementing
+[`__dlpack__`](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.array.__dlpack__.html) 
+and [`__dlpack_device__`](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.array.__dlpack_device__.html) 
+[interface](https://dmlc.github.io/dlpack/latest/python_spec.html) 
+to be converted to Python backend tensors. For instance:
+
+```python
+input0 = pb_utils.Tensor.from_dlpack("INPUT0", pytorch_tensor)
+```
 
 This method only supports contiguous Tensors that are in C-order. If the tensor
 is not C-order contiguous an exception will be raised.
--- a/src/pb_tensor.cc
+++ b/src/pb_tensor.cc
@@ -366,9 +366,14 @@ PbTensor::FromDLPack(const std::string& name, const py::object& tensor)
     }
     // In case there is a pending job on the data, where this capsule
     // is pointing to, we need to wait for it before consuming.
-    // This is important for when data is located in different
-    // context (GPU) or work is done on non-blocking streams.
-    err = cudaDeviceSynchronize();
+    // This is important for when data is located on different
+    // context (GPU) and work is done on the default stream.
+    // For this scenario, __dlpack__ implementation may skip 
+    // syncronization (since the work is on the default stream)
+    // and we will return pointer to the data on different GPU too early 
+    // (i.e. before pending work is done). Thus we sync on the default stream
+    // only in the case we switched to a different context.
+    err = overridden ? cudaStreamSynchronize(0) : cudaSuccess;
     if (err != cudaSuccess) {
       throw PythonBackendException(
           "Failed to synchronize CUDA device with id " +
@@ -399,6 +404,12 @@ PbTensor::FromDLPack(const std::string& name, const py::object& tensor)
         "DLPack capsule passed pointer to memory allocated on GPU device, \
           when GPU is not available");
 #endif
+  } else if (
+      capsule_device_info.first != DLDeviceType::kDLCPU &&
+      capsule_device_info.first != DLDeviceType::kDLCUDAHost) {
+    throw PythonBackendException(
+        "DLDevice type " + std::to_string(capsule_device_info.first) +
+        " is not support by Python backend.");
   }
 
   // If data is located on CPU, `stream=None` is the only accepted argument
--- a/src/pb_tensor.cc
+++ b/src/pb_tensor.cc
@@ -31,7 +31,6 @@
 #ifdef TRITON_PB_STUB
 #include "pb_stub_utils.h"
 namespace py = pybind11;
-using namespace py::literals;
 #endif
 #include "pb_tensor.h"
 
@@ -255,26 +254,10 @@ PbTensor::DeviceType()
 py::capsule
 PbTensor::DLPack(const py::object& stream)
 {
-// Array API requirements for the stream argument:
-// stream = None, producer must assume the legacy default stream,
-// stream = -1 is a signal for the producer not to perform any synchronization
-// stream = 1 the legacy default stream (in this case should synchronize
-// on CUDA stream 0)
-#ifdef TRITON_ENABLE_GPU
-  if (!IsCPU() && !stream.is(py::int_(-1))) {
-    cudaStream_t cuda_stream =
-        (stream.is(py::int_(1)) || stream.is(py::none()))
-            ? 0
-            : reinterpret_cast<cudaStream_t>(py::cast<uint64_t>(stream));
-    cudaError_t err = cudaStreamSynchronize(cuda_stream);
-    if (err != cudaSuccess) {
-      throw PythonBackendException(
-          "Failed to syncronize streams before PbTensor's dlpack capsule "
-          "creation.");
-    }
-  }
-#endif
-
+  // Here external tensor requests PbTensor's `__dlpack__` method to provide
+  // a PyCapsule. By the design of PbTensor, in a GPU case no pending work 
+  // is scheduled to work with PbTensor's data and we can simply pass 
+  // the capsule.
   return this->ToDLPack();
 }
 
@@ -356,6 +339,20 @@ PbTensor::FromDLPack(const std::string& name, const py::object& tensor)
   if (py::isinstance<py::capsule>(tensor)) {
     return FromDLPackCapsule(name, tensor);
   } else if (py::hasattr(tensor, "__dlpack__")) {
+#ifdef TRITON_ENABLE_GPU
+  cudaError_t err = cudaStreamSynchronize(0);
+  if (err != cudaSuccess) {
+    throw PythonBackendException(
+        "Failed to syncronize on the default stream before\
+             dlpack capsule consumption.");
+  }
+#endif
+    // Array API requirements for the stream argument:
+    // stream = None, producer must assume the legacy default stream,
+    // stream = -1 is a signal for the producer not to perform any 
+    //          synchronization
+    // stream = 1 the legacy default stream (in this case should synchronize
+    //          on CUDA stream 0)
     // Python backend does not support async executions on PbTensor objects,
     // thus we can count on the default `stream=None` argument.
     // For CPU, `stream=None` is the only accepted argument
@@ -364,7 +361,7 @@ PbTensor::FromDLPack(const std::string& name, const py::object& tensor)
     // Reference:
     // https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html
     return FromDLPackCapsule(
-        name, tensor.attr("__dlpack__")("stream"_a = py::none()));
+        name, tensor.attr("__dlpack__")(py::arg("stream") = py::none()));
   } else {
     throw PythonBackendException(
         "Provided tensor is not supported.\
@@ -376,6 +373,11 @@ std::shared_ptr<PbTensor>
 PbTensor::FromDLPackCapsule(
     const std::string& name, const py::capsule& dlpack_tensor)
 {
+
+// TO-DO ADD sync on the default stream either here for all apis 
+// or in __dlpack__ case for only new apis. Write tests and think about 
+// different contexts.
+
 #ifdef TRITON_ENABLE_GPU
   cudaError_t err = cudaStreamSynchronize(0);
   if (err != cudaSuccess) {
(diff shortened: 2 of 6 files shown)


src/pb_tensor.cc
```diff
@@ -305,11 +339,30 @@ PbTensor::Memory()
...
+    // Python backend does not support async executions on PbTensor objects,
+    // thus we can count on the default `stream=None` argument.
+    // For CPU, `stream=None` is the only accepted argument 
+    // according to array API. For GPU, when `stream=None`  producer must 
+    // assume the legacy default stream.
+    // Reference:
+    // https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html
+    return FromDLPackCapsule(name, tensor.attr("__dlpack__")());
+  }else{
+    throw PythonBackendException("Provided tensor is not supported.");
```
Tabrizian, line 359: Perhaps mention that the tensor must be dlpack capsule or have a `__dlpack__` attribute.

src/pb_tensor.cc
```diff
@@ -305,11 +339,30 @@ PbTensor::Memory()
...
+    return FromDLPackCapsule(name, tensor);
+  } else if (py::hasattr(tensor, "__dlpack__")) {
+    // Python backend does not support async executions on PbTensor objects,
+    // thus we can count on the default `stream=None` argument.
+    // For CPU, `stream=None` is the only accepted argument 
+    // according to array API. For GPU, when `stream=None`  producer must 
+    // assume the legacy default stream.
+    // Reference:
+    // https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html
+    return FromDLPackCapsule(name, tensor.attr("__dlpack__")());
```
Tabrizian, line 357: I think based on the API descriptions we should provide -1?

>  If stream is -1, the value may be used by the consumer to signal “producer must not perform any synchronization”.

README.md
```diff
@@ -1199,6 +1199,13 @@ class TritonPythonModel:
     # tensor.
     input0 = pb_utils.Tensor.from_dlpack("INPUT0", to_dlpack(pytorch_tensor))
 ```
+Starting from 23.04 release, Python backend allows external tensors being passed 
+directly to `pb_utils.Tensor.from_dlpack()` without an explicit DLPack encoding. 
+For instance:
```
Tabrizian, line 1204: Starting from 23.04 release, Python backend allows tensors that implement `__dlpack__` and `__dlpack_device__` interface to be converted to Python backend tensors.

Tabrizian (changes requested):

src/pb_tensor.cc
oandreeva-nv: We can do this. The way it is now , we are telling to the producer that he can assume default stream, which we are on now. I believe this is also an indication, that sync is not required, since python backend does not allow async execution.

Tabrizian: I didn't infer that from the API description. It just mentions that the tensor would be on the default stream but the consumer would still need to synchronize even if it is on the default stream.

oandreeva-nv: I see, I believe PyTorch treats the passed `None` as no sync as well, I probably was influenced by their implementation a little bit. I'll make a change to pass `-1` for no sync.

src/pb_tensor.cc
```diff
@@ -305,11 +339,30 @@ PbTensor::Memory()
...
-PbTensor::FromDLPack(const std::string& name, const py::capsule& dlpack_tensor)
+PbTensor::FromDLPack(const std::string& name, const py::object& tensor)
 {
   if (name == "") {
     throw PythonBackendException("Tensor name cannot be an empty string.");
   }
+  if(py::isinstance<py::capsule>(tensor)){
+    return FromDLPackCapsule(name, tensor);
+  } else if (py::hasattr(tensor, "__dlpack__")) {
+    // Python backend does not support async executions on PbTensor objects,
```
krishung5, line 350: I was wondering if this is documented somewhere? Just wanted to read more about it but couldn't find more information in README file.

krishung5 (commented):
nit: seems like the copyright for some of the files needs to be updated. e.g. `src/pb_stub_utils.cc`, `src/pb_tensor.h` ...

src/pb_tensor.cc
oandreeva-nv: @Tabrizian we've discussed offline that async execution is not supported. Do we have it explicitly stated in docs?

src/pb_tensor.cc
```diff
@@ -231,6 +231,44 @@ PbTensor::FromNumpy(const std::string& name, py::array& numpy_array)
...
+{
+  // Array API requirements for the stream argument:
+  // stream = None  producer must assume the legacy default stream, no sync 
+  // stream = -1 is a signal for the producer not to perform any synchronization
+  // stream = 1 the legacy default stream, since Python Backend does not support 
+  // async executions on PbTensor objects, this case also does not require 
+  // stream sync.
+  if (!stream.is(py::none()) && !stream.is(py::int_(-1)) 
+      && !stream.is(py::int_(1))) {
+    cudaStreamSynchronize(
```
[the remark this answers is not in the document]
oandreeva-nv: Added in Revision 1

src/pb_tensor.cc
oandreeva-nv: Added in Revision 1

src/pb_tensor.cc
oandreeva-nv: I'm passing `stream=None` in Revision 1, this way if tensor is on CPU, this is the only option that allowed, and if a tensor is on GPU, legacy stream is assumed according to API standard

krishung5 (dismissed):
LGTM! I'll defer the final approval to Iman.

src/pb_tensor.cc
```diff
@@ -31,6 +31,7 @@
 #ifdef TRITON_PB_STUB
 #include "pb_stub_utils.h"
 namespace py = pybind11;
+using namespace py::literals;
```
Tabrizian, line 34: I would rather not use `using namespace` since it can pollute the symbol space and create conflicts. You can create aliases for easier usability. For example, `namespace pyl = py::literals`.

src/pb_tensor.cc
```diff
@@ -305,12 +348,42 @@ PbTensor::Memory()
...
 std::shared_ptr<PbTensor>
-PbTensor::FromDLPack(const std::string& name, const py::capsule& dlpack_tensor)
+PbTensor::FromDLPack(const std::string& name, const py::object& tensor)
 {
   if (name == "") {
     throw PythonBackendException("Tensor name cannot be an empty string.");
   }
+  if (py::isinstance<py::capsule>(tensor)) {
+    return FromDLPackCapsule(name, tensor);
+  } else if (py::hasattr(tensor, "__dlpack__")) {
```
Tabrizian, line 358: Do we need to make sure it implements both `__dlpack__` and `__dlpack_device__`? It looks like we don't need `__dlpack_device__` when python backend is the consumer since the same information is available in dlpack capsule?

src/pb_tensor.cc
```diff
@@ -305,12 +348,42 @@ PbTensor::Memory()
...
+      Tensor must be a DLPack capsule or have a `__dlpack__` attribute");
+  }
+}
 
+std::shared_ptr<PbTensor>
+PbTensor::FromDLPackCapsule(
+    const std::string& name, const py::capsule& dlpack_tensor)
+{
+#ifdef TRITON_ENABLE_GPU
+  cudaError_t err = cudaStreamSynchronize(0);
```
Tabrizian, line 380: We only need the synchronization when the tensor type is `GPU`. We should add a condition around this so that it is executed only on GPU tensors. Also, what happens in the multi-GPU case? Do we need to change the context to the correct GPU before synchronization?

src/pb_tensor.cc
```diff
@@ -231,6 +232,52 @@ PbTensor::FromNumpy(const std::string& name, py::array& numpy_array)
...
+// stream = -1 is a signal for the producer not to perform any synchronization
+// stream = 1 the legacy default stream (in this case should synchronize
+// on CUDA stream 0)
+#ifdef TRITON_ENABLE_GPU
+  if (!IsCPU() && !stream.is(py::int_(-1))) {
+    cudaStream_t cuda_stream =
+        (stream.is(py::int_(1)) || stream.is(py::none()))
+            ? 0
+            : reinterpret_cast<cudaStream_t>(py::cast<uint64_t>(stream));
+    cudaError_t err = cudaStreamSynchronize(cuda_stream);
```
Tabrizian, line 269: We shouldn't synchronize their stream. I think to guarantee stream ordering we don't have to do anything with their stream since the pointers are already synchronized.

oandreeva-nv: @Tabrizian  I don't think that pointers are already synchronized at this point. `DLPack` is invoked by external tensor's `__dlpack__` method. We want to make sure that pointer contains correct data and no pending job is left, that can corrupt this data, right? if not `cudaStreamSunchronize`, we should at least create an event on their stream and wait for this event on our stream.

src/pb_tensor.cc
oandreeva-nv: We don't need to ensure __dlpack_device__ is implemented here, since we don't call it explicitly, and it is true, capsule should contain device info.

src/pb_tensor.cc
oandreeva-nv: I don't think that we can assume that pointers are synchronized at this point. When external stream calls pbTensor's `__dlpack__` method, which invokes this function, I believe we are not guaranteed that pointers will be sync'ed.  In older DLPack implementations with intermediate capsule, I believe some implementations assumed that `to_dlpack` and `from_dlpack` are on the same stream, thus they are aligned. in here, `__dlpack__` is called on the external tensor's stream and is accessing `pbTensor` on the legacy stream, so we must ensure that any pending operations on the default stream are finished. I agree that  current implementation is not the right way of ensuring that, but we still need to make sure that all work on `pbTensor` is done, before passing it to external tensor

src/pb_tensor.cc
```diff
@@ -305,12 +331,89 @@ PbTensor::Memory()
...
+    // is pointing to, we need to wait for it before consuming.
+    // This is important for when data is located in different
+    // context (GPU) or work is done on non-blocking streams.
+    err = cudaDeviceSynchronize();
+    if (err != cudaSuccess) {
+      throw PythonBackendException(
+          "Failed to synchronize CUDA device with id " +
+          std::to_string(
+              overridden ? capsule_device_info.second : current_device));
+    }
```
Tabrizian, line 377: Do we still need this synchronization even if we are passing default stream to `__dlpack__`?

src/pb_tensor.cc
```diff
@@ -305,12 +331,89 @@ PbTensor::Memory()
...
+          "with id " +
+          std::to_string(current_device));
+    }
+    return ptr_to_tensor;
+#else
+    throw PythonBackendException(
+        "DLPack capsule passed pointer to memory allocated on GPU device, \
+          when GPU is not available");
+#endif
+  }
```
Tabrizian, line 402: We need to make sure that the device type is CPU. DLPack supports other device types too which we are not able to handle.

Tabrizian:
> My issue is, it is quite enforcing and may have an impact on the speed, as an example in BLS cases probably? As an alternative, we can make this sync conditional on either an environment label, or parameter, set with --backend-config flag. Let me know what you think

I was curious why the sync was required. Did you observe a failing test without synchronization?

oandreeva-nv:
@Tabrizian Yes, in the case of multi-gpu multi-stream, when streams are non - blocking, I encountered, that without synchronization, this test fails and we don't get an expected data.

oandreeva-nv:
@Tabrizian I found out that an issue in observed failing tests maybe not to the lack of synchronization in the `from_dlpack` method, but rather something is impacting tests, when they ran together. In isolation each test is passing without device sync. Currently investigating an issue

oandreeva-nv:
@Tabrizian It's ready for re-review. The only time we need synchronization is when we need to change context and there is a pending work on the default stream. Since standard does not say to sync on default stream, not everyone does. Then in this case we can return earlier, than work is done, thus `cudaStreamSync(0)` is added. I've also filed a ticket to investigate a strange behavior with cupy , multi-gpu non-blocking streams : DLIS-4887. For now, I run this test on a different second GPU.

src/pb_tensor.cc
```diff
@@ -305,12 +331,100 @@ PbTensor::Memory()
...
+    // In case there is a pending job on the data, where this capsule
+    // is pointing to, we need to wait for it before consuming.
+    // This is important for when data is located on different
+    // context (GPU) and work is done on the default stream.
+    // For this scenario, __dlpack__ implementation may skip 
+    // syncronization (since the work is on the default stream)
+    // and we will return pointer to the data on different GPU too early 
+    // (i.e. before pending work is done). Thus we sync on the default stream
+    // only in the case we switched to a different context.
+    err = overridden ? cudaStreamSynchronize(0) : cudaSuccess;
```
Tabrizian, line 376: I'm not sure why synchronization is needed. Looking at CuPy's dlpack implementation it doesn't seem like they are synchronizing.

oandreeva-nv: This is based on testing and researching: when we need to switch devices, e.g. from GPU:0 to GPU:1, to consume the capsule, if there is a pending work on the default stream on GPU:1, wait-for-the-event trick will not be invoked, since the external tensor is on the default stream and cupy and pytorch do not call for  wait-for-the-event trick if the tensor and passed stream are on the default stream. If we don't sync on the default stream, we just consume capsule and switch back, so situation as in [multi-gpu](https://github.com/triton-inference-server/server/pull/5550/files#diff-31d73050a3564c33d370d2da0252b7a797412a27d98f33266b4c8d8205a046a9R225) test will fail, i.e. pending job on that tensor on the default stream

src/pb_tensor.cc
oandreeva-nv: I've added a check on L407-L409 to throw an exception, if device is not CPU or CUDAHost(which we support)

Tabrizian (approved):

oandreeva-nv:
Note for this PR: there is a necessity to follow up with CuPY and PyTorch to investigate why they do not (?) handle stream=1, i.e. do not convert it to the default stream (legacy stream)

PR #231, opened 2023-04-13
Make changes to python backend to run on inf2 and trn1 devices

nskool:
1. Add setup script changes to use the updated runtime for inf2 and trn1
2. Incoming:
 - container script changes 
 - generate model script changes
3. Separate inf2/inf1 scripts will be merged and supported via cmdline to specify which device to generate model for

Pending:
1. Doc updates
2. Updates to qa folder Dockerfile to allow for testing model generation script


GuanLuo:
Hi @nskool, thanks for working on the inferentia improvement, I notice that you are also making changes with respect to inf1, I wonder if you have tested the change fixes inf1 deployment as well? There was report regarding the inf1 scripts are outdated.

GuanLuo:
Hi @nskool , there is change on inf1 setup merged recently (https://github.com/triton-inference-server/python_backend/pull/236), just an reminder that you may need to resolve conflicts regarding to that, and feel free to let us know that whether the inf1 setup needs to be improved.

nskool:
@GuanLuo can you help review

inferentia/scripts/gen_triton_model.py
```diff
@@ -270,20 +271,27 @@ def get_tensorflow_initialize_impl():
...
+        tf.keras.models.load_model(compiled_model)
+        for _ in range(cores_per_instance)
+        ] * threads_per_core 
+'''
+    else:
+        init_impl += '''
         self.pred_list = [
-            tf.contrib.predictor.from_saved_model(compiled_model)
-            for _ in range(cores_per_instance)
+        tf.contrib.predictor.from_saved_model(compiled_model)
```
chen3933, line 287: miss indentation here. You will get `PEP 8: E122 continuation line missing indentation or outdented` when running format check on generated `model.py`

GuanLuo:
Looks like there is conflict in `inferentia/scripts/setup.sh`, can you look into resolving it?

inferentia/scripts/gen_triton_model.py
nskool: Thanks, updated

nskool:
> Looks like there is conflict in `inferentia/scripts/setup.sh`, can you look into resolving it?

Thanks, resolved - it was the part dependent on conda/python stub, I don't think it's needed if pip is being used - it works as expected. Let me know if you think otherwise.

chen3933:
LGTM

inferentia/scripts/setup.sh
```diff
@@ -115,53 +131,49 @@ apt-get update && \
...
-    if [ $TENSORFLOW_VERSION -eq 1 ]; then
-        # Install TensorFlow Neuron
-        python -m pip install tensorflow-neuron[cc]==1.15.5.* "protobuf"
-    else
-        # Install TensorFlow Neuron
-        python -m pip install tensorflow-neuron[cc] "protobuf"
+    if [ ${INSTALL_INF1} -eq 1] && [ ${TENSORFLOW_VERSION} -eq 1 ]; then
+        pip install --upgrade tensorflow-neuron==1.15.5.* neuron-cc "protobuf<4" tensorboard-plugin-neuron
+    elif [ ${INSTALL_INF1} -eq 1]; then
+        pip install --upgrade tensorflow-neuron[cc] "protobuf<4"
```
GuanLuo, line 160: @nskool do you know why `tensorboard-plugin-neuron` is not installed if inf1 and TF2 is set? And why `protobuf<4` is specified?

I am validating the case when inf1 and tf1 is set and there is dependency w.r.t. protobuf version, grpc dependency seems to be coming from `tensorboard-plugin-neuron` (`Requirement already satisfied: grpcio>=1.48.2 in /usr/local/lib/python3.8/dist-packages (from tensorboard>=1.15.0->tensorboard-plugin-neuron) (1.54.0)`)
```
grpcio-channelz 1.54.0 requires protobuf>=4.21.6, but you have protobuf 3.20.1 which is incompatible.
```

GuanLuo: some update, looks like TF1 has shipped with protobuf 3.x dependency. I wonder if `tensorboard-plugin-neuron` is required or can it be pinned to a version works with protobuf 3.x if exists such an version.

nskool: @GuanLuo thanks - I see that `tensorboard-plugin-neuron` was removed in this update - https://github.com/triton-inference-server/python_backend/pull/236/files. Do you happen to know the reason for that? It seems I can remove tensorboard plugin altogether as per the neuron documentation and also unpin the protobuf version - https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/quick-start/torch-neuron.html

I'm unaware of the protobuf dependency, but have chosen to retain the previous default configuration.

GuanLuo: The removal is probably unintended, however, from the [original discussion](https://github.com/triton-inference-server/python_backend/pull/92/files#r743935898) of adding this dependency, it was added just because it is being mentioned in the quickstart guide, i.e.:
```
# Install Neuron TensorBoard
python -m pip install tensorboard-plugin-neuron
```
Looking at the use of TensorBoard as a plugin, I think it is only used for further profiling the model and not necessary for running inference on Inferentia (current test runs fine without it). So I think we can remove the TensorBoard dependency to simplify the script if you agree that should not have significant impact.

nskool: @GuanLuo I rebased with mainline, and added tensorboard-plugin for both TF1 and TF2 as per neuron instructions. That is correct, it is only for additional profiling. I have unpinned the protobuf dependency, and updated the readme as well.

GuanLuo:
@nskool I have made some changes to fix the script format and actually remove TensorBoard dependency for inf1 (possibly can do the same with inf2), do you mind to apply it to your branch? I think at least for TF1, the TensorBoard dependency needs to be remove as it seems to be installing different protobuf version (both with and without pinning version) which will cause issue when using TF at later stage. https://github.com/triton-inference-server/python_backend/commit/6c70acd0a7e42cbf6c848ebfeb5a50b6c4198035

nskool:
> @nskool I have made some changes to fix the script format and actually remove TensorBoard dependency for inf1 (possibly can do the same with inf2), do you mind to apply it to your branch? I think at least for TF1, the TensorBoard dependency needs to be remove as it seems to be installing different protobuf version (both with and without pinning version) which will cause issue when using TF at later stage. [6c70acd](https://github.com/triton-inference-server/python_backend/commit/6c70acd0a7e42cbf6c848ebfeb5a50b6c4198035)

@GuanLuo That makes sense to me, thanks, I've updated the PR

GuanLuo:
@nskool The test passed, let me know if the PR is ready to be merged

nskool:
> @nskool The test passed, let me know if the PR is ready to be merged

Thanks @GuanLuo, yes, The PR can be merged.

PR #233, opened 2023-04-18
Add TF determinism section, fix TF32 typo, remove PyTorch 2.0 note

rmccorm4:
Adding similar TF determinism section based on https://github.com/triton-inference-server/server/issues/5640#issuecomment-1512608861

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ any C++ code.
   - [Input Tensor Device Placement](#input-tensor-device-placement)
 - [Frameworks](#frameworks)
   - [PyTorch](#pytorch)
+  - [TensorFlow](#tensorflow)
 - [Examples](#examples)
   - [AddSub in NumPy](#addsub-in-numpy)
   - [AddSubNet in PyTorch](#addsubnet-in-pytorch)
@@ -1270,7 +1271,7 @@ this workflow.
 For a simple example of using PyTorch in a Python Backend model, see the
 [AddSubNet PyTorch example](#addsubnet-in-pytorch).
 
-### Determinism and Reproducibility
+### PyTorch Determinism 
 
 When running PyTorch code, you may notice slight differences in output values
 across runs or across servers depending on hardware, system load, driver, or even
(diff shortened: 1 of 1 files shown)


tanmayv25 (dismissed):

README.md
```diff
@@ -1294,6 +1300,17 @@ relies on LibTorch/TorchScript (C++) which has been deprecated from
 So, users interested in new features introduced in PyTorch 2.0 should try the
```
Tabrizian, line 1300: The current changes look good but I think we should remove the deprecation note from the PyTorch 2.0 section. I don't think there has been any announcements about deprecating torch script.

README.md
```diff
@@ -1270,19 +1271,24 @@ this workflow.
...
-FP32 operations called TensorFlow32 (TF32). Typically this optimization will
-improve overall performance at the cost of minor precision loss, but similarly
-this precision loss is acceptable for most model predictions. For more info on
-TF32 in PyTorch and how to enable/disable it as needed, see 
+FP32 operations called 
+[TensorFloat32 (TF32)](https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/).
+Typically this optimization will improve overall performance at the cost of
+minor precision loss, but similarly this precision loss is acceptable for most
+model predictions. For more info on TF32 in PyTorch and how to enable/disable
+it as needed, see
```
Tabrizian, line 1291: I think what we want to highlight here is that TF32 is enabled by default (need to confirm) in PyTorch when using PyTorch backend but it is disabled in the upstream PyTorch and that's why the results could be different. https://pytorch.org/docs/stable/notes/cuda.html#tensorfloat-32-tf32-on-ampere-devices

README.md
rmccorm4: My original note was this, rather than claiming any specific support/deprecation:
> Currently, the [PyTorch Backend](https://github.com/triton-inference-server/pytorch_backend)
doesn't support PyTorch 2.0, so users should try the Python Backend for these
models instead.

I do think we should point out PyTorch 2.0 in some way though.

What do you think?

Tabrizian: Also, I think we are using PyTorch 2.0 in PyTorch backend. It is just that `torch.compile` and other optimizations cannot be enabled in PyTorch backend.

README.md
Tabrizian: Makes sense. Resolving.

README.md
tanmayv25: Exactly. Technically, we support pytorch 2.0 in our PyTorch backend. However, as the backend depends upon TorchScript which is no longer being maintained, users interested in new features introduced via pytorch 2.0 can not access them via pytorch backend.

tanmayv25: >  I don't think there has been any announcements about deprecating torch script.

If that's true we should just remove the deprecation note.

tanmayv25: Currently, the
[PyTorch Backend](https://github.com/triton-inference-server/pytorch_backend)
relies on LibTorch/TorchScript (C++). So, users interested in new workflows introduced in PyTorch 2.0 should try the ...

rmccorm4: I will remove the PyTorch 2.0 notes for now until we start seeing more usage. Related PR: https://github.com/triton-inference-server/pytorch_backend/pull/101

Tabrizian (approved):

PR #235, opened 2023-04-20
Implement custom metrics API

krishung5:
This PR adds the custom metrics API for Python backend. Two sets of files are added, `metric_family.h` and `metric.h`, representing the `MetricFamily` and `Metric` objects.
- Both classes have the functionality to read/write in SHM for the communication between stub process and the Python backend.
- `TRITONSERVER_MetricFamily`/ `TRITONSERVER_Metric` is wrapped inside class  `MetricFamily`/`Metric` as a void pointer, and will be cast to the correct type to be used in Triton C API calls in the parent process.

The custom metrics API can be called from anywhere in the model script as the stub process uses the dedicated `utils message queue`  to communicate with the parent process for custom metrics message, just like logging and BLS decoupled messages.

- [x] Add documentation and example
- [x] Add testing: https://github.com/triton-inference-server/server/pull/5669

Closes: https://github.com/triton-inference-server/server/issues/5227


src/pb_utils.h
```diff
@@ -165,18 +167,37 @@ struct ResponseBatch : SendMessageBase {
...
+};
+
+struct CustomMetricsMessage : SendMessageBase {
+  bi::managed_external_buffer::handle_t message;
+  bool has_error;
+  bool is_error_set;
+  bi::managed_external_buffer::handle_t error;
+  double value;
+};
+
```
Tabrizian, line 200: We should move these enums to `pb_custom_metric.cc` and `custom_metrics_registery.cc`.

src/custom_metrics_registry.cc
```diff
@@ -0,0 +1,163 @@
...
+    const MetricKind& kind)
+    : name_(name), description_(description), kind_(kind),
+      metric_family_(nullptr)
+{
+  TRITONSERVER_MetricKind triton_kind = ToTritonServerMetricKind(kind);
+  THROW_IF_TRITON_ERROR(TRITONSERVER_MetricFamilyNew(
+      &metric_family_, triton_kind, name_.c_str(), description_.c_str()));
+}
+
+MetricFamilyRegistry::~MetricFamilyRegistry()
```
Tabrizian, line 114: Just a suggestion, I was wondering do you think it might be better to have a file for `metrics.cc` and `metrics_family.cc`? I don't fully understand the reasoning behind the current breakdown of files. Each of them can have their own `SaveToSharedMemory` and `LoadFromSharedMemory` functionalities.

src/custom_metrics_registry.cc
```diff
@@ -0,0 +1,163 @@
...
+MetricFamilyRegistry::AddMetric(std::unique_ptr<PbCustomMetric> metric)
+{
+  std::lock_guard<std::mutex> lock(metric_map_mu_);
+  if (metric_map_.find(metric->Labels()) != metric_map_.end()) {
+    throw PythonBackendException("Metric with the same labels already exists.");
+  } else {
+    std::unique_ptr<MetricRegistry> pb_metric =
+        std::make_unique<MetricRegistry>(metric->Labels(), metric_family_);
+    metric_map_.insert({metric->Labels(), std::move(pb_metric)});
+  }
```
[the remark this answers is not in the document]
rmccorm4: Agreed, core should return a reference to the existing metric if a new metric with duplicate labels is added to an existing metric family: https://github.com/triton-inference-server/core/blob/ecbe0230d5770481d23c5f1f984ac6fac651e1df/src/test/metrics_api_test.cc#L111-L121. 

I don't think an exception should be thrown for duplicate labels and same kind, if I'm understanding correctvly.

Actually, I'm surprised that your test isn't throwing a `PythonBackendException` here when testing duplicate labels: https://github.com/triton-inference-server/server/blob/6890ad9312b5e8e9753daa6117922b1a499701b8/qa/python_models/custom_metrics/model.py#L76-L78. Is it?

src/custom_metrics_registry.cc
krishung5: Do you mean to combine the `custom_metrics_registry.h` and `pb_custom_metrics.h` classes to just one metric/metric_family object or do you mean we should just put metric and metric_family into different files? The reason behind having  `custom_metrics_registry.h` and `pb_custom_metrics.h` files is just that I thought it would be clearer to use two different objects instead of using a couple of `#ifdef` in the file.

krishung5: After some thought, I think maybe it would be better to combine both of the objects to just one `metric`/`metric_family`. I was thinking about how to read/write the `TRITONSERVER_MetricFamily`/`TRITONSERVER_Metric` into SHM in both stub and parent processes if we only have one `metric`/`metric_family` object, but it seems like we can just store the memory address like what we do for [`response_factory`](https://github.com/triton-inference-server/python_backend/blob/ad88027900392f7f481b1c9960fc386a32677f08/src/infer_request.h#L137).

Tabrizian: >Do you mean to combine the custom_metrics_registry.h and pb_custom_metrics.h classes to just one metric/metric_family object or do you mean we should just put metric and metric_family into different files?

I meant to put metric and metric family into two different files. I think combining them to just one file should be fine too.

> Do you mean to combine the custom_metrics_registry.h and pb_custom_metrics.h classes to just one metric/metric_family object or do you mean we should just put metric and metric_family into different files?

Personally, I think the OOP look-and-feel is more important. For example, if I want to look for the metric related changes they would be all in `metric/metric_family.cc`. We do have a lot of ifdefs in other files too.

src/custom_metrics_registry.cc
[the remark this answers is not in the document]
krishung5: > Actually, I'm surprised that your test isn't throwing a PythonBackendException here when testing duplicate labels:

That was due to some error checking misplaced in the stub side for the duplicate labels, which only prints out error message. This should be fixed as the core handles the duplicate cases now.

src/custom_metrics_registry.cc
krishung5: Thanks for the clarification. Rearranged the files.

krishung5:
Addressed all the comments above and updated the description for the new changes.

examples/custom_metrics/model.py
```diff
@@ -0,0 +1,177 @@
...
+        Returns
+        -------
+        list
+          A list of pb_utils.InferenceResponse. The length of this list must
+          be the same as `requests`
+        """
+
+        output0_dtype = self.output0_dtype
+        output1_dtype = self.output1_dtype
+
```
rmccorm4, line 119: Not really related, but I think we should start getting rid of this copy/paste from PB examples and just reference `self.output0_dtype` and `self.output1_dtype` below directly. If it's a line length issue, we can just shorten the variable names a bit.

krishung5: Fixed.

examples/custom_metrics/config.pbtxt
```diff
@@ -0,0 +1,60 @@
...
+]
+output [
+  {
+    name: "OUTPUT1"
+    data_type: TYPE_FP32
+    dims: [ 4 ]
+  }
+]
+
+instance_group [{ kind: KIND_CPU }]
```
rmccorm4, line 59: I think I misunderstood when I first [commented ](https://github.com/triton-inference-server/server/pull/5669#discussion_r1174194214) on the testing PR. I think this is the model that you'd want to increase the instance count on as as sanity check, as it will interact with the metrics API on `initialize()` and track metrics/families across multiple instances referring to the same underlying objects, as the instances will inherently create duplicates. 

As far as I can tell, the model used in testing PR only interacts with the new APIs via `execute()` function, so it's just distributing the requests to the instances (if sending more than one request, I'm not sure at a glance if that unittest logic is sending multiple requests or not). I also believe the metric/family objects would be getting initialized and cleaned up for each unittest run (request).

src/metric.cc
```diff
@@ -0,0 +1,324 @@
...
+#endif
+
+namespace triton { namespace backend { namespace python {
+
+Metric::Metric(
+    const std::string& family_name, const std::string& labels,
+    void* metric_address)
+    : family_name_(family_name), labels_(labels), value_(0),
+      metric_request_kind_(MetricNew), metric_address_(nullptr),
+      metric_family_address_(metric_address)
```
rmccorm4, line 40: just to make the naming clear, should `metric_address` arg be `family_address` here?

src/metric_family.cc
```diff
@@ -0,0 +1,230 @@
...
+
+MetricFamily::~MetricFamily()
+{
+#ifdef TRITON_PB_STUB
+  // Send the request to delete the MetricFamily to the parent process
+  if (metric_family_request_kind_ != MetricFamilyDelete) {
+    // Clear all the metrics first
+    for (auto& m : metric_map_) {
+      m.second->Clear();
+    }
```
rmccorm4, line 68: does this need to be protected by `metric_map_mu_` ?

src/metric_family.cc
```diff
@@ -0,0 +1,230 @@
...
+      return TRITONSERVER_METRIC_KIND_GAUGE;
+    default:
+      throw PythonBackendException("Unknown metric kind");
+  }
+}
+
+void
+MetricFamily::ClearTritonMetricFamily()
+{
+  metric_map_.clear();
```
rmccorm4, line 219: does this need to be protected by `metric_map_mu_` ?

src/metric_family.h
```diff
@@ -0,0 +1,145 @@
...
+  /// \return Returns the metric kind of the custom metric family.
+  const MetricKind& Kind();
+
+  /// Get the shm handle of the custom metric family.
+  /// \return Returns the shm handle of the custom metric family.
+  bi::managed_external_buffer::handle_t ShmHandle();
+
+  /// Get the custom metric family request kind.
+  /// \return Returns the custom metric family request kind.
+  const MetricFamilyRequestKind& RequestKind();
```
rmccorm4, line 81: This `MetricFamilyRequestKind` is used to indicate the action PB should take when communicating over shared memory, right? My understanding is that's why its part of `MetricFamilyShm`.

What is its purpose as a member of `MetricFamily` class?

src/metric.h
```diff
@@ -0,0 +1,177 @@
...
+  /// \return Returns the labels of the metric.
+  const std::string& Labels();
+
+  /// Get the shared memory handle of the metric.
+  /// \return Returns the shared memory handle of the metric.
+  bi::managed_external_buffer::handle_t ShmHandle();
+
+  /// Get the request kind of the metric.
+  /// \return Returns the request kind of the metric.
+  const MetricRequestKind& RequestKind();
```
rmccorm4, line 82: I'm not sure I understand why the Metric/MetricFamily objects have a RequestKind(). Shouldn't the same long-lived metric/metric family object be created, incremented, set, deleted, etc.? What does it mean for the object to have a RequestKind? Isn't that a moreso a property of the ShmHandles passed around for communication?

src/metric_family.cc
```diff
@@ -0,0 +1,230 @@
...
+  }
+  metric_family_address_ = custom_metrics_msg->address;
+#endif
+}
+
+MetricFamily::~MetricFamily()
+{
+#ifdef TRITON_PB_STUB
+  // Send the request to delete the MetricFamily to the parent process
+  if (metric_family_request_kind_ != MetricFamilyDelete) {
```
rmccorm4, line 64: If the kind is already set to MetricFamilyDelete, that means the Delete request has already been sent? Is this handling because stubs can concurrently send delete requests? do multiple stubs access the same object concurrently?

src/metric_family.cc
```diff
@@ -0,0 +1,230 @@
...
+  metric_family_address_ = custom_metric_family_shm_ptr_->metric_family_address;
+}
+
+void*
+MetricFamily::MetricFamilyAddress()
+{
+  return metric_family_address_;
+}
+
+#ifdef TRITON_PB_STUB
```
rmccorm4, line 185: I could use a crash course on when methods are called by stub only vs. called by both or non-stub

examples/custom_metrics/model.py
```diff
@@ -0,0 +1,174 @@
...
+        # Update metric to track cumulative requests processing latency.
+        # There are three operations you can do with the Metric object:
+        #   - Metric.increment(value): Increment the value of the metric by
+        #       the given value. The type of the value is double. The 'COUNTER'
+        #       kind does not support negative value.
+        #   - Metric.set(value): Set the value of the metric to the given
+        #       value. This operation is only supported in 'GAUGE' kind. The
+        #       type of the value is double.
+        #   - Metric.value(): Get the current value of the metric.
+        self.metric.increment(end_ns - start_ns)
```
rmccorm4, line 162: It might be a good idea to use the Logger in this example as well to test any potential issues with metrics/logs sharing the same queue.

src/pb_utils.h
```diff
@@ -165,18 +167,28 @@ struct ResponseBatch : SendMessageBase {
...
   void* id;
 };
 
+struct CustomMetricsMessage : SendMessageBase {
+  bi::managed_external_buffer::handle_t message;
+  bool has_error;
+  bool is_error_set;
+  bi::managed_external_buffer::handle_t error;
+  double value;
+  void* address;
```
rmccorm4, line 189: comment what these are

src/metric.h
```diff
@@ -0,0 +1,177 @@
...
+
+enum MetricRequestKind {
+  MetricNew,
+  MetricDelete,
+  MetricValue,
+  MetricIncrement,
+  MetricSet
+};
+
+struct MetricShm {
```
rmccorm4, line 51: Can you document this struct with some comments? Some examples, overall purpose of the struct, when is `value` needed (only for certain RequestKind), etc., `metric_family_address` corresponds to the family that this metric belongs to, etc.

src/metric.h
```diff
@@ -0,0 +1,177 @@
...
+  /// Use Triton C API to increment the value of the metric by the given value.
+  /// \param value The value to increment the metric by.
+  void Increment(const double& value);
+
+  /// Use Triton C API to set the value of the metric to the given value.
+  /// \param value The value to set the metric to.
+  void SetValue(const double& value);
+
+  /// Use Triton C API to get the value of the metric.
+  void UpdateValue();
```
rmccorm4, line 148: I think this is also updating the internal value, and must be called by other operations to keep local value state correct?

src/metric.cc
```diff
@@ -0,0 +1,324 @@
...
+  std::unique_ptr<PbString> family_name_shm = PbString::LoadFromSharedMemory(
+      shm_pool, custom_metric_shm_ptr->family_name_shm_handle);
+  std::unique_ptr<PbString> labels_shm = PbString::LoadFromSharedMemory(
+      shm_pool, custom_metric_shm_ptr->labels_shm_handle);
+
+  return std::unique_ptr<Metric>(
+      new Metric(custom_metric_shm, family_name_shm, labels_shm));
+}
+
+Metric::Metric(
```
rmccorm4, line 140: Can you document the difference between the locally loaded metric object (first constructor) vs object created from shared memory handles (second constructor)?

It seems like this class is kind of like 2 classes in 1. First half of the file is local class, and second half is SHM class.

src/metric.cc
```diff
@@ -0,0 +1,324 @@
...
+}
+
+std::unique_ptr<Metric>
+Metric::LoadFromSharedMemory(
+    std::unique_ptr<SharedMemoryManager>& shm_pool,
+    bi::managed_external_buffer::handle_t handle)
+{
+  AllocatedSharedMemory<char> custom_metric_shm = shm_pool->Load<char>(handle);
+  MetricShm* custom_metric_shm_ptr =
+      reinterpret_cast<MetricShm*>(custom_metric_shm.data_.get());
```
Tabrizian, line 129: Why not load a `MetricShm` directly?

src/metric.cc
```diff
@@ -0,0 +1,324 @@
...
+}
+
+void
+Metric::SaveToSharedMemory(std::unique_ptr<SharedMemoryManager>& shm_pool)
+{
+  AllocatedSharedMemory<char> custom_metric_shm =
+      shm_pool->Construct<char>(sizeof(MetricShm), true);
+  custom_metric_shm_ptr_ =
+      reinterpret_cast<MetricShm*>(custom_metric_shm.data_.get());
+
```
Tabrizian, line 103: You can construct a `MetricShm` directly.

src/metric.cc
```diff
@@ -0,0 +1,324 @@
...
+{
+  return metric_address_;
+}
+
+#ifdef TRITON_PB_STUB
+void
+Metric::SendIncrementRequest(const double& value)
+{
+  std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
+  metric_request_kind_ = MetricIncrement;
```
Tabrizian, line 169: I think `metric_request_kind_` should not be a part of this object. It should be the command field of the `IPCMessage`.

src/pb_stub.cc
```diff
@@ -428,6 +427,13 @@ Stub::StubSetup()
   py::setattr(
       python_backend_utils, "TRITONSERVER_MEMORY_CPU",
       c_python_backend_utils.attr("TRITONSERVER_MEMORY_CPU"));
+  py::setattr(
+      python_backend_utils, "MetricFamily",
+      c_python_backend_utils.attr("MetricFamily"));
+  py::setattr(
+      python_backend_utils, "COUNTER", c_python_backend_utils.attr("COUNTER"));
```
Tabrizian, line 434: Can we make the COUNTER/GUAGE part of the `MetricFamily` object?

src/python_be.cc
```diff
@@ -886,6 +889,107 @@ ModelInstanceState::ProcessBLSCleanupRequest(
...
+    LOG_IF_EXCEPTION(
+        pb_error_message =
+            PbString::Create(Stub()->ShmPool(), pb_exception.what()));
+    metrics_message_ptr->error = pb_error_message->ShmHandle();
+    metrics_message_ptr->is_error_set = true;
+  }
+}
+
+void
+ModelInstanceState::ProcessMetricRequest(
```
Tabrizian, line 943: It seems like there is a lot in common with `ProcessMetricRequest` and `ProcessMetricFamilyRequest`. Would it be possible to refactor the common part into a separate function?

src/metric.h
```diff
@@ -0,0 +1,177 @@
...
+  /// location.
+  static std::unique_ptr<Metric> LoadFromSharedMemory(
+      std::unique_ptr<SharedMemoryManager>& shm_pool,
+      bi::managed_external_buffer::handle_t handle);
+
+  /// Get the address of the TRITONSERVER_Metric object.
+  /// \return Returns the address of the TRITONSERVER_Metric object.
+  void* MetricAddress();
+
+  /// Send the request to the parent processto delete the Metric object.
```
Tabrizian, line 105:
```suggestion
  /// Send the request to the parent process to delete the Metric object.
```

src/metric.cc
```diff
@@ -0,0 +1,324 @@
...
+  for (const auto& member : members) {
+    std::string value;
+    THROW_IF_TRITON_ERROR(labels_json.MemberAsString(member.c_str(), &value));
+    labels_params.emplace_back(TRITONSERVER_ParameterNew(
+        member.c_str(), TRITONSERVER_PARAMETER_STRING, value.c_str()));
+  }
+}
+
+void
+Metric::HandleMetricOperation(CustomMetricsMessage** metrics_message_ptr)
```
Tabrizian, line 276: I don't think you need double pointer here.

examples/custom_metrics/model.py
krishung5: Good idea. I think adding the Logger in the testing would be better. commit: https://github.com/triton-inference-server/server/pull/5669/commits/79a5e9adc7af3ef3fe714034220cccc0d9a14965.

src/metric.h
krishung5: Comments added.

src/metric.cc
krishung5: Added comments. Let me know if it's still unclear or confusing.

src/metric_family.cc
krishung5: Good catch! Fixed.

src/pb_stub.h
```diff
@@ -152,6 +155,17 @@ class LogMessage {
...
+// The payload for the stub_to_parent message queue
+struct UtilsMessagePayload {
+  UtilsMessagePayload(
+      const PYTHONSTUB_CommandType& command_type, void* utils_message_ptr)
+      : command_type(command_type), utils_message_ptr(utils_message_ptr)
+  {
+  }
+  PYTHONSTUB_CommandType command_type;
+  void* utils_message_ptr;
+};
```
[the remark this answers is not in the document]
krishung5: If the buffer accepts `IPCMessage` class, we will need to reconstruct the message struct (e.g. [LogSendMessage](https://github.com/triton-inference-server/python_backend/blob/main/src/pb_stub.cc#L944) and [CleanupMessage](https://github.com/triton-inference-server/python_backend/blob/main/src/pb_stub.cc#L978C9-L979)) from `ipc_message->Args()` to [set the message handle](https://github.com/triton-inference-server/python_backend/blob/main/src/pb_stub.cc#L980-L981). The challenge is that since the buffer runs in a separate thread, we will need to make sure that all the objects used to form the `IPCMessage` before pushed to the buffer are still alive until the message is sent to the parent process. I'm not sure if modifying the `IPCMessage` class to handle the lifetime of those objects would be a good idea, hence created a new message payload.

src/metric_family.cc
krishung5: This line is no longer needed as we already clean the map during the stub stage. We don't pass the `metric_map_` to the Python backend.

src/metric.cc
krishung5: Yes, fixed.

examples/custom_metrics/config.pbtxt
krishung5: Updated the instance count in the example.

src/metric.h
krishung5: Removed all the request kind and use `PYTHONSTUB_CommandType` in `IPCMessage` instead.

src/metric_family.cc
krishung5: Removed the check for `MetricFamily` as the destructor should only be called once.

src/pb_utils.h
krishung5: Comments added.

src/metric.h
krishung5: Yes, that was the purpose. However, I think we don't need to update the value after each operation as we don't directly retrieve the value from the `Metric` object. Instead, we get the value directly from the result of Triton C API call. I changed this function name to `GetValue()`, and the `value_` variable now is called `operation_value_` indicating that it's for the operation calls.

src/metric_family.cc
```diff
@@ -0,0 +1,199 @@
...
+  try {
+    stub->SendCustomMetricsMessage(
+        &custom_metrics_msg, PYTHONSTUB_MetricFamilyRequestNew, shm_handle_);
+  }
+  catch (const PythonBackendException& pb_exception) {
+    throw PythonBackendException(
+        "Error when creating MetricFamily: " +
+        std::string(pb_exception.what()));
+  }
+  metric_family_address_ = custom_metrics_msg->address;
```
Tabrizian, line 55: Can this be also extracted to a separate function?

src/metric.cc
```diff
@@ -0,0 +1,279 @@
...
+    SetValue(operation_value_);
+  } else {
+    throw PythonBackendException("Unknown metric operation");
+  }
+}
+
+void
+Metric::Increment(const double& value)
+{
+  auto triton_metric = reinterpret_cast<TRITONSERVER_Metric*>(metric_address_);
```
rmccorm4, line 248: Does it make sense to just store the `TRITONSERVER_Metric*`  as a member variable along side the metric_address after initialized?

src/metric.cc
```diff
@@ -0,0 +1,279 @@
...
+  return value;
+}
+
+void
+Metric::ClearTritonMetric()
+{
+  auto metric = reinterpret_cast<TRITONSERVER_Metric*>(metric_address_);
+  if (metric != nullptr) {
+    LOG_IF_ERROR(TRITONSERVER_MetricDelete(metric), "deleting metric");
+  }
```
rmccorm4, line 274: `triton_metric` for consistency with the others
```suggestion
  auto metric = reinterpret_cast<TRITONSERVER_Metric*>(metric_address_);
  if (triton_metric != nullptr) {
    LOG_IF_ERROR(TRITONSERVER_MetricDelete(triton_metric), "deleting metric");
  }
```

src/metric_family.cc
```diff
@@ -0,0 +1,199 @@
...
+#endif
+}
+
+MetricFamily::~MetricFamily()
+{
+#ifdef TRITON_PB_STUB
+  // Clear all the metrics first
+  std::lock_guard<std::mutex> lock(metric_map_mu_);
```
rmccorm4, line 63: Should this be inside the scope of this for loop block? or the scope of the whole function?

src/pb_stub.cc
```diff
@@ -879,8 +881,7 @@ Stub::TerminateStubToParentQueueMonitor()
   Logger::GetOrCreateInstance()->SetBackendLoggingActive(false);
   {
     std::lock_guard<std::mutex> guard{stub_to_parent_message_mu_};
-    log_request_buffer_.push(DUMMY_MESSAGE);
-    bls_response_cleanup_buffer_.push(DUMMY_MESSAGE);
+    stub_to_parent_buffer_.push(DUMMY_MESSAGE);
```
rmccorm4, line 884: Can you add a comment on the purpose of the dummy message?

src/metric_family.cc
krishung5: The lock should be inside the scope only. Fixed.

src/python_be.cc
```diff
@@ -886,6 +908,112 @@ ModelInstanceState::ProcessBLSCleanupRequest(
   }
 }
 
+template <typename T>
```
rmccorm4, line 911: Comment that T should be Metric or MetricFamily

src/metric.cc
```diff
@@ -0,0 +1,285 @@
...
+Metric::LoadFromSharedMemory(
+    std::unique_ptr<SharedMemoryManager>& shm_pool,
+    bi::managed_external_buffer::handle_t handle)
+{
+  AllocatedSharedMemory<MetricShm> custom_metric_shm =
+      shm_pool->Load<MetricShm>(handle);
+  MetricShm* custom_metric_shm_ptr = custom_metric_shm.data_.get();
+
+  std::unique_ptr<PbString> labels_shm = PbString::LoadFromSharedMemory(
+      shm_pool, custom_metric_shm_ptr->labels_shm_handle);
```
rmccorm4, line 81: Had a conversation with Kris offline about the performance impact of having to Save/Load the metric object on every single operation. The `labels` in particular may be costly to serialize/deserialize, especially as number of labels increases. She will run some local tests to get a rough idea of the impact. If there is a large impact, we had some potential ideas to follow-up with like communicating only the address as an identifier, or persistent shared memory for long lived objects like metrics.

Tabrizian (approved):

src/metric.cc
Tabrizian: Yeah, that can be costly. I think this is a follow up that can be addressed in the next release.

rmccorm4: Follow-up sounds good to me, we can get this out the door to start seeing it used in the wild.

rmccorm4 (approved):
Code looks very clean, great job!

src/metric.cc
krishung5: Created a follow-up ticket for this. Let me address that in the next release. Thanks again for all the reviews and discussion!

PR #226, opened 2023-03-29
[WIP] updated inferentia build

jbkyang-nvi:
Addresses [5520](https://github.com/triton-inference-server/server/issues/5520)
1. Update setup to use inf1 with `neuronx-dkms` and `aws-neuronx-tools`
2. Update build process to use python venv instead of conda using [neuron build instructions(https://awsdocs-neuron.readthedocs-hosted.com/en/latest/frameworks/torch/torch-neuron/setup/prev-releases/neuron-2.5.0-pytorch-install.html?highlight=aws-neuronx-dkms)

Todo:
1. Update tutorial to use latest Deeplearning machine (pytorch neuron 20.04 Ubuntu).
2. Figure if we can update to latest version of neuron sdk
3. Figure if we should change to inf2, or add inf2 build to our process
4. Run tests manually and make sure they work

PR #236, opened 2023-04-24
Update inferentia 1 setup

GuanLuo:
Branch off from #226, with changes below:

- removed model-server (TFS) package that is not needed for running inf1 with Triton.
- remove Python virtual environment setting as the application already runs in Docker container to provide isolation.

### DIFF
--- a/inferentia/scripts/setup.sh
+++ b/inferentia/scripts/setup.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
--- a/inferentia/scripts/setup.sh
+++ b/inferentia/scripts/setup.sh
@@ -107,18 +107,6 @@ if [ $USE_TENSORFLOW -eq 1 ]; then
     fi
 fi
 
-# Get latest conda required https://repo.anaconda.com/miniconda/
-cd ${INFRENTIA_PATH}
-export CONDA_PATH=${INFRENTIA_PATH}/miniconda
-rm -rf $CONDA_PATH
-wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-py37_4.10.3-Linux-x86_64.sh \
-         -O ${PWD}/miniconda.sh --no-check-certificate && \
-    /bin/bash ${PWD}/miniconda.sh -b -p ${CONDA_PATH} && \
-    rm ${PWD}/miniconda.sh && \
-    $(eval echo "${CONDA_PATH}/bin/conda clean -ya")
-export PATH=${CONDA_PATH}/bin:${PATH}
-conda info
-
 # Install python_backend_stub installing dependencies
 apt-get update && \
     apt-get install -y --no-install-recommends \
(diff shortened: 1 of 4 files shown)


src/resources/triton_python_backend_utils.py
```diff
@@ -164,8 +164,8 @@ def get_output_tensor_by_name(inference_response, name):
             return output_tensor
 
     return None
-
-
+    
+
```
Tabrizian, line 168: Remove whitespace

tanmayv25 (dismissed):

tanmayv25:
@mc-nv  with this fix, we should be unblocked to integrate AWS inf1 testing on our CI.

inferentia/scripts/setup.sh
```diff
@@ -151,25 +135,25 @@ make triton-python-backend-stub -j16
...
     fi
 fi
 
 if [ $USE_PYTORCH -eq 1 ]
 then
-    conda install torch-neuron torchvision -y
-    # Upgrade torch-neuron and install transformers
-    pip install --upgrade torch-neuron neuron-cc[tensorflow] "protobuf<4" torchvision "transformers==4.6.0"
+    # Install PyTorch Neuron
+    python -m pip install torch-neuron neuron-cc[tensorflow] "protobuf" torchvision
```
Tabrizian, line 156: Should we pin the version of these libraries and maybe update them when we have tested with new packages?

krishung5:
I think we probably need to update the [documentation](https://github.com/triton-inference-server/python_backend/tree/main/inferentia#inferentia-setup) as we won't be using miniconda in the setup.

dyastremsky (approved):

inferentia/scripts/setup.sh
GuanLuo: I will defer this to @jbkyang-nvi as she removed the version pin in the reference PR.

However, I don't see it to be a major issue, because the protobuf and other libraries are for Inferentia use (relying on Inferentia validates against latest dependencies). And I don't think it will interfere with our use of protobuf (we build statically and thus no symbol expose to/depend on external protobuf), unless the Python side boilerplate of the backend uses Python protobuf?

krishung5 (approved):

tanmayv25 (approved):

inferentia/scripts/setup.sh
```diff
@@ -151,25 +135,25 @@ make triton-python-backend-stub -j16
...
+python -m pip config set global.extra-index-url https://pip.repos.neuron.amazonaws.com
 if [ $USE_TENSORFLOW -eq 1 ]; then
-    conda install tensorflow-neuron pillow -y
-    # Update Neuron TensorBoard
-    pip install --upgrade tensorboard-plugin-neuron
     # Update Neuron TensorFlow
     if [ $TENSORFLOW_VERSION -eq 1 ]; then
-        pip install --upgrade tensorflow-neuron==1.15.5.* neuron-cc "protobuf<4"
+        # Install TensorFlow Neuron
+        python -m pip install tensorflow-neuron[cc]==1.15.5.* "protobuf"
```
jbkyang-nvi, line 146: Should we even support TF1 for neuron if Triton doesn't support TF1?

inferentia/scripts/setup.sh
jbkyang-nvi: I believe this is the recommended way of installing the dependencies according to the [neuron documentation](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/quick-start/torch-neuron.html). The backend doesn't use protobuf. I would prefer to not pin libraries unless we want to stick to a specific version of inferentia.

PR #238, opened 2023-04-27
Fix Stub initialization

Tabrizian:
The stub process has a background health thread that sets a variable to true every 300ms. With [this](https://github.com/triton-inference-server/python_backend/commit/1b6fa772c4390ea036a3d911634699656eff47b8) change, if the stub process doesn't create the thread in one second, it would fail to load the model. We probably saw this on jetson since it takes longer to create that thread.

### DIFF
--- a/src/stub_launcher.cc
+++ b/src/stub_launcher.cc
@@ -321,6 +321,18 @@ StubLauncher::Launch()
 
     stub_pid_ = pid;
 
+    // The stub process would send two messages to the parent process during the
+    // initialization.
+    // 1. When the stub process's health monitoring thread has started.
+    // 2. When the initialization is fully completed and the Python model is
+    // loaded.
+    //
+    // The reason it is broken into two steps is that creation of the health
+    // monitoring thread may take longer which can make the server process think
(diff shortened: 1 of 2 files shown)


krishung5 (approved):
Nice catch! Thanks for fixing this.

src/stub_launcher.cc
```diff
@@ -321,6 +321,18 @@ StubLauncher::Launch()
...
+    // initialization.
+    // 1. When the stub process's health monitoring thread has started.
+    // 2. When the initialization is fully completed and the Python model is
+    // loaded.
+    //
+    // The reason it is broken into two steps is that creation of the health
+    // monitoring thread may take longer which can make the server process think
+    // that the stub process is unhealthy and return early. Waiting until the
+    // health thread is spawn would make sure would prevent this issue.
+    parent_message_queue_->Pop();
```
nnshah1, line 334: For my understanding: this will wait indefinitely for the first message from the stub? This is to avoid trying to get the health check and failing because the background thread hasn't started yet?

Tabrizian: Exactly.

PR #239, opened 2023-04-28
Copyright update

kthui:


oandreeva-nv (approved):
Thanks for the quick follow up!

PR #240, opened 2023-04-29
Fix BLS timeout issue

krishung5:
We have encountered a timeout issue in the L0_backend_python/bls test, specifically when executing the test case [test_multiple_bls](https://github.com/triton-inference-server/server/blob/main/qa/python_models/bls/model.py#L581) with decoupled support.

The problem arises when the parent process sends the decoupled response to the stub process. At the end of the communication, the [condition variable](https://github.com/triton-inference-server/python_backend/blob/main/src/pb_stub.cc#L1119-L1120) in the stub process fails to wake up when the parent process calls [notify_all()](https://github.com/triton-inference-server/python_backend/blob/main/src/python_be.cc#L1518-L1525). (I'm still not sure about the root cause for this part.) Consequently, the stub process becomes stuck and cannot proceed to process the next decoupled response. As a result, the client side experiences a raised timeout exception. This issue occurs sporadically, approximately once in thousands of calls to the [function](https://github.com/triton-inference-server/server/blob/main/qa/python_models/bls/model.py#L72) used in the test_multiple_bls test case.

By removing the unnecessary extra round of communication between the two processes, I have conducted approximately 10,000 function calls without encountering any further timeout failures.

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -1116,9 +1116,6 @@ Stub::ParentToStubMQMonitor()
           *(ipc_message->ResponseMutex())};
       response_batch->waiting_on_stub = true;
       ipc_message->ResponseCondition()->notify_all();
-      while (response_batch->waiting_on_stub) {
-        ipc_message->ResponseCondition()->wait(lock);
-      }
     }
   }
 }
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -1515,15 +1515,6 @@ ModelInstanceState::SendBLSDecoupledResponse(
     }
   }
 
-  ScopedDefer _([&ipc_message, response_batch] {
-    {
-      bi::scoped_lock<bi::interprocess_mutex> lock{
-          *(ipc_message->ResponseMutex())};
-      response_batch->waiting_on_stub = false;
-      ipc_message->ResponseCondition()->notify_all();
-    }
-  });
-
   {
     bi::scoped_lock<bi::interprocess_mutex> lock{
         *(ipc_message->ResponseMutex())};


src/pb_stub.cc
```diff
@@ -1116,9 +1116,6 @@ Stub::ParentToStubMQMonitor()
           *(ipc_message->ResponseMutex())};
       response_batch->waiting_on_stub = true;
       ipc_message->ResponseCondition()->notify_all();
-      while (response_batch->waiting_on_stub) {
-        ipc_message->ResponseCondition()->wait(lock);
-      }
```
Tabrizian, line 1121: I think we need this call since the response batch could be deallocated from shared memory. Feel free to correct me if I'm wrong here.

krishung5: I was wondering if you mean that as soon as the stub process reaches this line: https://github.com/triton-inference-server/python_backend/blob/main/src/pb_stub.cc#L1118, the parent process will exit from the function and hence the response batch may be deallocated?
If that's the case, I think since the stub process already retrieves the content from the response batch, forms an infer response and [store it into the response_iterator](https://github.com/triton-inference-server/python_backend/blob/main/src/pb_stub.cc#L1099-L1112) before calling `notify_all()` to inform the parent process, the response batch will not go out of scope before we finish processing it in the stub process. Please let me know if I'm misunderstanding something from here.

Tabrizian (approved):

src/python_be.cc
```diff
@@ -1515,15 +1515,6 @@ ModelInstanceState::SendBLSDecoupledResponse(
     }
   }
 
-  ScopedDefer _([&ipc_message, response_batch] {
-    {
-      bi::scoped_lock<bi::interprocess_mutex> lock{
-          *(ipc_message->ResponseMutex())};
-      response_batch->waiting_on_stub = false;
```
rainyfly, line 1522: This line has changed waiting_on_stub to false first, then notify the pb_stub. But why pb_stub's waiting_on_stub is still true after being waken up. @krishung5

PR #243, opened 2023-05-13
adding proxy CUDA streams for dlpack synchronization

oandreeva-nv:
A follow up to the recent PR. I am making `cudaStreamSynchronize(0)` for all `from_dlpack` calls, since we need to be sure that all work is done on the external tensor and not count on external implementation of `__dlpack__` method.

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -830,16 +830,14 @@ Stub::Finalize()
 #ifdef TRITON_ENABLE_GPU
   // We also need to destroy created proxy CUDA streams for dlpack, if any
   std::lock_guard<std::mutex> lock(dlpack_proxy_stream_pool_mu_);
-  if (!dlpack_proxy_stream_pool_.empty()) {
-    for (auto& entry : dlpack_proxy_stream_pool_) {
-      // We don't need to switch device to destroy a stream
-      // https://stackoverflow.com/questions/64663943/how-to-destroy-a-stream-that-was-created-on-a-specific-device
-      cudaError_t err = cudaStreamDestroy(entry.second);
-      if (err != cudaSuccess) {
-        LOG_INFO
-            << "Failed to destroy dlpack CUDA proxy stream on device with id " +
-                   std::to_string(entry.first);
-      }
+  for (auto& entry : dlpack_proxy_stream_pool_) {
+    // We don't need to switch device to destroy a stream
+    // https://stackoverflow.com/questions/64663943/how-to-destroy-a-stream-that-was-created-on-a-specific-device
+    cudaError_t err = cudaStreamDestroy(entry.second);
+    if (err != cudaSuccess) {
+      LOG_ERROR
+          << "Failed to destroy dlpack CUDA proxy stream on device with id " +
+                 std::to_string(entry.first);
     }
   }
 #endif
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -59,6 +59,10 @@
 namespace py = pybind11;
 using namespace pybind11::literals;
 namespace bi = boost::interprocess;
+#ifndef TRITON_ENABLE_GPU
+using cudaStream_t = void*;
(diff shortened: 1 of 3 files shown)


src/pb_tensor.cc
```diff
@@ -391,6 +375,22 @@ PbTensor::FromDLPack(const std::string& name, const py::object& tensor)
...
 
+    // In case there is a pending job on the data, where this capsule
+    // is pointing to, we need to wait for it to finish before returning
+    // capsule.
+    // We synchronize on the default stream explicitly since that what we 
+    // pass to external tensor's `__dlpack__` method and in case when memory
+    // is allocated on GPU, PbTensor should be on the default stream.
+    // If external tensor's `__dlpack__` method for some reason does not make
+    // default stream wait for when the capsule is ready, this sync will help. 
+    err = cudaStreamSynchronize(0);
```
Tabrizian, line 386: I don't think this solves the issue with PyTorch. PyTorch doesn't enqueue any work on the default stream. It incorrectly creates another stream instead of default stream and enqueues work on that stream. To fix that problem, we need to create our own stream (or use per-thread default stream) and pass that via dlpack instead of legacy default stream. Feel free to correct me if I'm wrong here.

Tabrizian (changes requested):

src/pb_tensor.cc
oandreeva-nv: >PyTorch doesn't enqueue any work on the default stream.
What work you are mentioning? If all work was on default stream, then pytorch will schedule it on the default stream. Problem comes from `__dlpack__` implementation, which can't retrieve default stream at the moment. Then it is trying to make that stream wait. That was causing issues, since before this implementation wasn't issuing any work on GPU and counting on external's implementation to make default stream wait. Instead, no work was scheduled and we were risking to return garbadge data. Since any work scheduled on the default stream will also make all non-blocking streams to finish their work, we were counting on it. 

Now when we do `cudaStreamSynchronize(0)` on our side, it will make the device wait for all prior non-blocking streams to finish work, since this call was issued on the default stream and then move on. This also shouldn't affect any other's implementations, since this will be called after the default stream get the `event done` call.

> To fix that problem, we need to create our own stream (or use per-thread default stream) and pass that via dlpack instead of legacy default stream.

I think this is an overkill. PyTorch doesn't support PTDS and they are tricky to work with.

src/pb_stub.cc
```diff
@@ -823,6 +827,22 @@ Stub::Finalize()
...
+#ifdef TRITON_ENABLE_GPU
+  // We also need to destroy created proxy CUDA streams for dlpack, if any
+  std::lock_guard<std::mutex> lock(dlpack_proxy_stream_pool_mu_);
+  if (!dlpack_proxy_stream_pool_.empty()) {
+    for (auto& entry : dlpack_proxy_stream_pool_) {
+      // We don't need to switch device to destroy a stream
+      // https://stackoverflow.com/questions/64663943/how-to-destroy-a-stream-that-was-created-on-a-specific-device
+      cudaError_t err = cudaStreamDestroy(entry.second);
+      if (err != cudaSuccess) {
+        LOG_INFO
```
Tabrizian, line 839: LOG_ERROR?

src/pb_stub.cc
```diff
@@ -823,6 +827,22 @@ Stub::Finalize()
       LOG_INFO << e.what();
     }
   }
+#ifdef TRITON_ENABLE_GPU
+  // We also need to destroy created proxy CUDA streams for dlpack, if any
+  std::lock_guard<std::mutex> lock(dlpack_proxy_stream_pool_mu_);
+  if (!dlpack_proxy_stream_pool_.empty()) {
```
Tabrizian, line 833: small nit: I think this check is not needed.

Tabrizian (approved):

Tabrizian (approved):

PR #245, opened 2023-05-19
Fix cuda synchronization

Tabrizian:
Closes https://github.com/triton-inference-server/server/issues/5733

After:
```
Number of threads: 3, number of inferences: 1000
Number of failures: [0, 0, 0]
Total: 0
```

Before:
```
Number of threads: 3, number of inferences: 1000
Number of failures: [396, 405, 402]
Total: 1203
```

### DIFF
--- a/src/pb_memory.cc
+++ b/src/pb_memory.cc
@@ -165,10 +165,12 @@ PbMemory::CopyBuffer(
     err = cudaStreamSynchronize(0);
     if (err != cudaSuccess) {
       throw PythonBackendException(
-          std::string("failed to synchronize the default CUDA stream").c_str());
+          std::string(
+              "failed to synchronize the default CUDA stream. error: " +
+              std::string(cudaGetErrorString(err)))
+              .c_str());
     }
   }
-
 #endif
 }
 
--- a/src/pb_memory.cc
+++ b/src/pb_memory.cc
@@ -142,8 +142,7 @@ PbMemory::CopyBuffer(
   }
 
   cudaError_t err;
-  if ((src->MemoryType() == TRITONSERVER_MEMORY_GPU &&
-       dst->MemoryType() == TRITONSERVER_MEMORY_GPU) &&
+  if ((kind == cudaMemcpyDeviceToDevice) &&
       (src->MemoryTypeId() != dst->MemoryTypeId())) {
     err = cudaMemcpyPeer(
         dst->DataPtr(), dst->MemoryTypeId(), src->DataPtr(),
@@ -160,8 +159,7 @@ PbMemory::CopyBuffer(
             .c_str());
   }
 
-  if ((src->MemoryType() == TRITONSERVER_MEMORY_GPU &&
-       dst->MemoryType() == TRITONSERVER_MEMORY_GPU)) {
+  if (kind == cudaMemcpyDeviceToDevice) {
     // Synchronize the default stream for d2d copies.
     // https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memcpy-sync
     err = cudaStreamSynchronize(0);
--- a/src/pb_memory.cc
+++ b/src/pb_memory.cc
@@ -141,8 +141,17 @@ PbMemory::CopyBuffer(
     kind = cudaMemcpyDeviceToDevice;
   }
 
-  cudaError_t err =
-      cudaMemcpy(dst->DataPtr(), src->DataPtr(), src->ByteSize(), kind);
+  cudaError_t err;
+  if ((src->MemoryType() == TRITONSERVER_MEMORY_GPU &&
+       dst->MemoryType() == TRITONSERVER_MEMORY_GPU) &&
+      (src->MemoryTypeId() != dst->MemoryTypeId())) {
+    err = cudaMemcpyPeer(
+        dst->DataPtr(), dst->MemoryTypeId(), src->DataPtr(),
+        src->MemoryTypeId(), src->ByteSize());
+
+  } else {
+    err = cudaMemcpy(dst->DataPtr(), src->DataPtr(), src->ByteSize(), kind);
+  }
 
   if (err != cudaSuccess) {
     throw PythonBackendException(
@@ -150,6 +159,18 @@ PbMemory::CopyBuffer(
             "failed to copy data: " + std::string(cudaGetErrorString(err)))
             .c_str());
   }
+
+  if ((src->MemoryType() == TRITONSERVER_MEMORY_GPU &&
+       dst->MemoryType() == TRITONSERVER_MEMORY_GPU)) {
+    // Synchronize the default stream for d2d copies.
+    // https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memcpy-sync
+    err = cudaStreamSynchronize(0);
+    if (err != cudaSuccess) {
+      throw PythonBackendException(
+          std::string("failed to synchronize the default CUDA stream").c_str());
+    }
+  }
+
 #endif
 }
 


src/pb_memory.cc
```diff
@@ -141,15 +141,36 @@ PbMemory::CopyBuffer(
     kind = cudaMemcpyDeviceToDevice;
   }
 
-  cudaError_t err =
-      cudaMemcpy(dst->DataPtr(), src->DataPtr(), src->ByteSize(), kind);
+  cudaError_t err;
+  if ((src->MemoryType() == TRITONSERVER_MEMORY_GPU &&
+       dst->MemoryType() == TRITONSERVER_MEMORY_GPU) &&
+      (src->MemoryTypeId() != dst->MemoryTypeId())) {
+    err = cudaMemcpyPeer(
```
dzier, line 148: Is this if condition necessary? Do you have a test that triggers this if condition?

rmccorm4:
Do you think this is related to the multi-gpu L0_io failures from a while back as well?

src/pb_memory.cc
Tabrizian: Yes, to both questions. `cudaMemcpyPeer` needs to used when UVA is not enabled between different GPUs.

https://stackoverflow.com/questions/22694518/what-is-the-difference-between-cudamemcpy-and-cudamemcpypeer-for-p2p-copy

I think [L0_io](https://github.com/triton-inference-server/server/tree/main/qa/L0_io) test would trigger this condition. I'll double check to confirm.

Tabrizian:
Not sure, the L0_io failures where on multi-GPU systems with peer access disabled. This failure has been observed on memory copies on the same GPU.

src/pb_memory.cc
```diff
@@ -141,15 +141,36 @@ PbMemory::CopyBuffer(
     kind = cudaMemcpyDeviceToDevice;
   }
 
-  cudaError_t err =
-      cudaMemcpy(dst->DataPtr(), src->DataPtr(), src->ByteSize(), kind);
+  cudaError_t err;
+  if ((src->MemoryType() == TRITONSERVER_MEMORY_GPU &&
```
nnshah1, line 145: can we reuse the kind variable above and make this if kind ==cudaMemcpyDeviceToDevice

src/pb_memory.cc
```diff
@@ -141,15 +141,36 @@ PbMemory::CopyBuffer(
...
         std::string(
             "failed to copy data: " + std::string(cudaGetErrorString(err)))
             .c_str());
   }
+
+  if ((src->MemoryType() == TRITONSERVER_MEMORY_GPU &&
+       dst->MemoryType() == TRITONSERVER_MEMORY_GPU)) {
+    // Synchronize the default stream for d2d copies.
+    // https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memcpy-sync
+    err = cudaStreamSynchronize(0);
```
nnshah1, line 167: Do we need to explicitly use the legacy stream here? CU_STREAM_LEGACY

src/pb_memory.cc
```diff
@@ -141,15 +141,36 @@ PbMemory::CopyBuffer(
...
+  }
 
   if (err != cudaSuccess) {
     throw PythonBackendException(
         std::string(
             "failed to copy data: " + std::string(cudaGetErrorString(err)))
             .c_str());
   }
+
+  if ((src->MemoryType() == TRITONSERVER_MEMORY_GPU &&
```
nnshah1, line 163: same as above can we use kind variable here to indicate device to device

src/pb_memory.cc
oandreeva-nv: I don't think that it is necessary, since default stream has id 0 and it will behave either like Legacy Default (or PTDS, which we don't use). I think it is better to keep 0. We can get rid of the magic number though, by specifying CUDA_DEFAULT_STREAM for our code.

src/pb_memory.cc
oandreeva-nv: +1 on this

oandreeva-nv:
I see, so an issue was in D2D cudaMemCopy requiring a sync. Interesting, great find! Left one question to discuss.

oandreeva-nv (commented):
I see, so an issue was in D2D cudaMemCopy requiring a sync. Interesting, great find! Left one question to discuss.

GuanLuo (commented):
Have you considered using the [CopyBuffer](https://github.com/triton-inference-server/backend/blob/main/include/triton/backend/backend_common.h#L478-L503) helper from backend utils? It should already cover all the handling here.

src/pb_memory.cc
Tabrizian: > Do we need to explicitly use the legacy stream here? CU_STREAM_LEGACY

It looks like this maps to 1. I double checked. synchronization here would synchronize with all the blocking streams which is the expected behavior. It doesn't synchronize the non-blocking streams.

> Do you think that this sync here may impact user's code? Should we consider async D2D copy on a separate stream and sync on that stream and not on the default stream?

We can do this in a follow up PR. Once you're PR is merged in we can use those streams instead of default stream.

src/pb_memory.cc
Tabrizian: fixed.

src/pb_memory.cc
Tabrizian: fixed.

Tabrizian:
> Have you considered using the [CopyBuffer](https://github.com/triton-inference-server/backend/blob/main/include/triton/backend/backend_common.h#L478-L503) helper from backend utils? It should already cover all the handling here.

Yes, but the problem is we are not linking against the backend libraries in the stub process and as a result we can't use those utilities.

src/pb_memory.cc
nnshah1: I guess my question comes back to - do we need this to be the legacy stream or is it ok to be a per thread default stream? If I understand correctly - we just need to ensure the copy completes before the host moves on. Is the behavior that cudamemcpy / peermemcpy happens on default stream (either legacy or per thread) - and synchronize(0) will ensure the copy happens before the host moves on?

src/pb_memory.cc
```diff
@@ -141,15 +141,34 @@ PbMemory::CopyBuffer(
...
             .c_str());
   }
+
+  if (kind == cudaMemcpyDeviceToDevice) {
+    // Synchronize the default stream for d2d copies.
+    // https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memcpy-sync
+    err = cudaStreamSynchronize(0);
+    if (err != cudaSuccess) {
+      throw PythonBackendException(
+          std::string("failed to synchronize the default CUDA stream").c_str());
```
krishung5, line 168: I think it would be helpful for debugging if we add `cudaGetErrorString` to return more detailed error description in the exception.

src/pb_memory.cc
oandreeva-nv: Default stream behaves either like Legacy Default Stream or Per-Thread Default Stream. By default it behaves like Legacy Default. To change this behavior, one needs to compile cuda code with the --default-stream flag. See [stream sync behavior docs](https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html#stream-sync-behavior). 


> The behavior that cudamemcpy / peermemcpy happens on default stream
It is true, their async counterparts allow specification of a stream to avoid default stream. 

> and synchronize(0) will ensure the copy happens before the host moves on
This is true, but explicit sync needed only for d2d copy, because according to the docs: https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memcpy-sync 
in this case no host-side sync is performed. 
@Tabrizian feel free to correct me

src/pb_memory.cc
Tabrizian: fixed.

krishung5 (approved):

src/pb_memory.cc
nnshah1: Thanks!

nnshah1 (approved):

rmccorm4 (approved):

oandreeva-nv (approved):

PR #248, opened 2023-05-23
Improve the documentation for BLS

krishung5:


rmccorm4:
Can you add an explicit note after like "it is not supported in the initialize or finalize methods" or something like that?

rmccorm4 (dismissed):

README.md
```diff
@@ -877,8 +877,10 @@ Scripting (BLS)*.
 
 Starting from 21.08, you can implement BLS in your Python model. A new set of
 utility functions allows you to execute inference requests on other models
-being served by Triton as a part of executing your Python model. Example below
-shows how to use this feature:
+being served by Triton as a part of executing your Python model. Note that BLS
+should only be implemented inside the `execute` function and is not supported
```
Tabrizian, line 881:
```suggestion
should only be used inside the `execute` function and is not supported
```

rmccorm4 (approved):

Tabrizian (approved):

PR #249, opened 2023-05-25
Fix error handling for GPU tensors

Tabrizian:


rmccorm4:
Can you summarize the bug and fix in the description?

src/pb_stub.cc
```diff
@@ -539,43 +540,50 @@ Stub::ProcessResponse(InferResponse* response)
...
-  bi::managed_external_buffer::handle_t* gpu_buffers_handle_shm =
-      reinterpret_cast<bi::managed_external_buffer::handle_t*>(
-          gpu_buffers_handle.data_.get() + sizeof(uint64_t));
+  AllocatedSharedMemory<GPUBuffersShm> gpu_buffers_handle =
+      shm_pool_->Load<GPUBuffersShm>(ipc_message->Args());
+
+  if (!gpu_buffers_handle.data_->success) {
+    std::unique_ptr<PbString> error = PbString::LoadFromSharedMemory(
+        shm_pool_, gpu_buffers_handle.data_->error);
+    LOG_ERROR << ("Failed to load GPU buffers: " + error->String());
```
krishung5, line 556: I was wondering if it would be better to throw an exception here and catch it outside at line 362 instead of logging the error?

Tabrizian: Agreed! Will make this change.

src/infer_request.cc
```diff
@@ -481,12 +482,20 @@ InferRequest::Exec(const bool is_decoupled)
...
               shm_pool->Load<bi::managed_external_buffer::handle_t>(
-                  request_batch_shm_ptr->gpu_buffers_handle);
+                  gpu_buffers_shm.data_->buffers);
       try {
 #ifdef TRITON_ENABLE_GPU
+        if (!gpu_buffers_shm.data_->success) {
+          std::unique_ptr<PbString> error = PbString::LoadFromSharedMemory(
+              shm_pool, gpu_buffers_shm.data_->error);
+          throw PythonBackendException(error->String());
+        }
```
rmccorm4, line 498: Can the scope of the `#ifdef TRITON_ENABLE_GPU` be moved after this `if()` block to make the scope as small as necessary? Looks like everything in this block might be defined/used regardless of the ifdef - including `gpu_buffers_shm`, correct me if I'm wrong.

Though at a glance, I do find it a bit strange that some of the stuff like `has_gpu_tensor` and `gpu_buffers_shm` happen outside of the ifdef today.

rmccorm4: Also, is the `else` case of the ifdef just a no-op? Looks like the whole try block is inside the ifdef, so I'm a bit confused by the logic when `TRITON_ENABLE_GPU=OFF`.

src/python_be.cc
```diff
@@ -661,46 +660,33 @@ ModelInstanceState::ExecuteBLSRequest(
...
+            gpu_buffer_transporter.ShmHandle();
 
         bi::scoped_lock<bi::interprocess_mutex> lock{
             *(ipc_message->ResponseMutex())};
         ipc_message->ResponseCondition()->notify_all();
         ipc_message->ResponseCondition()->wait(lock);
       }
 
-      if (pb_exception.what() != nullptr) {
+      if (pb_exception.what() == std::string{""}) {
```
rmccorm4, line 689: Was this the bug? Is the logic supposed to be flipped here?

If logic is correct, can you use `.empty()` instead?

src/response_sender.cc
```diff
@@ -130,28 +130,30 @@ ResponseSender::Send(
...
+        shm_pool_->Load<GPUBuffersShm>(
+            send_message_payload->gpu_buffers_handle);
+
+    AllocatedSharedMemory<bi::managed_external_buffer::handle_t>
+        gpu_buffers_handle_shm =
+            shm_pool_->Load<bi::managed_external_buffer::handle_t>(
+                gpu_buffers_handle.data_->buffers);
+    uint64_t gpu_buffer_count = gpu_buffers_handle.data_->buffer_count;
+    if (gpu_tensors.size() != gpu_buffer_count) {
       LOG_INFO
```
rmccorm4, line 143: Should this be LOG_ERROR or throw instead of LOG_INFO?

src/python_be.cc
Tabrizian: `.what()` returns a `const char *` so it is not possible to use `empty`.

src/infer_request.cc
Tabrizian: > Can the scope of the #ifdef TRITON_ENABLE_GPU be moved after this if() block to make the scope as small as necessary? Looks like everything in this block might be defined/used regardless of the ifdef - including gpu_buffers_shm, correct me if I'm wrong.

That's correct. Fixed.

>  I do find it a bit strange that some of the stuff like has_gpu_tensor and gpu_buffers_shm happen outside of the ifdef today.

They can be moved to ifdef but they are harmless since they don't use CUDA APIs. `has_gpu_tensor` will be simply false.

> Also, is the else case of the ifdef just a no-op?

Yes, since this special handling is only required for GPU tensors, if `#TRITON_ENABLE_GPU` is off it means that we cannot handle GPU tensors and there is no special handling required.

src/gpu_buffers.h
```diff
@@ -0,0 +1,64 @@
...
+#include "pb_utils.h"
+#include "scoped_defer.h"
+
+namespace triton { namespace backend { namespace python {
+
+/// \param success indicating whether the request was successful
+/// \param error if success is equal to false, the error object will be set.
+/// \param buffers list of buffers elements.
+/// \param buffer_count the number of buffers.
+struct GPUBuffersShm {
```
nnshah1, line 39: Question: Do we need the Shm suffix? You mention 'request', is this GPUResponseBuffers or GPUMemoryResponse? 

if we don't need the Shm suffix - would think about removing it.


Also - is it used outside the GPUBufferTransporter class? If not - it could be a private struct within the class.

src/gpu_buffers.cc
```diff
@@ -0,0 +1,87 @@
...
+GPUBufferTransporter::SetError(
+    std::unique_ptr<SharedMemoryManager>& shm_pool, const std::string& error)
+{
+  error_shm_ = PbString::Create(shm_pool, error);
+}
+
+void
+GPUBufferTransporter::Complete(std::unique_ptr<SharedMemoryManager>& shm_pool)
+{
+  if (completed_) {
```
nnshah1, line 59: Should this be an error case as above with adding a buffer to a completed transaction?

src/pb_stub.cc
```diff
@@ -539,43 +540,48 @@ Stub::ProcessResponse(InferResponse* response)
...
-          gpu_buffers_handle.data_.get() + sizeof(uint64_t));
-
-  if (gpu_tensors_.size() != *gpu_buffer_count) {
-    LOG_INFO
-        << (std::string(
-                "GPU buffers size does not match the provided buffers: ") +
-            std::to_string(gpu_tensors_.size()) +
-            " != " + std::to_string(*gpu_buffer_count));
-    return;
+  AllocatedSharedMemory<GPUBuffersShm> gpu_buffers_handle =
```
nnshah1, line 550: could we use GPUBuffers as the structure name (leave off the Shm suffix)?

src/pb_utils.h
```diff
@@ -212,23 +212,17 @@ struct ResponseSenderBase {
 struct ResponseSendMessage : ResponseSenderBase {
   bi::managed_external_buffer::handle_t response;
 
-  // GPU Buffers handle
+  // A pointer to GPUBuffersShm object.
```
nnshah1, line 215: seems like we could keep this the same if the structure was renamed to GPUBuffers - is it still a handle or is that seperate from pointer?

src/python_be.cc
```diff
@@ -661,46 +660,33 @@ ModelInstanceState::ExecuteBLSRequest(
             lbackend_memory.reset(backend_memory);
             input_tensor->SetMemory(std::move(PbMemory::Create(
                 Stub()->ShmPool(), std::move(lbackend_memory))));
+            gpu_buffer_transporter.AddBuffer(
```
nnshah1, line 663: instead of `transporter` would `response` work? `GPUBuffers` and `GPUBuffersResponse` instead of `GPUBuffersShm` and `GPUBufferTransporter`

src/python_be.cc
```diff
@@ -661,46 +660,33 @@ ModelInstanceState::ExecuteBLSRequest(
...
-            if (!input_tensor->IsCPU()) {
-              gpu_handles.data_.get()[i] = input_tensor->Memory()->ShmHandle();
-              ++i;
-            }
-          }
-        }
-        catch (const PythonBackendException& exception) {
-          pb_exception = exception;
-        }
+        gpu_buffer_transporter.Complete(Stub()->ShmPool());
```
nnshah1, line 679: we pass in the shmpool to complete and seterror - could this be added to the constructor and stored ? Does the pool change for the same transporter object?

nnshah1:
I don't quite follow the logic of when we use the GPUBufferTransporter and when we use the GPUBuffersShm structure directly - but I think that's from my lack of familiarity with the code.

src/gpu_buffers.h
Tabrizian: The convention has been to have a `Shm` suffix for the structs that will be stored in shared memory. I don't have a strong opinion on the benefits of keeping/removing the `Shm` suffix but for consistency I prefer to keep it for now.

The "request" is more like requesting some buffers from the server. I'll clarify the comment.

It is used outside `GPUBufferTransporter` in pb_stub.cc.

src/pb_stub.cc
Tabrizian: Responded here: https://github.com/triton-inference-server/python_backend/pull/249#discussion_r1214774976

src/python_be.cc
Tabrizian: I like transporter more as I think "response" may imply that it can only be used with the backend responses which is not true. Let me know if you have any other suggestions.

src/pb_utils.h
Tabrizian: By pointer I meant a handle to a GPUBuffersShm. Will update the comment.

src/python_be.cc
Tabrizian: ShmPool is a reference to a unique pointer. I think the lifetime of a reference to a unique pointer stored as a class attribute is a bit weird as the object can continue to use the attribute after it is lifetime is finished which would result in segfault. I think this is more clear in terms of lifetime. Alternatively, we could return a shared pointer but that is slightly less performant than what we have here (it would have to obtain a lock, increase/decrease ref count for shared pointer).

src/gpu_buffers.h
Tabrizian: clarified comment.

src/gpu_buffers.h
```diff
@@ -0,0 +1,64 @@
...
+  bool success;
+  bi::managed_external_buffer::handle_t error;
+  bi::managed_external_buffer::handle_t buffers;
+  uint32_t buffer_count;
+};
+
+class GPUBufferTransporter {
```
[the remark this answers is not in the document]
Tabrizian: Added comments.

src/python_be.cc
krishung5: I agree that we should be careful about managing the lifetime. Passing ShmPool instead of storing it looks clearer to me as well. It seems like we do store a reference to a unique pointer in some of the objects though, [ResponseSender](https://github.com/triton-inference-server/python_backend/blob/a156eba0cc03739a5bdcf59da545d00c531e4ee3/src/response_sender.h#L44) for example.

nnshah1 (dismissed):

krishung5 (dismissed):
LGTM!

rmccorm4 (approved):

PR #251, opened 2023-06-01
Remove the requirement to specify output_tensors when setting an error

Tabrizian:
response

Server: https://github.com/triton-inference-server/server/pull/5892

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,7 +469,9 @@ class TritonPythonModel:
 
         for request in requests:
             if an_error_occurred:
-              # If there is an error, the output_tensors are ignored
+              # If there is an error, there is no need to pass the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -468,7 +471,6 @@ class TritonPythonModel:
             if an_error_occurred:
               # If there is an error, the output_tensors are ignored
```
krishung5, line 472: Would it be clearer if we update the comment here to something like: `If there is an error, there is no need to pass the output_tensor to the InferenceResponse. The output_tensors that are passed in this case will be ignored.`?

krishung5 (approved):

README.md
Tabrizian: Agreed, updated.

krishung5 (approved):

PR #254, opened 2023-06-07
Add error checking for the case where MetricFamily is deleted before …

krishung5:
…deleting Metric. Update documentation.

When the `MetricFamily` is deleted before the corresponding `Metric` objects are deleted, although the `TRITONSERVER_MetricFamily` and `TRITONSERVER_Metric` will be cleaned up properly in the Python backend side, users will still be able to use the `Metric` object for metric operation in the model. This PR added a check and throw an exception when having this kind of user error.

Added testing: https://github.com/triton-inference-server/server/pull/5915
Resolves: https://github.com/triton-inference-server/server/issues/5901

### DIFF
--- a/README.md
+++ b/README.md
@@ -1326,7 +1326,10 @@ Starting from 23.05, you can utlize Custom Metrics API to register and collect
 custom metrics in the `initialize`, `execute`, and `finalize` functions of your
 Python model. The Custom Metrics API is the Python equivalent of the
 [TRITON C API custom metrics](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/metrics.md#custom-metrics)
-support.
+support. You will need to take the ownership of the custom metrics created
+through the APIs and must manage their lifetime. Note that a `MetricFamily`
+object should be deleted only after all the `Metric` objects under it are
+deleted if you'd like to explicitly delete the custom metrics objects.
 
 Example below shows how to use this feature:
 
--- a/src/metric.cc
+++ b/src/metric.cc
@@ -124,11 +124,12 @@ Metric::SendCreateMetricRequest()
 void
 Metric::SendIncrementRequest(const double& value)
 {
-  std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
-  operation_value_ = value;
-  SaveToSharedMemory(stub->ShmPool());
-  CustomMetricsMessage* custom_metrics_msg = nullptr;
   try {
+    CheckIfCleared();
+    std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
+    operation_value_ = value;
+    SaveToSharedMemory(stub->ShmPool());
+    CustomMetricsMessage* custom_metrics_msg = nullptr;
     stub->SendCustomMetricsMessage(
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1326,7 +1326,10 @@ Starting from 23.05, you can utlize Custom Metrics API to register and collect
 custom metrics in the `initialize`, `execute`, and `finalize` functions of your
 Python model. The Custom Metrics API is the Python equivalent of the
 [TRITON C API custom metrics](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/metrics.md#custom-metrics)
-support.
+support. You will need to take the ownership of the custom metrics created
+through the APIs and must manage their lifetime. Note that a `MetricFamily`
```
rmccorm4, line 1330: I didn't look too deeply at the code yet, but a high level if `MetricFamily` is being maintained as a `shared_ptr` or equivalent in PB, would it make sense to hold a reference to it in the `Metric` object to basically remove the possibility of `MetricFamily` being deleted before its metrics on the PB side of things? Let me know if that makes sense. 

This is just a passing idea without thinking too much about it, so it might not.

krishung5: I think it's a good idea, although users who wish to remove a custom metric from the server metrics will need to clean up not only the `MetricFamily` but also all associated `Metric` objects to remove all the references of `MetricFamily`.(Right now users can just delete the `MetricFamily` and the custom metric will be removed from the server metrics)
However, I think it would be beneficial to provide an easier way to manage the lifetime of these objects. We can consider documenting that in order to remove a custom metric from the server, users should delete all related objects.

rmccorm4: > although users who wish to remove a custom metric from the server metrics will need to clean up not only the MetricFamily but also all associated Metric objects to remove all the references of MetricFamily.

I think this is an acceptable tradeoff to improve the usability of it, i.e. not having to worry about returning both the metric_family and metric objects from a helper function if you only need the metric object around for your processing as happened in this [issue](https://github.com/triton-inference-server/server/issues/5901). 

I think users that need the metric_family object for other things (create more metrics, etc.) would inherently keep it around. 

I think the average use case would entail keep the metrics around and (optionally) cleaning up the objects on finalize or something - but I would expect most users to not even manually call `del metric/family` and just let them fall out of scope like in most python scripts.

I think it's fine to document the behavior for now and we can think on this a bit based on more feedback out in the wild. CC @Tabrizian if you have any thoughts.

rmccorm4 (approved):

README.md
Tabrizian: >  would it make sense to hold a reference to the family inside the Metric object on creation

I like this idea and I think it is more intuitive. On the other hand, I wonder how the users would handle metrics that may be used across the models. It seems like in the current APIs, the metric would be deleted on unloading of the model so it wouldn't be possible to keep a metric alive beyond the scope of a given model lifecycle.

krishung5: @rmccorm4 please correct me if I'm wrong. I believe that if two models both have a reference to the same metrics, and one of the models is unloaded, it should not affect the reference of the other model. This behavior should be consistent for both Python custom metrics and the Triton Metrics C API."

PR #255, opened 2023-06-09
Interfacing with platform models, support TF model serving

tanmayv25:
-  [x] Interface the plugin models with python backend.
-  [x] Implement the tensorflow model serving in model.py([src/resources/plugin_models/tensorflow_savedmodel/model.py](https://github.com/triton-inference-server/python_backend/compare/tanmayv-tf?expand=1#diff-85a107719d69d01d40916e530528ee789d51bf8085a30cab75d74988167cfc02))
- [x] Ship TF package and dependency (ongoing) - **Decided to let users install dependencies by themselves**


src/pb_stub.cc
```diff
@@ -459,6 +437,12 @@ Stub::AutoCompleteModelConfig(
   if (py::hasattr(sys.attr("TritonPythonModel"), "auto_complete_config")) {
     model_config = sys.attr("TritonPythonModel")
                        .attr("auto_complete_config")(model_config);
+  } else if (py::hasattr(
+                 sys.attr("TritonPythonModel"),
+                 "plugin_auto_complete_config")) {
```
Tabrizian, line 442: I don't think we need a new API for these models. Let's just call the regular Python model APIs.

Tabrizian (commented):
I don't think we need to have a separate plugin interface. For now, let's

src/pb_stub.cc
tanmayv25: The new APIs were added to additionally provide the complete path to the framework model. In this case the model.saved_model file living within model-repository.
Modifying the current function signature would not be backwards-compatible and all the current python models might start failings.
I think we can get by using a single initialize function as it consumes generic args, but for auto-complete-config, I don't know how we can retrieve the model file path.

src/resources/plugin_models/tensorflow_savedmodel/model.py
```diff
@@ -0,0 +1,457 @@
...
+        if 'serve' in tag_sets[0]:
+            graph_tag = 'serve'
+        else:
+            graph_tag = tag_sets[0][0]
+
+    meta_graph_def = saved_model_utils.get_meta_graph_def(
+                      savedmodel_path, graph_tag)
+    signature_def_map = meta_graph_def.signature_def
+    signature_def_k = _parse_signature_def(config)
+    if (signature_def_k is None):
```
Tabrizian, line 117: nit: remove parenthesis in the `if` conditions here and everywhere

src/python_be.cc
```diff
@@ -1759,6 +1767,26 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
...
+    return "model.savedmodel";
+  } else {
+    return "model.py";
+  }
+}
+
+std::string
+ModelState::PluginModel()
+{
+  if (platform_ == "tensorflow_savedmodel") {
```
nnshah1, line 1783: Is there a way that this could be managed by the platform_model.py and not the python_be?

Ideally the python_be wouldn't need to know any details of how the platform_model_loader works

src/resources/plugin_models/tensorflow_savedmodel/model.py
```diff
@@ -0,0 +1,457 @@
...
+                        batching_enabled))
+
+        if batching_enabled:
+            if (config['max_batch_size'] == 0):
+                auto_complete_model_config.set_max_batch_size(4)
+            auto_complete_model_config.set_dynamic_batching()
+
+        return auto_complete_model_config
+
+    def initialize(self, args, fw_model_path):
```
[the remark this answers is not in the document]
nnshah1: I think even the default artifact name - would be known to the platform_model class?
If needed we could add an additional arg with model_config_path? class can infer from model_config path the model itself

src/python_be.cc
```diff
@@ -1759,6 +1767,26 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
   }
 }
 
+std::string
+ModelState::DefaultArtifactName()
```
nnshah1, line 1771: Is there a way that this could be managed by the platform_model.py and not the python_be?

Ideally the python_be wouldn't need to know any details of how the platform_model_loader works

src/pb_stub.cc
```diff
@@ -79,9 +79,10 @@ Stub::Instantiate(
     const std::string& shm_region_name, const std::string& model_path,
     const std::string& model_version, const std::string& triton_install_path,
     bi::managed_external_buffer::handle_t ipc_control_handle,
-    const std::string& name)
+    const std::string& name, const std::string& plugin_model)
```
krishung5, line 82: nit: `plugin_model` -> `plugin_model_path` for consistency with the function signature.

CMakeLists.txt
```diff
@@ -304,6 +306,13 @@ install(
     ${INSTALL_CONFIGDIR}
 )
 
+install(
+  DIRECTORY
+    src/resources/plugin_models
```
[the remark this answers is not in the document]
tanmayv25: "model_loaders" just give an impression that it loads the model. However, the model.py loads and serve the models as well. How about "platform_models"?

src/python_be.cc
tanmayv25: I can push the logic of getting ArtifactName into platform model. We would need to parse the default_model_filename.

src/python_be.cc
tanmayv25: I think it will still be good to do a check whether a platform_model exists in the backend. If yes, then use the model.py from the platform_models or else fallback to previous pipeline.
This is to maintain backwards compatibility.

src/resources/plugin_models/tensorflow_savedmodel/model.py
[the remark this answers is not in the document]
tanmayv25: Let me try cleaning this up. The logic to retrieve default artifact name can be added here.

src/pb_stub.cc
tanmayv25: This is actually the name of the platform_model(previously plugin_model)

Tabrizian (commented):
Mostly looks good. Some minor comments.

src/pb_stub.cc
krishung5: I see. Then I think we would need to change the `platform_model_path` to `platform_model` here https://github.com/triton-inference-server/python_backend/pull/255/files#diff-b5637c6a5b8dd3bddd8aa5c4b9f2ac30895e75fd678cdcf66b50c9b0d607e94eR188.

src/stub_launcher.cc
```diff
@@ -61,6 +61,10 @@ StubLauncher::Initialize(ModelState* model_state)
   model_state->ModelConfig().Write(&model_config_buffer_);
   is_decoupled_ = model_state->IsDecoupled();
   model_repository_path_ = model_state->RepositoryPath();
+  platform_model_ = "DUMMY";
```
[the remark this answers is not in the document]
tanmayv25: Using "NONE" for more clarity.

Tabrizian (commented):
We probably need testing for all the parameters parsing too.

tanmayv25:
> We probably need testing for all the parameters parsing too.

I am working on adding testing as a separate ticket.

src/pb_stub.h
```diff
@@ -161,6 +161,30 @@ class LogMessage {
 
 #define LOG_FL(FN, LN, LVL) LogMessage((char*)(FN), LN, LVL).stream()
 
+
+class ModelScanner {
```
nnshah1, line 165: Instead of a separate class - is it possible to have the state return the reconciled model path and then keep the rest of the stub setup the same (just based on the path that was returned based on the platform field in model state)?

src/stub_launcher.cc
```diff
@@ -72,12 +76,13 @@ StubLauncher::Initialize(ModelState* model_state)
 
   model_version_ = model_state->Version();
 
+  // Should check for model.py only when not using platform models.
```
nnshah1, line 79: can remove comment - maybe move to commit message. If I understand correctly the check is now moved to stub

src/pb_stub.h
```diff
@@ -185,7 +209,8 @@ class Stub {
       const std::string& shm_region_name, const std::string& model_path,
       const std::string& model_version, const std::string& triton_install_path,
       bi::managed_external_buffer::handle_t ipc_control_handle,
-      const std::string& model_instance_name);
+      const std::string& model_instance_name,
+      const std::string& platform_model);
```
nnshah1, line 213: for consistency I would name this 'platform'

src/pb_stub.cc
```diff
@@ -1655,13 +1726,14 @@ main(int argc, char** argv)
   int64_t shm_growth_size = std::stol(argv[4]);
   std::string triton_install_path = argv[6];
   std::string name = argv[8];
+  std::string platform_model = argv[9];
```
nnshah1, line 1729: change to `platform`

src/pb_stub.cc
```diff
@@ -1613,6 +1593,97 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
       module, "TritonModelException");
 }
 
+
+void
```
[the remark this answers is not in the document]
tanmayv25: I would still prefer to keep the platform handler logic localized within a class. All the paths, versions and other details are now a part of ModelContext.

src/python_be.cc
```diff
@@ -1718,6 +1719,14 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
       }
     }
 
+    triton::common::TritonJson::Value platform;
+    if (model_config_.Find("platform", &platform)) {
+      auto error = platform.AsString(&platform_);
```
Tabrizian, line 1724: Can we add a check to disable platforms other than `tensorflow_savedmodel` for now as discussed?

Tabrizian (commented):
Looks good to me! One minor comment.

src/pb_stub.cc
```diff
@@ -1613,6 +1602,97 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
+    // Check if model file exists in the path.
+    struct stat buffer;
+    if (stat(platform_model_path.c_str(), &buffer) == 0) {
+      // Use the Platform model for serving the model.
+      python_model_found = true;
+      type_ = ModelType::PLATFORM;
+      python_model_path_ = platform_model_path;
+      // Trimming the model name from the model path, the platform model
+      // will populate the expected default model file name into model_path_.
+      model_path_ = model_path.substr(0, model_path.find_last_of("\\/"));
```
nnshah1, line 1626: should we set `model_path_` in non platform mode as well?

src/pb_stub.cc
```diff
@@ -1613,6 +1602,97 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
+    } else {
+      LOG_WARN << "Unable to find model(handler) \'" << platform_model_path
+               << "\' for platform field \'" << platform << "\'";
+    }
+  }
+
+  if (!python_model_found) {
+    python_model_path_ = model_path;
```
nnshah1, line 1634: potential: `python_module_path` to distinguish from `model`

src/python_be.cc
nnshah1: I think we wanted to keep for backwards compat - but we can put the check in the model context for handlers?

Tabrizian: platform for python backend is a new functionality so I don't think there would be any backward compat issues.

tanmayv25: @Tabrizian  The field always existed. There might be users using the field value for something specific in their model.py. The only check we have right now is to look for the presence of directory within platform_handlers. If found, then execute the python model. If not, then fallback to previous logic.

src/pb_stub.cc
tanmayv25: The model_path_ is the `python_model_path` for non-platform mode.  The model_path_ points model.py in the model-repo.

src/python_be.cc
Tabrizian: I wouldn't worry about that too much. This is not part of the API that we offer to the Python backend users. The platform field is for the core to consume not for the Python model developer.

src/pb_stub.cc
tanmayv25: I would prefer python_model_path for now... It makes more sense as we have always referred model.py as python model in Triton.

src/python_be.cc
tanmayv25: For disabling platform field for values other than "tensorflow_savedmodel", I believe it is ok to leave the logic as is. We are only installing tensorlfow_savedmodel in the platform handler. We would fallback to the previous implementation for all the rest field values.

Tabrizian (approved):

krishung5 (dismissed):

nnshah1 (approved):
@Tabrizian, @tanmayv25 LGTM - we can change things for GA

PR #259, opened 2023-06-14
Update neuron readme and setup

nskool:
In previous update to setup.sh making `inf1` default caused a break in `inf2` installation in pytorch. Fixing that.

### DIFF
--- a/inferentia/scripts/setup.sh
+++ b/inferentia/scripts/setup.sh
@@ -91,6 +91,7 @@ for OPTS; do
         ;;
         -inf1|--inf1-setup)
         INSTALL_INF1=1
+        INSTALL_INF2=0
         echo "Installing framework and tools for inf1."
         shift 1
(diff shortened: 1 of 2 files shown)


inferentia/scripts/setup.sh
```diff
@@ -96,11 +96,13 @@ for OPTS; do
         ;;
         -inf2|--inf2-setup)
         INSTALL_INF2=1
+        INSTALL_INF1=0
```
GuanLuo, line 99: I think for completeness, similar handling should be done if `-inf1` is specified (line 92), i.e. set `INSTALL_INF2=0` even it is 0 by default.

nskool: Thank you Guan for the quick review, just updated - if all looks good, request you to merge

GuanLuo (approved):

PR #262, opened 2023-06-23
Model loading API for Python BLS

krishung5:
This PR adds support for the model loading API in Python BLS. It enables users to load the required model during the initialization of their BLS model.

Closes: https://github.com/triton-inference-server/server/issues/4366
Testing: https://github.com/triton-inference-server/server/pull/5980

### DIFF
--- a/src/metric.cc
+++ b/src/metric.cc
@@ -144,16 +144,12 @@ void
 Metric::SendSetValueRequest(const double& value)
 {
   try {
-<<<<<<< HEAD
     CheckIfCleared();
     std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
     operation_value_ = value;
     SaveToSharedMemory(stub->ShmPool());
     CustomMetricsMessage* custom_metrics_msg = nullptr;
-    stub->SendCustomMetricsMessage(
-=======
     stub->SendMessage<CustomMetricsMessage>(
->>>>>>> Use template functions for custom metrics
         &custom_metrics_msg, PYTHONSTUB_MetricRequestSet, shm_handle_);
   }
   catch (const PythonBackendException& pb_exception) {
@@ -167,14 +163,10 @@ Metric::SendGetValueRequest()
 {
   CustomMetricsMessage* custom_metrics_msg = nullptr;
   try {
-<<<<<<< HEAD
     CheckIfCleared();
     std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
     SaveToSharedMemory(stub->ShmPool());
-    stub->SendCustomMetricsMessage(
-=======
     stub->SendMessage<CustomMetricsMessage>(
->>>>>>> Use template functions for custom metrics
         &custom_metrics_msg, PYTHONSTUB_MetricRequestValue, shm_handle_);
   }
   catch (const PythonBackendException& pb_exception) {
--- a/src/metric.cc
+++ b/src/metric.cc
@@ -65,7 +65,7 @@ Metric::SaveToSharedMemory(std::unique_ptr<SharedMemoryManager>& shm_pool)
   // Save the references to shared memory.
   custom_metric_shm_ = std::move(custom_metric_shm);
   labels_shm_ = std::move(labels_shm);
-  shm_handle_ = custom_metric_shm.handle_;
+  shm_handle_ = custom_metric_shm_.handle_;
 }
 
 std::unique_ptr<Metric>
@@ -111,7 +111,7 @@ Metric::SendCreateMetricRequest()
   SaveToSharedMemory(stub->ShmPool());
   CustomMetricsMessage* custom_metrics_msg = nullptr;
   try {
-    stub->SendCustomMetricsMessage(
+    stub->SendMessage<CustomMetricsMessage>(
         &custom_metrics_msg, PYTHONSTUB_MetricRequestNew, shm_handle_);
   }
   catch (const PythonBackendException& pb_exception) {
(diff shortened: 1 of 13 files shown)


MatthieuTPHR:
This is amazing ! Can't wait to see this merged ! 😍

src/pb_stub.cc
```diff
@@ -1604,6 +1546,17 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
   module.attr("MetricFamily").attr("COUNTER") = MetricKind::COUNTER;
   module.attr("MetricFamily").attr("GAUGE") = MetricKind::GAUGE;
 
+  module.def(
+      "load_model", &LoadModel, py::arg("model_name").none(false),
+      py::arg("config").none(false) = "",
+      py::arg("files").none(false) = py::dict());
```
Tabrizian, line 1552: I think empty dictionary as initializer can have unexpected consequences. Can we use none here instead and do type checking inside `load_model`?

https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

src/model_loader.h
```diff
@@ -0,0 +1,158 @@
...
+  /// Send a request to load the model.
+  void SendLoadModelRequest();
+  /// Send a request to unload the model.
+  void SendUnloadModelRequest();
+  /// Send a request to check if the model is ready.
+  bool SendModelReadinessRequest();
+#else
+  /// Use Triton C API to load the model.
+  /// \param server The Triton server object.
+  void LoadModel(TRITONSERVER_Server* server);
```
Tabrizian, line 105: It would be good to leave one empty line between each function definition.

src/metric_family.cc
```diff
@@ -59,7 +59,7 @@ MetricFamily::~MetricFamily()
   SaveToSharedMemory(stub->ShmPool());
   CustomMetricsMessage* custom_metrics_msg = nullptr;
```
Tabrizian, line 60: Is this used anywhere? Can we remove the first argument of `SendCustomMetricsMessage`?

Tabrizian (commented):
The API looks great! Some minor comments.

src/metric_family.cc
krishung5: This argument is used when the parent process returns some value back to the stub. For example, [the address of metric_family](https://github.com/triton-inference-server/python_backend/blob/f2156ad5fcd4c421a83a5091b30ccd5b3b3f7f80/src/metric_family.cc#L162), [the value of a metric](https://github.com/triton-inference-server/python_backend/blob/f2156ad5fcd4c421a83a5091b30ccd5b3b3f7f80/src/metric.cc#L177), and the [model readiness](https://github.com/triton-inference-server/python_backend/blob/f2156ad5fcd4c421a83a5091b30ccd5b3b3f7f80/src/model_loader.cc#L147).

Tabrizian: How is the lifetime of this object managed? It looks like the unique pointer holding this object will be deallocated in `SendMessage` and this can cause segfault because of use after free.

krishung5: Nice catch! Fixed the lifetime issue by passing the unique pointer of `ipc_message` to the template function.

Tabrizian: I don't think the lifetime is fixed. The problem is `AllocatedSharedMemory<MessageType> msg_shm` will be destructed in `SendMessage`.

krishung5: You're right. `AllocatedSharedMemory<MessageType> msg_shm` should be managed by the caller function to `SendMessage` so that when retrieving the value from the `MessageType* msg_ptr` it won't be used after free.

src/metric.cc
```diff
@@ -130,8 +134,12 @@ Metric::SendIncrementRequest(const double& value)
     operation_value_ = value;
     SaveToSharedMemory(stub->ShmPool());
     CustomMetricsMessage* custom_metrics_msg = nullptr;
-    stub->SendCustomMetricsMessage(
-        &custom_metrics_msg, PYTHONSTUB_MetricRequestIncrement, shm_handle_);
+    AllocatedSharedMemory<CustomMetricsMessage> custom_metrics_shm;
+    std::unique_ptr<IPCMessage> ipc_message;
+
+    stub->SendMessage<CustomMetricsMessage>(
+        ipc_message, &custom_metrics_msg, custom_metrics_shm,
```
Tabrizian, line 141: I think we no longer need to have custom_metrics_msg separately. It can be created from `custom_metrics_shm`.

src/metric_family.cc
Tabrizian: I don't think there was a lifetime issue with `ipc_message`. It was only an issue with msg_shm so I think we can remove it from SendMessage.

src/pb_stub.h
```diff
@@ -348,4 +351,70 @@ class Stub {
   std::mutex dlpack_proxy_stream_pool_mu_;
   std::unordered_map<int, cudaStream_t> dlpack_proxy_stream_pool_;
 };
+
+template <typename MessageType>
+void
+Stub::PrepareMessage(
+    AllocatedSharedMemory<MessageType>& msg_shm, MessageType** msg)
```
Tabrizian, line 358: I think you can remove `msg` as well from the argument list.

krishung5: I was wondering why we can remove `msg` from the argument list. This function is to initialize some variables of the `msg`, for example, the mutex and the condition variable.

Tabrizian: As far as I can tell from the code, you can access `msg` from `msg_shm` so I'm not sure why you need both of them to be passed.

krishung5: I understand it now. Updated the `PrepareMessage`. Thanks for the clarification!

Tabrizian (approved):

PR #263, opened 2023-06-23
Improve CUDA context management

Tabrizian:
There was an extra CUDA context created when async BLS was used.

The root cause for this issue is that the primary CUDA context is managed per-thread. It means that if you change the CUDA context for one thread, it will not be changed for other threads. See this article for more info: https://developer.nvidia.com/blog/cuda-pro-tip-always-set-current-device-avoid-multithreading-bugs/

This resulted in an extra CUDA context for GPU0 in multi-GPU deployments of Python backend.

Inspired from [MaybeSetDevice](https://github.com/pytorch/pytorch/blob/727458239027222a630c72354adf42c28041ff6f/c10/cuda/CUDAFunctions.cpp#L190-L206) from PyTorch, the restoration of old context would only happen if there is already a primary CUDA context for that device.

### DIFF
--- a/src/pb_utils.cc
+++ b/src/pb_utils.cc
@@ -61,7 +61,7 @@ CUDAHandler::CUDAHandler()
 {
   dl_open_handle_ = dlopen("libcuda.so", RTLD_LAZY);
 
-  // If libcuda.so is succesfully opened, it must be able to find
+  // If libcuda.so is successfully opened, it must be able to find
   // "cuPointerGetAttribute", "cuGetErrorString", and
   // "cuDevicePrimaryCtxGetState" symbols.
   if (dl_open_handle_ != nullptr) {


src/pb_utils.h
```diff
@@ -255,7 +259,27 @@ class CUDAHandler {
...
+
+  /// Set the device only if the primary context has already been created for
+  /// this device. Inspired from PyTorch's MaybeSetDevice.
+  /// \param device The cuda device index.
+  void MaybeSetDevice(int device);
+};
+
+
+/// A helper class to change the current device and restore the old context. The
+/// old context will be restored only when if the primary context for that
```
oandreeva-nv, line 271: *be restored only ~when~ if the primary ..

krishung5 (commented):
Great improvement!
nit: I think some copyright needs to be updated and the files need to be formatted, lgtm otherwise.

src/pb_utils.cc
```diff
@@ -86,6 +86,13 @@ CUDAHandler::CUDAHandler()
     }
     *((void**)&cu_init_fn_) = cu_init_fn;
 
+    void* cu_device_primary_ctx_get_state_fn = dlsym(dl_open_handle_, "cuDevicePrimaryCtxGetState");
```
oandreeva-nv, line 89: Should we update comment on L63 to keep everything consistent?

oandreeva-nv:
Thank you for this PR!
Could you please clarify the original problem? The context creation on GPU0, when it wasn't needed or the second context on GPU0 was created at some point?

src/pb_utils.h
Tabrizian: good catch. Fixed.

Tabrizian:
> Could you please clarify the original problem? The context creation on GPU0, when it wasn't needed or the second context on GPU0 was created at some point?

The issue was that there was an additional CUDA context created on GPU0 when none of the input/output tensors where using GPU0. The root cause was that when  we call cudaGetDevice on thread A and before that we have set cuda device to device N (N is any other number than 0) on thread B, the CudaGetDevice on thread A still returns 0 and when we restore the context it would create the additional CUDA context.

oandreeva-nv:
@Tabrizian I see, thank you for clarification. Amazing work on the root-causing the issue!

oandreeva-nv (approved):

krishung5 (approved):

dyastremsky (approved):

PR #264, opened 2023-06-24
WIP Allow Python backend to directly write Numpy arrays to SHM 

asos-danielbunting:


Tabrizian (commented):
@asos-danielbunting thanks for the PR. I was wondering what is the use-case that this PR is trying to address? Is the idea to pre-allocate the buffers in shared memory and directly work with them to speed up the inference process?

Could you please share more details about the places where this becomes useful?

asos-danielbunting:
Hi @Tabrizian I'm looking at trying to speed up passing a large tensor between a Python BLS model doing preprocessing and a Tensorflow inference model. 

As you say the idea is to allocate the buffer and directly write my data into it from the python side and so avoid an extra allocation + copy time. I've run a couple of tests and for my use case this can speed up my inference time by a decent amount eg for a 100000 x 200 float32 tensor the saving was 30ms

Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM asnpdsacr.azurecr.io/public/tritonserver:23.05-tf2-python-py3
```
Tabrizian, line 1: Remove

src/pb_stub.cc
```diff
@@ -431,8 +431,12 @@ Stub::StubSetup()
   py::setattr(
```
Tabrizian, line 431: Remove all the changes except the ones in the `src` directory.

src/pb_stub.cc
```diff
@@ -431,8 +431,12 @@ Stub::StubSetup()
...
       python_backend_utils, "MetricFamily",
       c_python_backend_utils.attr("MetricFamily"));
+  py::setattr(
+      python_backend_utils, "new_shm_tensor",
+      c_python_backend_utils.attr("new_shm_tensor"));
 
   c_python_backend_utils.attr("shared_memory") = py::cast(shm_pool_.get());
+  python_backend_utils.attr("shared_memory") = py::cast(shm_pool_.get());
```
Tabrizian, line 439: This is not needed.

src/pb_stub.cc
```diff
@@ -494,6 +498,7 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
       python_backend_utils, "InferenceResponse",
       c_python_backend_utils.attr("InferenceResponse"));
   c_python_backend_utils.attr("shared_memory") = py::cast(shm_pool_.get());
+  python_backend_utils.attr("shared_memory") = py::cast(shm_pool_.get());
```
Tabrizian, line 501: Not required.

src/pb_stub.cc
```diff
@@ -1603,6 +1608,8 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
 
   py::register_exception<PythonBackendException>(
       module, "TritonModelException");
+
+  module.def("new_shm_tensor", &PbTensor::CreateInSHM, "Creates a new Tensor directly into shared memory");
```
Tabrizian, line 1612: Can we rename this to `pb.Tensor.new(shape, dtype, device='cpu')`?

src/pb_tensor.cc
```diff
@@ -226,6 +229,68 @@ delete_unused_dltensor(PyObject* dlp)
...
+
+    int64_t memory_type_id_ = 0; // Maybe
+
+    auto pb_memory = PbMemory::Create(
+          memory_type_, memory_type_id_, byte_size_,
+          nullptr,
+          reinterpret_cast<char*>(tensor_shm_ptr) + pb_memory_offset,
+          shm_handle + pb_memory_offset, false);
+    tensor_shm_ptr->memory = 0;
+    std::cout << "Offset is - " << pb_memory_offset<<  "\n";
```
Tabrizian, line 287: Remove print statement.

src/pb_tensor.cc
```diff
@@ -226,6 +229,68 @@ delete_unused_dltensor(PyObject* dlp)
   }
 }
 
+std::shared_ptr<PbTensor>
+PbTensor::CreateInSHM(const std::string& name, SharedMemoryManager& shm_pool, std::vector<int64_t> dims, py::object data_type)
+{
+
+    // Input params of tensor
+    //std::vector<int64_t> dims = std::vector<int64_t>({10, 10});
```
Tabrizian, line 237: Remove comment.

PR #266, opened 2023-06-29
Allow uncompressed conda execution enviroments

jbkyang-nvi:
Allow uncompressed conda execution environments and add documentation for custom execution environments.
Test: https://github.com/triton-inference-server/server/pull/6005

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,8 @@ any C++ code.
   - [Running Multiple Instances of Triton Server](#running-multiple-instances-of-triton-server)
 - [Business Logic Scripting](#business-logic-scripting)
   - [Using BLS with Decoupled Models](#using-bls-with-decoupled-models)
-  - [Using BLS with Stateful Models](#using-bls-with-stateful-models)
   - [Model Loading API](#model-loading-api)
+  - [Using BLS with Stateful Models](#using-bls-with-stateful-models)
   - [Limitation](#limitation)
 - [Interoperability and GPU Support](#interoperability-and-gpu-support)
   - [`pb_utils.Tensor.to_dlpack() -> PyCapsule`](#pb_utilstensorto_dlpack---pycapsule)
@@ -706,9 +706,9 @@ If this variable is not exported and similar packages are installed outside your
 conda environment, your tar file may not contain all the dependencies required
 for an isolated Python environment.
 
-Alternatively, Triton also supports unpacked Conda execution environments, given
-it  points to an activation script to setup the Conda environment. Usually the
-Conda activation script is located in:
+Alternatively, Python backend also supports unpacked conda execution 
+environments, given it points to an activation script to setup the conda 
+environment. Usually the conda activation script is located in:
 ````$path_to_conda_pack/lib/python3.7/site-packages/conda_pack/scripts/posix/activate```
 This speeds up the server loading time for models.
 
--- a/README.md
+++ b/README.md
@@ -681,7 +681,7 @@ If you want to create a tar file that contains all your Python dependencies or
 you want to use different Python environments for each Python model you need to
 create a *Custom Execution Environment* in Python backend.
 Currently, Python backend supports
-[conda-pack](https://conda.github.io/conda-pack/) for this purpose. 
+[conda-pack](https://conda.github.io/conda-pack/) for this purpose.
 [conda-pack](https://conda.github.io/conda-pack/) ensures that your conda
 environment is portable. You can create a tar file for your conda environment
 using `conda-pack` command:
@@ -706,10 +706,10 @@ If this variable is not exported and similar packages are installed outside your
 conda environment, your tar file may not contain all the dependencies required
 for an isolated Python environment.
 
-Alternatively, Triton also supports unpacked Conda execution environments, given 
-it  points to an activation script to setup the Conda environment. Usually the 
+Alternatively, Triton also supports unpacked Conda execution environments, given
+it  points to an activation script to setup the Conda environment. Usually the
 Conda activation script is located in:
-````$path_to_conda_pack/lib/python3.7/site-packages/conda_pack/scripts/posix/activate``` 
+````$path_to_conda_pack/lib/python3.7/site-packages/conda_pack/scripts/posix/activate```
 This speeds up the server loading time for models.
 
 After creating the tar file from the conda environment or creating a conda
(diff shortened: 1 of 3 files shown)


src/pb_env.cc
```diff
@@ -253,6 +253,18 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
 
   bool env_extracted = false;
   bool re_extraction = false;
+
+  // If the path is not a conda-packed file, then bypass the extraction process
+  std::string subPath = env_path.substr(env_path.size() - 6);
```
dyastremsky, line 258: I think this can throw an error. If the env_path is < 6 characters, size_t will interpret it as a very large number (close to the max possible value), which will throw an out of range.

Reference: https://cplusplus.com/reference/string/string/substr/

src/pb_env.cc
```diff
@@ -253,6 +253,18 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+  if (subPath != "tar.gz") {
+    LOG_MESSAGE(
+        TRITONSERVER_LOG_VERBOSE,
```
dyastremsky, line 261: This might be useful to have at the info-level logging level.

dyastremsky (changes requested):
Looks good! Added a couple of questions. Also, are you adding documentation for this feature somewhere?

jbkyang-nvi:
> Looks good! Added a couple of comments. Also, are you adding documentation for this feature somewhere?

This is the documentation

dyastremsky:
> > Looks good! Added a couple of comments. Also, are you adding documentation for this feature somewhere?
> 
> This is the documentation in the README

Missed that. Thanks for clarifying!

src/pb_env.cc
jbkyang-nvi: I'm keeping this inline with the other messages. See line 287

dyastremsky (approved):

src/pb_env.cc
```diff
@@ -253,6 +253,19 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
 
   bool env_extracted = false;
   bool re_extraction = false;
+
+  // If the path is not a conda-packed file, then bypass the extraction process
+  std::size_t subStrPos = env_path.size() > 6 ? env_path.size() - 6 : 0;
+  std::string subPath =  env_path.substr(subStrPos);
+  if (subPath != "tar.gz") {
```
rmccorm4, line 260: Instead of using magic numbers like `6`, can you add a helper like `endsWith(string, suffix)` and use it here? I think it will be clearer. 

I believe PB already has boost dependencies installed, so you can probably just use:
```cpp
if (!boost::algorithm::ends_with(env_path, "tar.gz")) {
  ...
}
```
Otherwise lots of simple implementations here: https://www.techiedelight.com/check-if-a-string-ends-with-another-string-in-cpp/
```
if (!EndsWith(env_path, "tar.gz")) {
  ...
}
```

rmccorm4 (dismissed):

dyastremsky (approved):

Tabrizian (changes requested):
@jbkyang-nvi Have you verified that the environment wouldn't be deleted after the server terminates? By default Python backend would delete the environments once it terminates.

README.md
```diff
@@ -703,7 +706,14 @@ If this variable is not exported and similar packages are installed outside your
 conda environment, your tar file may not contain all the dependencies required
 for an isolated Python environment.
 
-After creating the tar file from the conda environment, you need to tell Python
+Alternatively, Triton also supports unpacked Conda execution environments, given
+it  points to an activation script to setup the Conda environment. Usually the
+Conda activation script is located in:
```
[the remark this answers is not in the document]
jbkyang-nvi: This is not true. The user needs to add it themselves.

README.md
```diff
@@ -703,7 +706,14 @@ If this variable is not exported and similar packages are installed outside your
...
+````$path_to_conda_pack/lib/python3.7/site-packages/conda_pack/scripts/posix/activate```
+This speeds up the server loading time for models.
+
+After creating the tar file from the conda environment or creating a conda
```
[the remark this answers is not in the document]
jbkyang-nvi: see next sentence where I mention conda execution environment which is "unpacked"

README.md
```diff
@@ -703,7 +706,14 @@ If this variable is not exported and similar packages are installed outside your
...
+environment with a custom activation script, you need to tell Python
```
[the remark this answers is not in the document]
jbkyang-nvi: I do not think we need to force the user to us conda-pack and then unpack their environment for Triton to use it. They can simply copy the activation script over to the right path for Triton

src/pb_env.cc
```diff
@@ -253,6 +254,18 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
 
   bool env_extracted = false;
   bool re_extraction = false;
+
+  // If the path is not a conda-packed file, then bypass the extraction process
+  struct stat info;
+  if (stat(canonical_env_path, &info) == 0 && S_ISDIR(info.st_mode)) {
```
Tabrizian, line 260: I think we need to return an error if stat returns non-zero code. 

Would it be possible to use boost is directory?
https://www.boost.org/doc/libs/1_79_0/libs/filesystem/doc/reference.html#is_directory

README.md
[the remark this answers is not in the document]
Tabrizian: I think it is a bit hacky to ask the users to copy the activation script.

For me the activation script is in a different location: ./lib/python3.9/venv/scripts/common/activate

I'm a bit hesitant to ask the users to copy their activation script since it looks like it depends on conda version.

I think asking the users to use conda-unpack or having a utility that automatically looks for the activation script is a safer path.

src/pb_env.cc
jbkyang-nvi: The current boost installation we have some linking issues with `libtriton_python.so` and `filesystem.hpp`.  We also use `stat` elsewhere in this folder. Lets keep it as is for now and update everything later? I think `boost::exists` is faster than `stat()`.

README.md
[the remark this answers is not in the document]
jbkyang-nvi: chatted offline with @Tabrizian. It seems like it is a bit restricting to require the user to have to use `conda-pack` and then `conda-unpack` for all conda environments, especially if they are trying to save time by using the environment in its unpacked form in the first place.

Will follow up with auto-searching for the activation script in a later PR. However must note there's a risk we will find the wrong activation script (the python venv activation script does not work for a conda environment, as an example).

Tabrizian (approved):

PR #270, opened 2023-07-07
Fallback to CPU tensors if cuInit call fails

tanmayv25:
cuInit can fail if there are no GPUs are available in the system.

### Before

```
terminate called after throwing an instance of 'triton::backend::python::PythonBackendException'
  what():  failed to get cuda pointer device attribute: no CUDA-capable device is detected

Thread 41 "tritonserver" received signal SIGABRT, Aborted.
```

### After

The model loading and inference runs successfully.

Test for these changes are part of the TF platflorm handler test changes.

### DIFF
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -462,7 +462,7 @@ ModelInstanceState::GetInputTensor(
   CUDAHandler& cuda_handler = CUDAHandler::getInstance();
   // If CUDA driver API is not available, the input tensors will be moved to
   // CPU.
-  if (!cuda_handler.IsAvailable()) {
+  if (!cuda_handler.IsAvailable() && !cpu_only_tensors) {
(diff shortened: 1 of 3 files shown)


src/python_be.cc
```diff
@@ -463,6 +463,14 @@ ModelInstanceState::GetInputTensor(
   // If CUDA driver API is not available, the input tensors will be moved to
   // CPU.
   if (!cuda_handler.IsAvailable()) {
```
Tabrizian, line 465:
```suggestion
  if (!cuda_handler.IsAvailable() && !cpu_only_tensors) {
```

jbkyang-nvi (approved):

src/python_be.cc
tanmayv25: That makes sense! Thanks for the catch!

krishung5 (approved):

PR #277, opened 2023-07-14
Adding the support tracing of child models invoked from a BLS model

oandreeva-nv:
To allow tracing for sub models, user must provide `request` when creating `InferenceRequest` object in python model.
Sample:
```
for request in requests:
           ...

            infer_request = pb_utils.InferenceRequest(
                model_name=model_name_string,
                requested_output_names=["OUTPUT0", "OUTPUT1"],
                inputs=[in_0, in_1],
                request=request)
```
This is necessary, because `request` has an access to the main trace. Open to the discussion on this.

### DIFF
--- a/src/infer_request.h
+++ b/src/infer_request.h
@@ -42,6 +42,17 @@ namespace triton { namespace backend { namespace python {
 class Stub;
 
 //
+// Inference Trace
+//
+struct InferenceTrace {
+#ifndef TRITON_PB_STUB
+  TRITONSERVER_InferenceTrace* triton_trace_;
+#else
+  void* triton_trace_;
+#endif
+};
+
+//
 // Inference Request
 //
 struct InferRequestShm {
@@ -55,7 +66,7 @@ struct InferRequestShm {
   bool is_decoupled;
   int32_t timeout;
   PreferredMemory preferred_memory;
-  TRITONSERVER_InferenceTrace* trace;
+  InferenceTrace trace;
 };
 
 class InferRequest {
@@ -70,7 +81,7 @@ class InferRequest {
       const intptr_t request_address = 0,
       const PreferredMemory& preferred_memory =
           PreferredMemory(PreferredMemory::DEFAULT, 0),
-      TRITONSERVER_InferenceTrace* trace = nullptr);
+      const InferenceTrace& trace = {.triton_trace_ = nullptr});
 
   const std::vector<std::shared_ptr<PbTensor>>& Inputs();
   const std::string& RequestId();
@@ -86,7 +97,7 @@ class InferRequest {
   bool IsDecoupled();
   void SetIsDecoupled(const bool is_decoupled);
   PreferredMemory& GetPreferredMemory();
-  TRITONSERVER_InferenceTrace* Trace();
+  InferenceTrace& Trace();
 
 #ifdef TRITON_PB_STUB
   std::shared_ptr<InferResponse> Exec(const bool is_decoupled);
@@ -142,7 +153,7 @@ class InferRequest {
   intptr_t request_address_;
   bool is_decoupled_;
   PreferredMemory preferred_memory_;
-  TRITONSERVER_InferenceTrace* trace_;
+  InferenceTrace trace_;
 
   // Shared Memory Data Structures
   AllocatedSharedMemory<char> infer_request_shm_;
--- a/src/infer_request.h
+++ b/src/infer_request.h
@@ -55,6 +55,7 @@ struct InferRequestShm {
   bool is_decoupled;
   int32_t timeout;
   PreferredMemory preferred_memory;
+  TRITONSERVER_InferenceTrace* trace;
 };
 
 class InferRequest {
@@ -85,6 +86,9 @@ class InferRequest {
   bool IsDecoupled();
   void SetIsDecoupled(const bool is_decoupled);
   PreferredMemory& GetPreferredMemory();
+  TRITONSERVER_InferenceTrace* Trace();
+  uint32_t CustomTrace();
+
 
 #ifdef TRITON_PB_STUB
   std::shared_ptr<InferResponse> Exec(const bool is_decoupled);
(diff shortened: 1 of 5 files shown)


src/infer_request.h
```diff
@@ -139,6 +142,7 @@ class InferRequest {
   intptr_t request_address_;
   bool is_decoupled_;
   PreferredMemory preferred_memory_;
+  TRITONSERVER_InferenceTrace* trace_;
```
Tabrizian, line 145: I wonder how this compiles? The infer request class needs to protect the parts that require access to `libtritonserver.so` by `#ifdef TRITON_PB_STUB`. We don't link `libtritonserver.so` to the stub process.

src/pb_stub.cc
```diff
@@ -1371,17 +1371,19 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
                       const std::string& model_name,
                       const int64_t model_version, const uint32_t flags,
                       const int32_t timeout,
-                      const PreferredMemory& preferred_memory) {
+                      const PreferredMemory& preferred_memory,
+                      std::shared_ptr<InferRequest>& request) {
```
Tabrizian, line 1375: How about adding an extra argument named `trace` for parity with the C-API in exec/async_exec ? 

We can add a `.trace()` argument for each `request` object that returns the current trace for each inference request. The API would look like below:

```
trace = request.trace()
infer_request = pb_utils.InferenceRequest(...)
child_trace = pb_utils.InferenceTrace.create_child_trace(trace)
infer_request.exec(child_trace)
``` 

 @GuanLuo / @tanmayv25  what do you think?
I think an inference request that has a request argument looks a bit strange..

oandreeva-nv: I had this as an option. Upon discussion with Tanmay, an idea of passing request to the constructor came from the possibility of using some additional objects, extracted from the request in future. 

`request.trace()` would expose a void pointer to the trace to user, and I am not sure if it makes sense to do that.

src/infer_request.h
oandreeva-nv: It compiles, and I've tested BLS traces with TRITON Tracing APIs and OpenTelemetry, but now after you mentioned that, I'll probably need to verify what happens if `TRITON_PB_STUB` is not defined.

src/pb_stub.cc
Tabrizian: cc @nnshah1 for viz

Tabrizian: > request.trace() would expose a void pointer to the trace to user, and I am not sure if it makes sense to do that.

I don't think we have to expose a raw pointer. We can wrap it in a `InferenceTrace` object. We can probably add more methods to this object to get id, model name, request id, etc.

oandreeva-nv: Sure, that can be an option (cc @GuanLuo , @tanmayv25 , @nnshah1)

GuanLuo: Do we foresee that Python BLS user will interact with trace object directly? I can think of custom trace activity which the user will need an API to capture the activity (although it can be `request.log_trace_activity()`).

A separate node on passing request as argument, a user will expect some documentation on the relationship between the two requests and are there any constraints, etc. From the signature, I know that one is derived from the other, but I would want to know what are the attributes that the child inherits, and whether I need to keep the parent valid while interacting with the children.

nnshah1: I would agree using a trace parameter / object here instead of the entire request makes sense. I think we can make it opaque if we want to (i.e. not provide any methods around it except to retrieve it from a request). If a request doesn't have a trace - can we use `None` ?

oandreeva-nv: @GuanLuo Yes, there is a request for adding custom spans for OpenTelemetry tracing. I imagined this to be implemented through `pb_utils`, something like:
```
pb_utils.trace_start_span(...)
pb_utils.trace_add_activity(...)
pb_utils.trace_end_span(...)
```
and as a parameter we can pass `request` , or (based on current discussion) `request.trace()`.

@nnshah1 Noted, regarding `None`, I am not allowing `None` for `request` (or `trace`) , to keep it consistent with current constructor implementation: https://github.com/triton-inference-server/python_backend/blob/main/src/pb_stub.cc#L1393

nnshah1: @oandreeva-nv - Got it thanks! I think I misread the thread above. If creating a full object for a trace is cumbersome for the first round - I think keeping it opaque is fine.

Tabrizian: > I imagined this to be implemented through pb_utils, something like:
pb_utils.start_span_for_trace(...)
pb_utils.add_activity_to_trace(...)
pb_utils.end_span_for_trace(...)

We don't want to introduce open telemetry specific terminology in the Python utils. The way I think about it is that we didn't need these APIs in the C-API so there shouldn't be any need to do that in Python API as well.

Having an opaque trace object for now sounds good to me.

> Do we foresee that Python BLS user will interact with trace object directly? 

I think there maybe some use-cases for that. For example, in the future we may give the user the ability to create their own trace object (instead of having to always use a pre-created object from the parent request) and they would get the metrics about the request execution of the BLS request.

oandreeva-nv: @Tabrizian Just an FYI, we do have a request for an access to OpenTelemetry tracer from BLS: DLIS-5106. However, it may probably be implemented by creating a custom trace

Tabrizian: Looking at the discussion, I still don't think we have to provide OTel specific APIs. I think if we just allow the backend to attach custom information it should satisfy the requirements.

tanmayv25: pros of opaque Trace object:
1. Gives more flexibility for model developer. The user can still pass nullptr to the Request constructor and that request will not be traced. 
2. It matches with the C API design.

These come with a cost of more boilerplate code for forming the request objects.

oandreeva-nv: I've added a struct for the Trace object, containing only `TRITONSERVER_InferenceTrace`. I think for this PR it should be sufficient. Later, when we will work on custom tracing for BLS models, we can populate it/change to class.

Tabrizian (approved):

rmccorm4 (approved):

tanmayv25 (approved):

PR #282, opened 2023-07-26
Add PyTorch platform handler

kthui:
Add the ability to run PyTorch models on Python backend without having to craft a `model.py` file.

PR for linking docs: https://github.com/triton-inference-server/server/pull/6139


src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self._model_module = _import_module_from_path(self._model_name, model_path)
+        self._model_class = getattr(self._model_module, self._model_name)
+        self._raw_model = self._model_class()
+        if data_path != "":
+            self._raw_model.load_state_dict(torch.load(data_path))
+        else:
+            self._logger.log_info("Model weights not found for " + self._model_name)
+        self._raw_model.eval()
+        self._model = torch.compile(self._raw_model)
```
Tabrizian, line 115: I think this should be an option and shouldn't be enabled by default.

src/resources/platform_handlers/pytorch_libtorch/README.md
```diff
@@ -0,0 +1,88 @@
...
+`-- pytorch_model
+    |-- 1
+    |   |-- model.py
+    |   `-- model.py.pt
+    `-- config.pbtxt
+```
+
+The `model.py` contains the class definition of the PyTorch model. The class
+name needs to match the model name (i.e. `pytorch_model`), and the class should
+extend the
```
Tabrizian, line 53: does this mean that the class name in model.py file should be `class pytorch_model`? Do you have an example model.py to share?

src/resources/platform_handlers/pytorch_libtorch/README.md
```diff
@@ -0,0 +1,88 @@
...
+[`torch.nn.Module`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module).
+
+The `model.py.pt` may be optionally provided which contains the saved
+[`state_dict`](https://pytorch.org/tutorials/beginner/saving_loading_models.html#saving-loading-model-for-inference)
+of the model.
+
+By default, Triton will use the
+[PyTorch backend](https://github.com/triton-inference-server/pytorch_backend) to
+load and serve PyTorch models in TorchScript format. In order to use the Python
+backend to serve PyTorch models directly without TorchScript,
```
Tabrizian, line 63: I think we can even provide `TorchScript` support in Python backend if anyone is having troubles with memory issues.

src/resources/platform_handlers/pytorch_libtorch/README.md
Tabrizian: I just saw the example you shared in Teams chat. Is there a way to automatically find the model name instead of forcing the user to specify `pytorch_model`? It looks like TorchServe is able to do that.

kthui: Yes. [Add ability to auto find model class](https://github.com/triton-inference-server/python_backend/pull/282/commits/03893163f551e1fdf11a3d145831eca5f1f192b7)

src/resources/platform_handlers/pytorch_libtorch/README.md
kthui: Yes. [Add TorchScript support](https://github.com/triton-inference-server/python_backend/pull/282/commits/af568648aba1369816a78b841ddd92632aebe70d)

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: This is a good point. I think it would be better to allow opt-out instead of opt-in, as long as proper checks for its support are in place (i.e. PyTorch >= 2.0).

The user is able to disable it in parameters: [Add ability to disable torch.compile](https://github.com/triton-inference-server/python_backend/pull/282/commits/9e83e246c6f20948b7e9dcee82b3f8b1a30aad83)

src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,238 @@
...
+
+import importlib
+import json
+import os
+
+try:
+    import torch
+except ModuleNotFoundError as error:
+    raise RuntimeError(
+        "Missing/Incomplete PyTorch package installation... (Did you run `pip3 install torch`?)"
```
oandreeva-nv, line 37: Do you think that we need to specify version of torch here?

src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,238 @@
...
+                torch.load(data_path, map_location=self._device)
+            )
+        else:
+            self._logger.log_info(
+                "Model parameter file not found for '" + self._model_name + "'"
+            )
+        self._raw_model.to(self._device)
+        self._raw_model.eval()
+        if _enable_torch_compile(self._model_config):
+            self._model = torch.compile(self._raw_model)  # PyTorch 2.0+ only
```
oandreeva-nv, line 173: Could you please move this comment int a separate line

src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,238 @@
...
+
+        Returns
+        -------
+        list
+          A list of pb_utils.InferenceResponse. The length of this list must
+          be the same as `requests`
+        """
+        responses = []
+
+        # FIXME: Instead of iterating through each request, run the inference as a single batch.
```
oandreeva-nv, line 203: Could you please clarify for me, is there a JIRA ticket for this (i.e. is there a feature request).

kthui: I cannot find the ticket on JIRA. @tanmayv25 @Tabrizian Do you know if we are going to provide `requests` as batched tensor(s) instead of having to iterate over each `request` on a batch, in the future?

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: I only want to remind the user the needs to install `torch`, but not create an additional "maintenance item" if we are going to upgrade the torch version in the future, so I changed it to `(Did you install PyTorch?)`.

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: I moved the comment into `_enable_torch_compile()` function, because the version is checked over there.

src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,238 @@
...
+            if not isinstance(raw_output_tensors, tuple) and not isinstance(
+                raw_output_tensors, list
+            ):
+                raw_output_tensors = [raw_output_tensors]
+
+            output_tensors = []
+            for i in range(len(self._outputs)):
+                io = self._outputs[i]
+                tensor = torch.to_dlpack(raw_output_tensors[i].detach())
+                tensor = pb_utils.Tensor.from_dlpack(io["name"], tensor)
```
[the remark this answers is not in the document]
kthui: Great catch! I did not know the info is already embedded. It is working without explicitly calling `pb_utils.Tensor.from_dlpack()`.

src/resources/platform_handlers/pytorch_libtorch/README.md
```diff
@@ -0,0 +1,86 @@
...
+-->
+
+# Serving PyTorch models using Python Backend \[Experimental\]
+
+*NOTE*: This feature is subject to change and removal, and should not
+be used in production.
+
+Starting from 23.08, we are adding an experimental support for loading and
+serving PyTorch models directly via Python backend. The model can be provided
+within the triton server model repository without crafting the `model.py`, and a
```
tanmayv25, line 36: This line reads weird. Isn't the pytorch model still a `model.py` file sitting in the model directory. Can you improve the language for this statement?

src/resources/platform_handlers/pytorch_libtorch/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+By default, Triton will use the
+[PyTorch backend](https://github.com/triton-inference-server/pytorch_backend) to
+load and serve PyTorch models. In order to serve from Python backend,
+[model configuration](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_configuration.md)
+should explicitly provide the following settings:
+
+```
+backend: "python"
+platform: "pytorch_libtorch"
```
tanmayv25, line 67: Drop libtorch from the platform name. Libtorch is the C++ frontend to the pytorch. You can read more about it here: https://pytorch.org/cppdocs/installing.html#installing-c-distributions-of-pytorch
With this handler we are not using the C++ frontend anymore.

src/resources/platform_handlers/pytorch_libtorch/README.md
```diff
@@ -0,0 +1,86 @@
...
+pip install torch==2.0.1
+```
+Alternatively, a
+[Python Execution Environment](#using-custom-python-execution-environments)
+with the PyTorch dependency may be used.
+
+Following are few known limitations of this feature:
+- List of requests received in model [`execute`](../../../../README.md#execute)
+function are not ran in a single batch but one after the other.
+- Serving of arbitrary function(s) in the `model.py` file is not supported.
```
tanmayv25, line 86: Can you elaborate more? What function(s) are considered arbitrary? Can these functions be user-configurable with model config parameters?

src/resources/platform_handlers/pytorch_libtorch/model.py
tanmayv25: @kthui  The python backend side ticket is here: DLIS-2873

Till we add this enhnacement on the python backend, I would suggest we can modify the implementation here itself. We can merge the request buffers and obtain batched result with a single model execution. This can be followed by splitting the result into their multiple response objects.

src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,241 @@
...
+        for request in requests:
+            input_tensors = []
+            for io in self._inputs:
+                tensor = pb_utils.get_input_tensor_by_name(
+                    request, io["name"]
+                ).to_dlpack()
+                tensor = torch.from_dlpack(tensor).to(self._device)
+                input_tensors.append(tensor)
+
+            raw_output_tensors = self._model(*input_tensors)
```
tanmayv25, line 216: We should use inference_mode optimization for inference runs. Read more here: https://pytorch.org/docs/stable/generated/torch.inference_mode.html

src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,241 @@
...
+            "Initializing model instance for '" + self._model_name + "'"
+        )
+
+        self._model_config = json.loads(args["model_config"])
+        self._inputs = _parse_io_config(self._model_config["input"])
+        self._outputs = _parse_io_config(self._model_config["output"])
+
+        kind = args["model_instance_kind"]
+        device_id = args["model_instance_device_id"]
+        self._device = torch.device(_get_device_name(kind, device_id))
```
tanmayv25, line 151: What about using KIND_MODEL? We should ideally not set explicit devices when using KIND_MODEL.

src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,241 @@
...
+    return io
+
+
+def _get_device_name(model_instance_kind, model_instance_device_id):
+    if model_instance_kind == "GPU":
+        return "cuda:" + model_instance_device_id
+    return "cpu"
+
+
+def _enable_torch_compile(config):
```
tanmayv25, line 107: You should probably expose config parameters to set parallelism. 
https://pytorch.org/docs/stable/torch.html#parallelism
[set_num_threads](https://pytorch.org/docs/stable/generated/torch.set_num_threads.html#torch.set_num_threads)
[set_num_threads](https://pytorch.org/docs/stable/generated/torch.set_num_threads.html#torch.set_num_threads)
From our past experience the default value should be 1.

src/resources/platform_handlers/pytorch_libtorch/model.py
```diff
@@ -0,0 +1,241 @@
...
+                torch.load(data_path, map_location=self._device)
+            )
+        else:
+            self._logger.log_info(
+                "Model parameter file not found for '" + self._model_name + "'"
+            )
+        self._raw_model.to(self._device)
+        self._raw_model.eval()
+        if _enable_torch_compile(self._model_config):
+            self._model = torch.compile(self._raw_model)
```
tanmayv25, line 176: torch.compile has several arguments that can be provided. See here for the documentation: https://pytorch.org/docs/stable/generated/torch.compile.html#torch.compile
We should expose these via config parameters.

src/resources/platform_handlers/pytorch_libtorch/README.md
```diff
@@ -0,0 +1,86 @@
...
+same Python environment.
+
+```
+pip install torch==2.0.1
+```
+Alternatively, a
+[Python Execution Environment](#using-custom-python-execution-environments)
+with the PyTorch dependency may be used.
+
+Following are few known limitations of this feature:
```
tanmayv25, line 83: Can you add another limitation of how the model weights can not be shared across multiple instances on the same GPU?

src/resources/platform_handlers/pytorch_libtorch/README.md
kthui: I basically mean serving https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html#basic-usage functions are not supported. I updated the statement to be more descriptive.

src/resources/platform_handlers/pytorch_libtorch/README.md
kthui: Yes. Updated.

src/resources/platform_handlers/pytorch_libtorch/README.md
kthui: Yes. Updated.

src/resources/platform_handlers/pytorch_libtorch/README.md
kthui: Yes. Added the statement, but I am not sure why this limitation exists.

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: Yes, `KIND_MODEL` is now supported. If specified, the torch object will be left at its default device.

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: Yes. Added.

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: The ability to set those values is added.

From my testing, both values are defaulting to 32.  Should we leave the default value as it is, if the user is not providing the setting?

The values used for each model instance can be viewed by setting `log-verbose=1+` with/after [Enable setting torch parallelism parameters](https://github.com/triton-inference-server/python_backend/pull/282/commits/899dd7345b82d6f68b047648473d3b61b7c49841) commit.

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: Yes, added ability to supply all optional `torch.compile` parameters. Since one of the parameter is to disable `torch.compile` (no-op as described [on the doc](https://pytorch.org/docs/stable/generated/torch.compile.html#torch.compile)), the `DISABLE_TORCH_COMPILE` option is removed as part of this change to reduce code complexity and eliminate redundancy.

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: Yes. The ability to batch requests into a single tensor is added.

src/resources/platform_handlers/pytorch_libtorch/model.py
tanmayv25: I would advice setting the default value to be 1. Scaling the number of instances might quickly cause over-subscription of CPU otherwise.

src/resources/platform_handlers/pytorch_libtorch/README.md
tanmayv25: The model instances are running in separate process. Currently, there are no objects that can be held across the model instances. This is a python backend limitation.

src/resources/platform_handlers/pytorch/model.py
```diff
@@ -0,0 +1,329 @@
...
+        for input_tensors in requests_tensors:
+            output_tensors = self._model(*input_tensors)
+            if not isinstance(output_tensors, tuple) and not isinstance(
+                output_tensors, list
+            ):
+                output_tensors = [output_tensors]
+            responses_tensors.append(output_tensors)
+
+        if self._support_batching:
+            responses_tensors = self._unbatcher(responses_tensors[0], sections)
```
Tabrizian, line 307: "_unbatcher" -> "_scatter"

src/resources/platform_handlers/pytorch/model.py
```diff
@@ -0,0 +1,329 @@
...
+    batch_tensors = []
+    sections = []
+    for i in range(len(requests_tensors)):
+        request_tensors = requests_tensors[i]
+        for j in range(len(request_tensors)):
+            request_tensor = request_tensors[j]
+            if j < len(batch_tensors):
+                # add to existing tensor
+                tensor = batch_tensors[j]
+                torch.cat((tensor, request_tensor), 0)
```
Tabrizian, line 160: What if the request tensors are on different devices? Would this still work properly?

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: Yes. [Use default value 1 for torch parallelism](https://github.com/triton-inference-server/python_backend/pull/282/commits/26bd52a50cd78e554ad6cec3f08b8194cf438741)

src/resources/platform_handlers/pytorch/README.md
```diff
@@ -0,0 +1,88 @@
...
+model.
+
+The model repository structure should look like:
+
+```
+model_repository/
+`-- model_directory
+    |-- 1
+    |   |-- model.py
+    |   `-- model.py.pt
```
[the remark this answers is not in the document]
kthui: Yes. [Rename model.py.pt to model.pt](https://github.com/triton-inference-server/python_backend/pull/282/commits/c206e8e3f2b50ee84ebb808a1d3c751b523efb25)

src/resources/platform_handlers/pytorch/model.py
```diff
@@ -0,0 +1,329 @@
...
+
+# triton_python_backend_utils is available in every Triton Python model. You
+# need to use this module to create inference requests and responses. It also
+# contains some utility functions for extracting information from model_config
+# and converting Triton input/output types to numpy types.
+import triton_python_backend_utils as pb_utils
+
+
+def _get_model_path(config):
+    filenames = ["model.py", "model.pt", "model.pth"]
```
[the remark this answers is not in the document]
kthui: It is another file extension suggested by PyTorch.

> A common PyTorch convention is to save models using either a `.pt` or `.pth` file extension.

https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-state-dict-recommended

src/resources/platform_handlers/pytorch/model.py
```diff
@@ -0,0 +1,329 @@
...
+    for extension in data_path_extensions:
+        data_path = model_path + extension
+        if os.path.exists(data_path):
+            return data_path
+    # data file not provided
+    return ""
+
+
+def _is_py_class_model(model_path):
+    return model_path[-3:] == ".py"
```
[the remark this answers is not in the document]
kthui: I think this is ok, because Python will internally handle out of bound, for example:
```
Python 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> model_path = ""
>>> model_path[-3:]
''
>>> model_path = "/repo/model/1/model.py"
>>> model_path[-3:]
'.py'
>>>
```

src/resources/platform_handlers/pytorch/model.py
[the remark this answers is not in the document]
vadimkantorov: Despite the fact that `*.pth` extension is popular, this file extension is actually used by python interpreter to search for library paths, more details in https://github.com/pytorch/pytorch/issues/14864, which sometimes lead to python trying to read the text from the huge `*.pth` which may be slow and unpredictable. So IMO best would be to not encourage the users use it :( 

Also, this PR would be nice to be considered for a more streamlined PyTorch experience

src/resources/platform_handlers/pytorch/model.py
kthui: I guess it will not work, but the device info is embedded into each torch tensor, so if it does not work, the error will be internal to the torch framework.

To address the issue, the user will have to specify the model instance kind, which the script will ensure both the model and tensors are on the same device as described by the model instance kind.

kthui: In other words, this can only happen if the `KIND_MODEL` is used.

src/resources/platform_handlers/pytorch/model.py
```diff
@@ -0,0 +1,329 @@
...
+                ).to_dlpack()
+                tensor = _torch_object_to_device(
+                    torch.from_dlpack(tensor), self._device
+                )
+                tensors.append(tensor)
+            requests_tensors.append(tensors)
+
+        sections = None
+        if self._support_batching:
+            requests_tensors, sections = self._batcher(requests_tensors)
```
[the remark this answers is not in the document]
kthui: [Rename to gather and scatter](https://github.com/triton-inference-server/python_backend/pull/282/commits/81626bee69039cff253e92b3fa786e3aa8d60ef4)

src/resources/platform_handlers/pytorch/model.py
kthui: [Rename to gather and scatter](https://github.com/triton-inference-server/python_backend/pull/282/commits/81626bee69039cff253e92b3fa786e3aa8d60ef4)

src/resources/platform_handlers/pytorch/model.py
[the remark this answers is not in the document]
vadimkantorov: I think for this case the most clear and reliable is to use `os.path.splitext(model_path)[-1] == '.py'`

src/resources/platform_handlers/pytorch/model.py
```diff
@@ -0,0 +1,329 @@
...
+        self._device = _get_device(
+            args["model_instance_kind"], args["model_instance_device_id"]
+        )
+
+        params = _get_torch_compile_params(self._model_config)
+        self._logger.log_verbose(
+            "'torch.compile' optional parameter(s) " + for_model + ": " + str(params)
+        )
+        if self._support_batching:
+            self._batcher = torch.compile(_batch_torch_tensors, **params)
```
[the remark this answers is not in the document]
kthui: Yes. Added "... make sure the [PyTorch 2.0+ pip package](https://pypi.org/project/torch/2.0.1/) is available ..." statement.

src/resources/platform_handlers/pytorch/model.py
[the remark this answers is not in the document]
kthui: `.pth` extensions are removed from defaults.

src/resources/platform_handlers/pytorch/README.md
```diff
@@ -0,0 +1,88 @@
...
+Alternatively, a
+[Python Execution Environment](#using-custom-python-execution-environments)
+with the PyTorch dependency may be used.
+
+Following are few known limitations of this feature:
+- Python functions optimizable by `torch.compile` may not be served directly in
+the `model.py` file, they need to be enclosed by a class extending the
+[`torch.nn.Module`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module).
+- Model weights cannot be shared across multiple instances on the same GPU
+device.
```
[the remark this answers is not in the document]
kthui: Yes. Added customization section on the documentation.

src/resources/platform_handlers/pytorch_libtorch/model.py
vadimkantorov: and in general letting the user provide environment variables might be nice for pytorch, as sometimes need to pass extra parallelization configs to openmp, and not always pytorch allows to control them via calling python methods...

src/resources/platform_handlers/pytorch/model.py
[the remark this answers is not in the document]
kthui: I am not 100% convinced this is better, because the benefit from the `os.path.splitext()` is to allow the extension to vary in its number of characters (i.e. `.py` vs. `.python`), but ultimately the extracted extension is compared against a fixed string `.py`. If the extension does not have a length of 3 characters, the comparison will fail anyway (i.e. `'.python'[-3:] => 'hon' != '.py'`). I think the most important thing is to have the check logic separated from the rest of the code (`_is_py_class_model()` in this case) at this time, so it could be easily revised in the future if different extensions are to be added (or even stop relying on file extension).

src/resources/platform_handlers/pytorch/model.py
Tabrizian: I mean let's say a tensor from request A is stored on GPU and tensor from request B is on CPU, is torch.cat able to concatenate them into a single tensor? I believe torch was sensitive for all the tensors to have the same type so just wanted to make sure.

src/resources/platform_handlers/pytorch/README.md
```diff
@@ -0,0 +1,130 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# Serving PyTorch models using Python Backend \[Experimental\]
```
Tabrizian, line 29: Could you add a link to this document in the main readme?

src/resources/platform_handlers/pytorch/model.py
kthui: > I mean let's say a tensor from request A is stored on GPU and tensor from request B is on CPU, is torch.cat able to concatenate them into a single tensor?

I do not think this will work.

If the user uses `KIND_GPU` or `KIND_CPU`, the input tensors will be moved to either CPU or GPU before `gather` is called (before `torch.cat()` is called), so all tensors will be on the same device. However, if the user uses `KIND_MODEL`, the tensors will not be moved, which could cause the issue. To be safe, the user should avoid using `KIND_MODEL`.

src/resources/platform_handlers/pytorch/README.md
kthui: Yes. https://github.com/triton-inference-server/server/pull/6139

Tabrizian: Sorry for not being clear, I meant in the main Python backend readme.

src/resources/platform_handlers/pytorch/model.py
kthui: I gave it a try: Place model parameters on GPU, and place tensors on CPU. It does not work. The model parameters and tensors must be on the same device.

How about we do it this way, for `KIND_MODEL`:

If the model has at least one parameter, the first parameter of the model `next(model.parameters()).device` will be used to determine the model device. The entire model (i.e. any subsequence parameters not on the same device as the first parameter) will be moved to the model device determined. Then, any input tensors will be moved to the model device.

If no parameter is specified for the model, input tensors will be unified at `cpu`.

src/resources/platform_handlers/pytorch/README.md
kthui: [Link docs from primary Python docs](https://github.com/triton-inference-server/python_backend/pull/282/commits/5fb6dd47478f34d86f971829a0a3b5d48ae56360)

src/resources/platform_handlers/pytorch/model.py
kthui: [Update kind_model behavior](https://github.com/triton-inference-server/python_backend/pull/282/commits/f0f03b9c10e112fb332c2c3507321ba041199497)

kthui: [Simplify kind_model logic](https://github.com/triton-inference-server/python_backend/pull/282/commits/9cf3041720d9f700465ec290e6929ecd8c04061d)

Tabrizian (approved):

tanmayv25 (approved):

oandreeva-nv (approved):

src/resources/platform_handlers/pytorch_libtorch/model.py
kthui: Interesting... will keep an eye out for possible issues, as we are only trying to expose some PyTorch configs at this time.

PR #291, opened 2023-08-18
Reduce the default required shm size to 1MB

Tabrizian:
There has been a lot of requests to reduce the default share memory size to avoid running into issues when using Docker.
Since Python backend automatically adjusts the shm size, I think it should be safe to decrease the default to 1MB.

### DIFF
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -1901,8 +1901,8 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
 
   std::unique_ptr<BackendState> backend_state(new BackendState());
   triton::common::TritonJson::Value cmdline;
-  backend_state->shm_default_byte_size = 1 * 1024 * 1024;  // 1 MBs
-  backend_state->shm_growth_byte_size = 1 * 1024 * 1024;   // 1 MBs
+  backend_state->shm_default_byte_size = 1 * 1024 * 1024;  // 1 MB
+  backend_state->shm_growth_byte_size = 1 * 1024 * 1024;   // 1 MB
   backend_state->stub_timeout_seconds = 30;
   backend_state->shm_message_queue_size = 1000;
   backend_state->number_of_instance_inits = 0;
@@ -1936,7 +1936,7 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
       RETURN_IF_ERROR(shm_default_size.AsString(&shm_default_byte_size));
       try {
         backend_state->shm_default_byte_size = std::stol(shm_default_byte_size);
-        // Shared memory default byte size can't be less than 1 MBs.
+        // Shared memory default byte size can't be less than 1 MB.
         if (backend_state->shm_default_byte_size < 1 * 1024 * 1024) {
           return TRITONSERVER_ErrorNew(
               TRITONSERVER_ERROR_INVALID_ARG,
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -1901,8 +1901,8 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
 
   std::unique_ptr<BackendState> backend_state(new BackendState());
   triton::common::TritonJson::Value cmdline;
-  backend_state->shm_default_byte_size = 64 * 1024 * 1024;  // 64 MBs
-  backend_state->shm_growth_byte_size = 64 * 1024 * 1024;   // 64 MBs
+  backend_state->shm_default_byte_size = 1 * 1024 * 1024;  // 1 MBs
+  backend_state->shm_growth_byte_size = 1 * 1024 * 1024;   // 1 MBs
   backend_state->stub_timeout_seconds = 30;
   backend_state->shm_message_queue_size = 1000;
   backend_state->number_of_instance_inits = 0;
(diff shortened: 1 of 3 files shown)


src/python_be.cc
```diff
@@ -1901,8 +1901,8 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
 
   std::unique_ptr<BackendState> backend_state(new BackendState());
   triton::common::TritonJson::Value cmdline;
-  backend_state->shm_default_byte_size = 64 * 1024 * 1024;  // 64 MBs
-  backend_state->shm_growth_byte_size = 64 * 1024 * 1024;   // 64 MBs
+  backend_state->shm_default_byte_size = 1 * 1024 * 1024;  // 1 MBs
+  backend_state->shm_growth_byte_size = 1 * 1024 * 1024;   // 1 MBs
```
rmccorm4, line 1905: Out of curiosity, where did the 64MB minimum come from? And how early would the backend/model identify it needs to grow the size? Are there cases where it realizes it needs to grow by say 100MB, but will still only attempt to grow 1MB at a time for 100 times?

src/python_be.cc
```diff
@@ -1901,8 +1901,8 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
 
   std::unique_ptr<BackendState> backend_state(new BackendState());
   triton::common::TritonJson::Value cmdline;
-  backend_state->shm_default_byte_size = 64 * 1024 * 1024;  // 64 MBs
-  backend_state->shm_growth_byte_size = 64 * 1024 * 1024;   // 64 MBs
+  backend_state->shm_default_byte_size = 1 * 1024 * 1024;  // 1 MBs
```
krishung5, line 1904: nit: 1MB

src/python_be.cc
```diff
@@ -1936,8 +1936,8 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
       RETURN_IF_ERROR(shm_default_size.AsString(&shm_default_byte_size));
       try {
         backend_state->shm_default_byte_size = std::stol(shm_default_byte_size);
-        // Shared memory default byte size can't be less than 4 MBs.
-        if (backend_state->shm_default_byte_size < 4 * 1024 * 1024) {
+        // Shared memory default byte size can't be less than 1 MBs.
```
krishung5, line 1939: nit: 1MB

src/python_be.cc
Tabrizian: > Out of curiosity, where did the 64MB minimum come from?

I believe it was originally suggested by one of the former team members.

> How early would the backend/model identify it needs to grow the size?

It would grow the region whenever it runs out of shared memory. https://github.com/triton-inference-server/python_backend/blob/0f2ce85f789dddffec51aba6c483f759675f66f1/src/shm_manager.h#L96-L99

> Are there cases where it realizes it needs to grow by say 100MB, but will still only attempt to grow 1MB at a time for 100 times?

There might be cases like this but I think it won't be common. If the user has a large 100MB tensor the region would grow to 100MB in one pass. I think the only case that this might affect the perf is if they gradually in small increments increase the shared memory usage. For those cases, the options are still exposed if they want to change the default.

> What is the [default constructed SHM Manager](https://github.com/triton-inference-server/python_backend/blob/0f2ce85f789dddffec51aba6c483f759675f66f1/src/shm_manager.cc#L97-L102) with default growth size of 1 KB used for? It looks like it's used for SHM Monitor testing, but can you summarize the reasoning / why different growth value?

I think this is only used in the shared memory monitor: https://github.com/triton-inference-server/python_backend/blob/0f2ce85f789dddffec51aba6c483f759675f66f1/src/shm_monitor/shm_monitor.cc so it is not used in the production Python backend code.

krishung5 (changes requested):
Great improvement! Minor comments: I think we need to update the comment here:
https://github.com/triton-inference-server/python_backend/blob/0f2ce85f789dddffec51aba6c483f759675f66f1/src/shm_manager.cc#L79

and also in our qa testing:
https://github.com/triton-inference-server/server/blob/7fd85163ea87508955276b31c52bb7bf52e36324/qa/L0_infer/test.sh#L82

krishung5 (approved):

dyastremsky (approved):

src/shm_manager.cc
```diff
@@ -76,7 +76,7 @@ SharedMemoryManager::SharedMemoryManager(
          "' to requested size (" + std::to_string(shm_size) +
          " bytes). If you are running Triton inside docker, use '--shm-size' "
          "flag to control the shared memory region size. Each Python backend "
-         "model instance requires at least 64MBs of shared memory. Error: " +
+         "model instance requires at least 1 MB of shared memory. Error: " +
```
rmccorm4, line 79: nit if you feel like changing this:

We can just define a constant and use it everywhere so we don't have to remember to fix each spot:
```cpp
// constants.h
constexpr uint64_t DEFAULT_SHM_SIZE = 1*1024*1024; // 1 MB
constexpr uint64_t DEFAULT_SHM_GROWTH_SIZE = 1*1024*1024; // 1 MB

...
// source
backend_state->shm_default_byte_size = DEFAULT_SHM_SIZE;
...
        if (backend_state->shm_default_byte_size < DEFAULT_SHM_SIZE) {
...
error_message = "... at least " + std::to_string(DEFAULT_SHM_SIZE) + " of shared memory..."
```

Tabrizian: It is already only in two places so I think it should be fine.

shixianc:
I'm curious in which release would this change in effect?
I think this would be a problem for user who're dealing with larger batch size, right?
Is there any instruction on which value we should choose given input tensor size?

PR #293, opened 2023-08-31
Optimize GPU tensor support for Python backend

krishung5:
Currently, the CUDA IPC calls dominate the time for transferring GPU tensors between processes. Specifically, the functions cudaIpcOpenMemHandle and cudaIpcCloseMemHandle are heavily used. These functions are necessary because the allocated buffers are not in the same pool. So, we need to call these functions to open an interprocess memory handle exported from another process and get a device pointer that can be used in the local process.

This PR makes use of Triton's CUDA shared memory pool for GPU tensor transfers. The parent process will get the base address of the CUDA pool and share it with the stub. Here's how the data transfer process works:
- **Data transfer from parent to stub**
When the parent process wants to send a tensor to the stub using the pool, it stores the data, calculates the offset, and shares this offset with the stub. The stub process then uses this offset to retrieve the data.
- **Data transfer from stub to parent**
Because only the parent process can interact with the CUDA pool memory allocation, the stub first notifies the parent about the byte size of the data using an IPC message. Then, the parent pre-allocates a buffer from the memory pool and communicates the calculated offset details back to the stub. Afterward, the stub fills the buffer with the tensor data and notifies the parent once the task is done.

Testing: https://github.com/triton-inference-server/server/pull/6276


Tabrizian:
Discussed with @krishung5 offline. Looks like there is one additional data copy that is introduced as a part of this change that is affecting the single model latency and throughput. @krishung5 is working on removing that extra copy and gather profiling numbers again.

src/python_be.cc
```diff
@@ -1542,11 +1593,20 @@ ModelInstanceState::ProcessRequests(
...
           cuda_copy |= cuda_used;
+        } else if (
+            pb_memory->MemoryType() == TRITONSERVER_MEMORY_GPU &&
+            pb_memory->UseCudaSharedPool()) {
+#ifdef TRITON_ENABLE_GPU
+          // Copy the data from the CUDA shared memory pool to the output buffer
+          // provided by Triton
+          pb_memory->WriteBackGPUOutput(Stub()->ShmPool(), stream_);
+          write_back = true;
+#endif  // TRITON_ENABLE_GPU
```
Tabrizian, line 1604: q: why is the data copy happening here instead of the stub process?

src/python_be.cc
```diff
@@ -591,13 +628,24 @@ ModelInstanceState::GetInputTensor(
           const_cast<void*>(dev_ptr), input_byte_size,
           nullptr /* DLManagedTensor */);
 
+      if (use_cuda_shared_pool) {
+        input_tensor->SetMemory(std::move(
+            PbMemory::Create(Stub()->ShmPool(), std::move(lbackend_memory))));
+      }
```
Tabrizian, line 634: If use backend memory we can remove the condition here.

src/pb_memory.h
```diff
@@ -57,9 +62,11 @@ class PbMemory {
   static std::unique_ptr<PbMemory> Create(
       std::unique_ptr<SharedMemoryManager>& shm_pool,
       TRITONSERVER_MemoryType memory_type, int64_t memory_type_id,
-      uint64_t byte_size, char* data, bool copy_gpu = true);
+      uint64_t byte_size, char* data, bool copy_gpu = true,
+      bool write_back_data = false, bool copy_data = true);
```
Tabrizian, line 66: why do we need an additional argument for `copy_data`. It looks like it is pretty much similar to `copy_gpu` argument.

To me it looks like `write_back_data` doesn't belong to this class and should be handled outside `PbMemory`.

src/python_be.cc
krishung5: This is after the stub process copies the GPU tensors to the parent-provided buffer. We need to copy the result back to Triton provided buffer so that the output is properly passed back to Triton.

src/pb_memory.h
krishung5: > why do we need an additional argument for copy_data. It looks like it is pretty much similar to copy_gpu argument.

For most of the cases, when creating a new `PbMemory`, we need to copy the data from the original buffer to the newly created buffer using `BackendMemory`. The `copy_data` argument is to handle the case where we just want to prepare the buffer for GPU output and ask the stub process to fill in the data later. Please see here: https://github.com/triton-inference-server/python_backend/pull/293/files#diff-32938d76ce399e799a75fbec5050c99c8546fcc08fb75ab526ea331342731932R310. It is possible that `copy_data` and `copy_gpu` is not aligned, hence using a separate argument here.

> To me it looks like write_back_data doesn't belong to this class and should be handled outside PbMemory. We can discuss more offline.

Makes sense to me. I went ahead and removed `write_back_data` from the `PbMemory` class and handle the data copy outside: https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR1612.

src/stub_launcher.h
```diff
@@ -151,6 +151,11 @@ class StubLauncher {
   TRITONSERVER_Error* ReceiveMessageFromStub(
       bi::managed_external_buffer::handle_t& message);
 
+#ifdef TRITON_ENABLE_GPU
+  // Share CUDA memory pool with stub process
+  void ShareCUDAMemoryPool(TRITONBACKEND_MemoryManager* triton_mem_manager);
```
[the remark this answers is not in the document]
krishung5: Switched to use the [`parent_to_stub_mq_`](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-f2bb64cbc9d4dc5d938ce965813b23178f53732e02e8c3a2afa707e287e92e57R1085) instead of the original message queue to enable flexible sharing of the CUDA memory pool with the stub for all cases:
- [BLS request has GPU input](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR698)
- [BLS response has GPU output](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR1683)
- [Model has GPU input](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR522)
- [Model has GPU output](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR1552), [decoupled](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR1169)

src/pb_memory.cc
```diff
@@ -193,8 +223,72 @@ PbMemory::FillShmData(
         THROW_IF_CUDA_ERROR(cudaIpcGetMemHandle(
             reinterpret_cast<cudaIpcMemHandle_t*>(memory_data_shm), data));
       }
+#ifndef TRITON_PB_STUB
```
Tabrizian, line 226: I think PbMemory should be just responsible for transferring the memory to and from the stub process. It looks like PbMemory is currently doing more.

My mental model is that with this optimization if `cuda_pool_offset` is set we would create the pointer based on the start address of the memory pool and there shouldn't be any copies involved in the `PbMemory` class. Please feel free to correct my understanding.

src/pb_stub.cc
```diff
@@ -1225,6 +1172,142 @@ Stub::GetProxyStream(const int& device_id)
...
+  CUDAMemPoolMessage* cuda_pool_message_ptr = nullptr;
+  try {
+    AllocatedSharedMemory<CUDAMemPoolMessage> cuda_handle_shm =
+        shm_pool_->Load<CUDAMemPoolMessage>(ipc_message->Args());
+    cuda_pool_message_ptr = cuda_handle_shm.data_.get();
+
+    CUDAHandler& cuda_api = CUDAHandler::getInstance();
+    void* cuda_pool_address;
+    cuda_api.OpenCudaHandle(
+        device_id_, &cuda_pool_message_ptr->cuda_handle, &cuda_pool_address);
```
Tabrizian, line 1192: While each model instance is assigned to a single GPU currently it is possible for a model instance on GPU A to return tensors on GPU B so there might be scenarios where we need to open memory pool for GPU B on model instance assigned to GPU A. In the current logic, it looks like the stub would only open the memory pool that corresponds to the `device_id_` is that correct?

src/python_be.cc
Tabrizian: I thought if the memory buffer returned by Triton is not from GPU memory pool, we would revert back to the previous cudaIpc policy. Is this extra copy more performant than opening/closing cudaIpc handle?

krishung5: > Is this extra copy more performant than opening/closing cudaIpc handle?

Yes, from the nsight traces, it appears that doing the extra copy has better performance.
Originally, the time for the stub to open cudaIpc handle, copy output to the buffer, and close the cudaIpc handle, is about  550 microseconds.
 _(red: cuda copy; green: open/close cuda handle)_
original_screenshot

After the optimization, the time for the stub to copy output to the intermediate buffer, which is created by the parent from GPU memory pool (the bottom section in the trace), and the time for the parent to do an extra copy from that intermediate buffer to Triton-provided buffer (the upper section in the trace), is about 350 microseconds. 
_(red: cuda copy; green: cuda synchronization)_
new_screenshot

src/pb_stub.cc
krishung5: Fixed. Right now the stub can open cuda memory pool on different devices based on the device_id of the tensor.

krishung5:
Updated the functionality of `PbMemory` - when creating a `PbMemory` object, it only sets the `cuda_pool_offset` if the data is allocated from the cuda pool; when loading `PbMemory` from memory, it'll return a pointer based on the offset. No extra data copy/logic happens inside `PbMemory`.

Summarize the logic for GPU tensor transfer below in case it could be helpful for reviewing.

Different cases for the GPU tensor transfer:

**Inference**
- input 
  - [non-decoupled](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR560-R610) - Get the input buffer from input_collector. If the buffer is not using cuda pool, create a new `BackendMemory` from the pool and copy the input data from the original buffer to the new one.
  - [decoupled](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR624-R639) - Allocate the GPU memory using `BackendMemory`, and call `backend::ReadInputTensor` to read input tensor to the buffer.
- output
  - non-decoupled - In `InferResponse::Send` function, get the Triton-provided output buffer. If the buffer is not using cuda pool,  [try to allocate a new buffer from the pool](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-32938d76ce399e799a75fbec5050c99c8546fcc08fb75ab526ea331342731932R290-R320), and add the new buffer to `gpu_buffer_helper`. Will need to [copy the output tensor back](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR1664-R1677) to the Triton-provided buffer once the stub fills in the buffer.
  - decoupled - Same as the non-decoupled case. The final copy happens [here](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR1242-R1254).

**BLS**
- input - No differences between decoupled and non-decoupled cases. [Allocate memory using `BackendMemory`](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR732-R746) and add the buffer to `gpu_buffer_helper`.
- output - The request_executor now uses [`BackendMemory`](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-15082b362416a681b048d30a39e09ba97f5838719c97ded622b9023a5f4057b0R250-R264) to allocate buffer for BLS output. Both non-decoupled and decoupled cases call `ModelInstanceState::PrepareResponseHandle` to prepare the response. It's possible that the cuda memory pool hasn't been shared with the stub process at the time the BLS output is allocated during the callback, and there is no way to share the cuda pool with the stub since we are not passing the `StubLauncher` object to the `ResponseAlloc` callback (thought it would be more complicated if we do so, but open to any ideas!). Hence, update the cuda pool offset [here](https://github.com/triton-inference-server/python_backend/pull/293/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR1737-R1745) after the associated `PbMemory` is created.

src/infer_response.cc
```diff
@@ -286,6 +287,37 @@ InferResponse::Send(
...
+                        ->TritonMemoryManager()),
+                BackendMemory::AllocationType::GPU_POOL, actual_memory_type_id,
+                output_tensor->ByteSize(), &backend_memory));
+            lbackend_memory.reset(backend_memory);
+            buffer = lbackend_memory->MemoryPtr();
+          }
+        }
+        catch (const PythonBackendException& pb_exception) {
+          LOG_MESSAGE(
+              TRITONSERVER_LOG_WARN,
```
Tabrizian, line 313: Is it possible to print this warning only once?

src/python_be.cc
```diff
@@ -1213,36 +1244,49 @@ ModelInstanceState::ResponseSendDecoupled(
...
+            THROW_IF_TRITON_ERROR(CopyBuffer(
+                "Failed to copy the CPU output tensor to buffer.",
+                TRITONSERVER_MEMORY_CPU, 0, TRITONSERVER_MEMORY_CPU, 0,
+                pb_memory->ByteSize(), pb_memory->DataPtr(), pointer,
+                CudaStream(), &cuda_used));
+            cuda_copy |= cuda_used;
+          } else if (
+              (pb_memory->MemoryType() == TRITONSERVER_MEMORY_GPU) &&
+              pb_memory->UseCUDASharedPool() &&
+              (pb_memory->DataPtr() != pointer)) {
```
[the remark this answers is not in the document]
krishung5: That's correct. If the pointer is the same, it means that the output transfer is already using the correct buffer(Triton-provided buffer).

src/python_be.cc
```diff
@@ -557,6 +557,57 @@ ModelInstanceState::GetInputTensor(
...
           nullptr /* DLManagedTensor */);
 
+      // When the the CUDA shared memory pool is shared between parent and stub
+      // processes, check whether the buffer is already using the CUDA shared
+      // memory pool. If not, attempt to allocate memory from the pool and copy
+      // the input tensor to the newly allocated buffer so that we can use the
+      // CUDA pool for optimized data transfer.
+      std::unique_ptr<CUDAMemoryPoolManager>& cuda_pool =
+          Stub()->ShmPool()->GetCUDAMemoryPoolManager();
+      if (cuda_pool->UseCudaSharedPool(src_memory_type_id)) {
```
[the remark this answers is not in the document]
krishung5: On second thought, I think the `ProcessTensor` is already allocating the buffer using CUDA pool first, so I think this section here can be removed as it's doing duplicated things.

src/infer_response.cc
krishung5: Fixed.

src/memory_manager.cc
```diff
@@ -101,7 +107,14 @@ MemoryManager::QueueMonitorThread()
 
       // Call the release callback.
       it->second->ReleaseCallback()(it->second->MemoryId());
+      it->second.reset();
       records_.erase(it);
+      {
+        bi::scoped_lock<bi::interprocess_mutex> lock{
+            *(ipc_message->ResponseMutex())};
+        memory_release_message_ptr->waiting_on_stub = true;
+        ipc_message->ResponseCondition()->notify_all();
```
Tabrizian, line 116: Why the previous approach didn't work?

src/python_be.cc
```diff
@@ -662,6 +675,8 @@ ModelInstanceState::ExecuteBLSRequest(
         for (auto& input_tensor : infer_request->Inputs()) {
           if (!input_tensor->IsCPU()) {
 #ifdef TRITON_ENABLE_GPU
+            // Attempt to use the cuda shared memory pool for GPU tensor.
+            ShareCUDAMemoryPool(input_tensor->MemoryTypeId());
```
Tabrizian, line 679: I was wondering would it be possible to initially get all the memory pool addresses and make it part of the `IPCControl` struct. Then, whenever the stub wants to use a memory pool it would check whether it exists in the map of the memory pools it has. If not it would initialize it. This would remove the need for IPC transfer for every inference and improve performance.

Tabrizian (commented):
Great work. Few questions and it should be done.

src/memory_manager.cc
krishung5: Synced offline. Revert the changes for extra round of message transfer. The queue of MemoryManager is of pointer type.

src/python_be.cc
krishung5: Resolved offline.

Tabrizian (approved):

PR #294, opened 2023-09-11
Python Backend Windows Support

fpetrini15:
**Goal:** Support base python backend functionality for Windows.

_Note:_ This does not include support custom execution environments or GPU tensors. This base port also will not provide the full range of functionality supported in our Unix release, such as access to our model load/unload API.

Most of the code in this ticket rigidly separates Windows implementation from Unix implementation at the risk of being redundant. This is intentional as I think it will help us more quickly identify the root of any potential issues that arise, while insulating our robust Unix release. We can more tightly couple the implementations as we build our confidence with the platform.

Server changes: https://github.com/triton-inference-server/server/pull/6830


src/pb_preferred_memory.h
```diff
@@ -30,10 +30,10 @@ namespace triton { namespace backend { namespace python {
 
 class PreferredMemory {
  public:
-  enum MemoryType { GPU, CPU, DEFAULT };
+  enum MemoryType { kGPU, kCPU, kDefault };
```
rmccorm4, line 33: What's the significance of all the `CONSTANT` -> `kConstant` changes? Is it a windows requirement? Or style change?

fpetrini15: `ERROR` is defined as a macro when we include `windows.h`, so there is a naming collision. This would also be true for names such as `IN` or `OUT`. Probably for this reason, Google's style sheet says enum fields should be named like constants and not macros, so I went ahead changed how they are declared universally in the pybe.

CMakeLists.txt
```diff
@@ -129,7 +138,10 @@ if(${TRITON_ENABLE_NVTX})
 endif() # TRITON_ENABLE_NVTX
 
 find_package(ZLIB REQUIRED)
-find_package(Threads REQUIRED)
+
+if(NOT WIN32)
+  find_package(Threads REQUIRED)
```
Tabrizian, line 143: I wonder why `Threads` is not available on Windows?

src/python_be.cc
```diff
@@ -2293,7 +2308,7 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
   return nullptr;
 }
 
-TRITONSERVER_Error*
+TRITONBACKEND_ISPEC TRITONSERVER_Error*
```
Tabrizian, line 2311: What does `TRITONBACKEND_ISPEC` do?

src/stub_launcher.cc
```diff
@@ -195,6 +197,139 @@ StubLauncher::Setup()
...
+  launch_command = ss.str();
+
+  LOG_MESSAGE(
+      TRITONSERVER_LOG_INFO,
+      (std::string("Starting Python backend stub: ") + launch_command).c_str());
+
+  LPSTR launch_command_lpstr = const_cast<char*>(launch_command.c_str());
+  // Start the child process. Unlike fork(), the remainder of this
+  // function exists in the context of the parent, only.
+  if (!CreateProcess(
```
Tabrizian, line 259: Would be good to have a ticket for this to use `boost:interprocess`. It includes cross-platform functions for spawning a process, etc.

Tabrizian (approved):
Great work @fpetrini15 !

src/python_be.cc
fpetrini15: As I understand it, in Triton, we define TRITONBACKEND_ISPEC as a macro, confusingly, for both `__declspec(dllexport)` and `__declspec(dllimport)`, depending on which module is being compiled ([export](https://learn.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=msvc-170&source=recommendations)/[import](https://learn.microsoft.com/en-us/cpp/build/importing-into-an-application-using-declspec-dllimport?view=msvc-170) docs). These keywords are used to import/export functions, classes, data, etc., from a dll and we need it to import our backend definitions of the functions that core expects. From my understanding, when compiling core, TRITONBACKEND_ISPEC refers to dllimport and imports the functions defined by the backend dll. When compiling the backend, TRITONBACKEND_ISPEC refers to dllexport and exports the implementation of the functions that core expects to the backend dll ([see](https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonbackend.h#L37) the use of `_COMPILING_TRITONBACKEND`). This macro-defined keyword is necessary for proper functionality on Windows and is used to decorate these common API functions in our backends that support Windows (see [here](https://github.com/triton-inference-server/identity_backend/blob/main/src/identity.cc#L455), for example).

CC @nv-kmcgill53 for resolution of our discussion.

CMakeLists.txt
fpetrini15: I tested locally including Threads and there was no issue. This is probably an early stage removal that was never revisited. I'll run a pipeline to verify functionality.

src/stub_launcher.cc
fpetrini15: This is a great idea, I'll make a ticket!

src/stub_launcher.cc
```diff
@@ -600,6 +781,22 @@ StubLauncher::ReceiveMessageFromStub(
   return nullptr;  // success
 }
 
+void
+StubLauncher::WaitForStubProcess()
+{
+  std::cerr << __FILE__ << " " << __FUNCTION__ << " " << __LINE__ << '\n';
```
fpetrini15, line 787: Debug statements to be removed

krishung5 (approved):
LGTM!

src/python_be.cc
Tabrizian: Thanks for detailed response.

nv-kmcgill53: @fpetrini15 sums it up well. The only think I have to add is this other [MS doc page](https://learn.microsoft.com/en-us/cpp/cpp/dllexport-dllimport?view=msvc-170) which discusses this topic some more. One clarification to @fpetrini15's explanation is that it's not about exporting the "implementation" it's about exporting the "symbol name." `dllexport` will export the decorated name (name mangled name), omitting any decoration will cause the function to be `internal` and not visible to the using program, and `cdecl` will export the non-decorated name.  

Comparing this to `.so` files, all symbols are exported by default, so if you can load the `.so` file, you can call the function. I guess I had more to say than I thought....

PR #295, opened 2023-09-13
Fix returning error  when tracing is off

oandreeva-nv:
We should not use `RETURN_IF_ERROR`  for `TRITONBACKEND_RequestTrace` call, since when tracing is disabled, core side returns error.


rmccorm4 (approved):

shayanhoshyari:
Guys this error wasted 3 weeks of a mid-level engineer ...

PR #296, opened 2023-09-13
Add CUDA cache cleaning flag to python backend

leloykun:
I implemented the `ENABLE_CACHE_CLEANING` flag we have in the PyTorch backend similar to https://github.com/triton-inference-server/pytorch_backend/pull/61 . With ENABLE_CACHE_CLEANING set to true, CUDA cache gets cleared after each model execution. This will enable us to serve more models in one GPU.

This potentially solves https://github.com/triton-inference-server/server/issues/6070 and https://github.com/triton-inference-server/server/issues/6038


Tabrizian:
Thanks for your contribution but Python backend shouldn't have a dependency on PyTorch. You can clean the CUDA memory cache at the end of your model's execution in your `model.py`.

PR #302, opened 2023-09-25
Fix response iterator memory leak

Tabrizian:
It looks like that the response iterator keep_alive policy was incorrect. It was binding the lifetime of the return value to itself which resulted in the response iterator to never get destructed.


krishung5 (approved):
Great catch! Thanks for the fix.

rmccorm4 (approved):
LGTM locally on resolving the response-tensor-proportional leak, nice find!

PR #304, opened 2023-09-28
Add Python backend request cancellation

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/6364

Add `is_cancelled()` check API into Python model requests, and allow Python model to respond request cancelled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,6 +502,7 @@ Supported error codes:
 * `pb_utils.TritonError.UNAVAILABLE`
 * `pb_utils.TritonError.UNSUPPORTED`
 * `pb_utils.TritonError.ALREADY_EXISTS`
+* `pb_utils.TritonError.CANCELLED` (since 23.10)
 
 #### Decoupled mode
 
--- a/README.md
+++ b/README.md
@@ -576,9 +576,7 @@ object, use InferenceResponseSender.send() to send response with the
 error back to the user.
 
 Starting from 23.10, request cancellation can be checked directly on the
-`InferenceResponseSender` object using `response_sender.is_cancelled()`. If
-`response_sender.is_cancelled()` returned `True`, then no further steps are
-needed to be performed on this object.
+`InferenceResponseSender` object using `response_sender.is_cancelled()`.
 
 ##### Use Cases
 
--- a/README.md
+++ b/README.md
@@ -508,10 +508,10 @@ Supported error codes:
 #### Request Cancellation Handling
 
 One or more requests may be cancelled by the client during execution. Starting
-from 23.10, `request.is_cancelled()` returns up-to-date `True` or `False` on
-whether the request is cancelled. If a request is cancelled, the model may
-respond with any dummy object in place of the normal output tensors on the
-request. For example:
+from 23.10, `request.is_cancelled()` returns whether the request is cancelled.
+
+If a request is cancelled, the model may respond with any dummy object in place
+of the normal output tensors on the request. For example:
 
 ```python
 import triton_python_backend_utils as pb_utils
@@ -576,10 +576,9 @@ object, use InferenceResponseSender.send() to send response with the
 error back to the user.
 
 Starting from 23.10, request cancellation can be checked directly on the
-`InferenceResponseSender` object, for example `response_sender.is_cancelled()`,
-even after the request has gone out-of-scope. If
-`response_sender.is_cancelled()` returned `True`, the
-TRITONSERVER_RESPONSE_COMPLETE_FINAL flag is sent automatically.
+`InferenceResponseSender` object using `response_sender.is_cancelled()`. If
+`response_sender.is_cancelled()` returned `True`, then no further steps are
+needed to be performed on this object.
 
 ##### Use Cases
 
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ any C++ code.
     - [`execute`](#execute)
       - [Default Mode](#default-mode)
       - [Error Handling](#error-handling)
+      - [Request Cancellation](#request-cancellation)
       - [Decoupled mode](#decoupled-mode)
         - [Use Cases](#use-cases)
         - [Known Issues](#known-issues)
@@ -503,6 +504,37 @@ Supported error codes:
 * `pb_utils.TritonError.UNSUPPORTED`
 * `pb_utils.TritonError.ALREADY_EXISTS`
 * `pb_utils.TritonError.CANCELLED` (since 23.10)
+
+#### Request Cancellation
+
+One or more requests may be cancelled during execution, for example, cancelled
+by the user. Starting from 23.10, `request.is_cancelled()` returns up-to-date
+`True` or `False` on whether the request is cancelled. If a request is
+cancelled, the model should respond `pb_utils.TritonError.CANCELLED` in place of
+the normal output tensors on the request. For example:
+
+```python
+import triton_python_backend_utils as pb_utils
+
+class TritonPythonModel:
+    ...
+
+    def execute(self, requests):
+        responses = []
+
+        for request in requests:
+            if request.is_cancelled():
+                responses.append(pb_utils.InferenceResponse(
+                    error=pb_utils.TritonError("Message", pb_utils.TritonError.CANCELLED)))
+            else:
+                ...
+
+        return responses
+```
+
+Although checking for request cancellation is optional, it is recommended to
+check for cancellation at strategic request execution stages that can early
+terminate the execution in the event of its response is no longer needed.
 
 #### Decoupled mode
 
--- a/README.md
+++ b/README.md
@@ -575,7 +575,9 @@ object, use InferenceResponseSender.send() to send response with the
 error back to the user.
 
 Starting from 23.10, request cancellation can be checked directly on the
-`InferenceResponseSender` object using `response_sender.is_cancelled()`.
+`InferenceResponseSender` object using `response_sender.is_cancelled()`. Sending
+the TRITONSERVER_RESPONSE_COMPLETE_FINAL flag at the end of response is still
+needed even the request is cancelled.
 
 ##### Use Cases
 
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ any C++ code.
     - [`execute`](#execute)
       - [Default Mode](#default-mode)
       - [Error Handling](#error-handling)
-      - [Request Cancellation](#request-cancellation)
+      - [Request Cancellation Handling](#request-cancellation-handling)
       - [Decoupled mode](#decoupled-mode)
         - [Use Cases](#use-cases)
         - [Known Issues](#known-issues)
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -502,6 +503,38 @@ Supported error codes:
 * `pb_utils.TritonError.UNAVAILABLE`
 * `pb_utils.TritonError.UNSUPPORTED`
 * `pb_utils.TritonError.ALREADY_EXISTS`
+* `pb_utils.TritonError.CANCELLED` (since 23.10)
```
nnshah1, line 506: We will also need to update for the response factory a method for determining if the associated request is cancelled (To handle decoupled mode where requests are released before responses sent) - does that pattern hold for python backend?

kthui: If I am reading the code correctly, the scenario of returning cancelled on decoupled will work, because the `InferenceResponseSender` on decoupled takes the same `InferenceResponse` object that is used on non-decoupled scenario, and the cancelled status is set when the `InferenceResponse` object is constructed.

I will look into the cancelled query scenario on decoupled (and add a test case to show it works), because it is more worrying if the "requests are released before responses sent"

src/infer_request.cc
```diff
@@ -400,6 +400,23 @@ InferRequest::DeleteResponseFactory()
...
+bool
+InferRequest::IsCancelled()
+{
+  std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
+  if (!stub->StubToParentServiceActive()) {
+    LOG_ERROR << "Cannot communicate with parent service";
+    return false;
+  }
+  if (request_address_ == 0) {
+    LOG_ERROR << "Request address not provided (default initialized?)";
```
Tabrizian, line 412: Instead of logging the error you can throw an exception.

README.md
Tabrizian: Yeah, I think we probably need a `is_cancelled` method for `InferenceResponseSender` as well.

kthui: Added a test case on decoupled model. The logic on request check is to check for cancellation on response factory pointer first. If it is undefined (i.e. non-decoupled case), then the cancellation is checked on the request pointer. If both response factory and request are undefined (should not happen), then an exception is throw.

src/infer_request.cc
kthui: Sure. The logic is moved to "python_be.cc" after some changes on decoupled case.

README.md
Tabrizian: > @Tabrizian Can you elaborate more on why we want to check for cancellation on InferenceResponseSender?

It is not required to keep the inference request object alive. The use-case I had in mind is similar to the C++ backend APIs. In C Backend APIs, we have APIs for both `TRITONBACKEND_RequestIsCancelled` and `TRITONBACKEND_ResponseFactoryIsCancelled`. In Python backend response factory is similar to response sender.

As shown in the example [here](https://github.com/triton-inference-server/python_backend/blob/b136bf3fad77d5d467a6db5bd739be16eb9d9400/examples/decoupled/square_model.py#L201-L217) the request object may not be in scope when the request is being cancelled.

kthui: Thanks for clarifying! Will update.

kthui: Updated. The user is able to call either of the following:
```
request.is_cancelled()
request.get_response_sender().is_cancelled()  # does not depend on request after getting response sender
```

README.md
```diff
@@ -502,6 +503,38 @@ Supported error codes:
...
+
+#### Request Cancellation
+
+One or more requests may be cancelled during execution, for example, cancelled
```
tanmayv25, line 510: Can you improve this language? We should make clear distinction that this is client side request cancellation and not request being cancelled by the server.

How about changing `Request Cancellation` -> `Handling Client-Side Request Cancellation`?

README.md
```diff
@@ -502,6 +503,38 @@ Supported error codes:
...
+by the user. Starting from 23.10, `request.is_cancelled()` returns up-to-date
```
tanmayv25, line 511: Just a reminder, when reviewing release branch for 23.11, make sure 23.10 here is not replaced with 23.11.

README.md
```diff
@@ -502,6 +503,38 @@ Supported error codes:
...
+`True` or `False` on whether the request is cancelled. If a request is
+cancelled, the model should respond `pb_utils.TritonError.CANCELLED` in place of
```
tanmayv25, line 513: > If a request is cancelled, the model should respond `pb_utils.TritonError.CANCELLED`....

Is this a hard requirement? If checking for the cancelled state is optional, howcome responding in a certain way a requirement? Additionally, can't this be handled internally? As in if the request is cancelled then why the model developer should be required to create a cancelled response and return it?

README.md
```diff
@@ -502,6 +503,38 @@ Supported error codes:
...
+import triton_python_backend_utils as pb_utils
+
+class TritonPythonModel:
+    ...
+
+    def execute(self, requests):
+        responses = []
+
+        for request in requests:
+            if request.is_cancelled():
```
tanmayv25, line 526: I think a decoupled example will be a better choice of demonstrating the power of checking request cancellation. You should use `response_sender.is_cancelled()` in the same. 
May be add an example in the Decoupled mode section below.

README.md
kthui: How about changing to `Request Cancellation Handling`, and have
```
One or more requests may be cancelled by the client during execution.
```
to follow immediately on the first line of the paragraph? Just to have a shorter title.

README.md
kthui: The interface is simplified. On non-decoupled model, the model may return any object on the responses list if the request is cancelled. On decoupled model, the `TRITONSERVER_RESPONSE_COMPLETE_FINAL` flag is sent automatically upon checking for cancellation on a cancelled request, so the model does not need to do anything after checking for cancellation if cancelled.

README.md
kthui: I added
```
Starting from 23.10, request cancellation can be checked directly on the
`InferenceResponseSender` object, for example `response_sender.is_cancelled()`,
even after the request has gone out-of-scope. If
`response_sender.is_cancelled()` returned `True`, the
TRITONSERVER_RESPONSE_COMPLETE_FINAL flag is sent automatically.
```
under decoupled section. Do you think it is enough?

README.md
```diff
@@ -502,6 +503,37 @@ Supported error codes:
 * `pb_utils.TritonError.UNAVAILABLE`
 * `pb_utils.TritonError.UNSUPPORTED`
 * `pb_utils.TritonError.ALREADY_EXISTS`
+* `pb_utils.TritonError.CANCELLED` (since 23.10)
+
+#### Request Cancellation Handling
+
+One or more requests may be cancelled by the client during execution. Starting
+from 23.10, `request.is_cancelled()` returns up-to-date `True` or `False` on
```
tanmayv25, line 511: `request.is_cancelled()` returns whether the request is cancelled.

README.md
```diff
@@ -502,6 +503,37 @@ Supported error codes:
...
+whether the request is cancelled. If a request is cancelled, the model may
+respond with any dummy object in place of the normal output tensors on the
+request. For example:
```
tanmayv25, line 514: Add a line about the object will be ignored if the request was cancelled.

README.md
```diff
@@ -543,6 +575,12 @@ request. After setting errors for an pb_utils.InferenceResponse
 object, use InferenceResponseSender.send() to send response with the
 error back to the user.
 
+Starting from 23.10, request cancellation can be checked directly on the
+`InferenceResponseSender` object, for example `response_sender.is_cancelled()`,
```
tanmayv25, line 579: `InferenceResponseSender` object using `response_sender.is_cancelled()`.
If `response_sender.is_cancelled()` returned `True`, then no further steps
are needed to be performed on this object.

kthui: Updated: [Documentation wording updates](https://github.com/triton-inference-server/python_backend/pull/304/commits/3a43ef3a9d0550935ae9857f3c3d54e73987ab35)

README.md
kthui: Updated: [Documentation wording updates](https://github.com/triton-inference-server/python_backend/pull/304/commits/3a43ef3a9d0550935ae9857f3c3d54e73987ab35)

README.md
kthui: Updated: [Documentation wording updates](https://github.com/triton-inference-server/python_backend/pull/304/commits/3a43ef3a9d0550935ae9857f3c3d54e73987ab35)

tanmayv25 (approved):

README.md
```diff
@@ -543,6 +575,9 @@ request. After setting errors for an pb_utils.InferenceResponse
 object, use InferenceResponseSender.send() to send response with the
 error back to the user.
 
+Starting from 23.10, request cancellation can be checked directly on the
+`InferenceResponseSender` object using `response_sender.is_cancelled()`.
```
tanmayv25, line 579: We should mention that the user need to explicitly send a final flag even if the request is cancelled.

README.md
```diff
@@ -502,6 +503,37 @@ Supported error codes:
...
+
+class TritonPythonModel:
+    ...
+
+    def execute(self, requests):
+        responses = []
+
+        for request in requests:
+            if request.is_cancelled():
+                responses.append(None)
```
Tabrizian, line 527: I believe `append(None)` would not work. You would need to create an explicit response object (e.g., `InferenceResponse(error=TritonError(CANCELLED))`)

src/pb_stub.cc
```diff
@@ -771,10 +771,17 @@ Stub::ProcessRequests(RequestBatch* request_batch_shm_ptr)
...
+      // replace returned type with a cancelled response.
+      if (py_request_list[i].cast<InferRequest*>()->IsCancelledLastResponse()) {
+        responses[i] = std::make_shared<InferResponse>(
+            std::vector<std::shared_ptr<PbTensor>>{},
+            std::make_shared<PbError>("", TRITONSERVER_ERROR_CANCELLED));
+      }
       // Check the return type of execute function.
-      if (!py::isinstance<InferResponse>(response)) {
-        std::string str = py::str(response.get_type());
+      else if (!py::isinstance<InferResponse>(responses[i])) {
```
Tabrizian, line 783: Let's keep the logic here simple. We would return whatever response they have returned and shouldn't try to create cancelled requests automatically.

README.md
kthui: Sure. [Decoupled final flag docs update](https://github.com/triton-inference-server/python_backend/pull/304/commits/09e358adab39322dbf807d00a9c993f16c375edf)

src/pb_stub.cc
kthui: Yes. Reverted the logic change: [Rollback non-decoupled any response on cancel](https://github.com/triton-inference-server/python_backend/pull/304/commits/3de5922c48ece05b5d55862063e67ab3a22c97f5)

README.md
kthui: Updated, as the logic is reverted: [Rollback non-decoupled any response on cancel](https://github.com/triton-inference-server/python_backend/pull/304/commits/3de5922c48ece05b5d55862063e67ab3a22c97f5)

Tabrizian (approved):

PR #309, opened 2023-10-10
Restructure cleaning up of the futures in decoupled mode

Tabrizian:
Fixes https://github.com/triton-inference-server/server/issues/6270#issuecomment-1755827698

Huge shout out to @rmccorm4 for finding the root cause.

### DIFF
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -777,9 +777,7 @@ ModelInstanceState::DecoupledMessageQueueMonitor()
       std::packaged_task<void()> task([this, response_send_message] {
         ResponseSendDecoupled(response_send_message);
       });
-      std::future<void> future =
-          boost::asio::post(*thread_pool_, std::move(task));
-      futures_.emplace_back(std::move(future));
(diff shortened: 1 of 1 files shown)


rmccorm4:
Nice fix! Spoke offline, let's try to make the messages in these blocks also `unique_ptr` instead of `shared_ptr` if possible by moving the unique_ptr into the lambda captures, to reduce risk of reintroducing a leak due to long lived reference.

Tabrizian:
@rmccorm4 I tried using unique_ptr but for some reason it looks like it needs to be a `const unique_ptr` so it looks like it might need further digging to make it work with unique_ptr.

rmccorm4 (approved):

rmccorm4:
Can you add a ticket to look into it later? I think it would be good to reduce future risk.

krishung5 (approved):
Nice work on finding the root cause and fixing the leak!

Tabrizian:
@rmccorm4 Filed a ticket.

PR #313, opened 2023-10-19
Add parameters support to InferenceRequest

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/6452

Add support for BLS models to specify parameters as a dictionary to be passed to the next model. The key and value pairs on the dictionary require keys to be a string and values can be a string, integer or boolean.

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -76,6 +76,27 @@ SignalHandler(int signum)
   // Skip the SIGINT and SIGTERM
 }
 
+template <typename PYTYPE>
+PYTYPE
+PyDefaultArgumentToMutableType(const py::object& argument)
+{
+  // The default argument on Python functions always reference the same copy,
+  // meaning if the default argument is changed by the function, then it is
+  // changed for all subsequent calls to the function. Thus, default arguments
+  // should be limited to basic types (i.e. None). This helper function returns
+  // an empty expected type, if the argument is None (i.e. default initialized).
+  // If the argument is neither None nor expected type, an exception is thrown.
+  if (py::isinstance<py::none>(argument)) {
+    return PYTYPE();
+  }
+  if (py::isinstance<PYTYPE>(argument)) {
+    return argument;
+  }
+  throw PythonBackendException(
+      std::string("Expect ") + typeid(PYTYPE).name() + ", got " +
+      std::string(py::str(argument.get_type())));
+}
+
 void
 Stub::Instantiate(
     int64_t shm_growth_size, int64_t shm_default_size,
@@ -1464,7 +1485,10 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
                       const int64_t model_version, const uint32_t flags,
                       const int32_t timeout,
                       const PreferredMemory& preferred_memory,
-                      const InferenceTrace& trace, const py::dict& parameters) {
+                      const InferenceTrace& trace,
+                      const py::object& parameters_) {
+            py::dict parameters =
+                PyDefaultArgumentToMutableType<py::dict>(parameters_);
             std::set<std::string> requested_outputs;
             for (auto& requested_output_name : requested_output_names) {
               requested_outputs.emplace(requested_output_name);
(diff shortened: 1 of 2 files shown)


src/request_executor.cc
```diff
@@ -365,6 +365,36 @@ RequestExecutor::Infer(
...
+            irequest, key.c_str(), string_value.c_str()));
+      } else if (value.IsInt()) {
+        int64_t int_value = 0;
+        THROW_IF_TRITON_ERROR(value.AsInt(&int_value));
+        THROW_IF_TRITON_ERROR(TRITONSERVER_InferenceRequestSetIntParameter(
+            irequest, key.c_str(), int_value));
+      } else if (value.IsBool()) {
+        bool bool_value = false;
+        THROW_IF_TRITON_ERROR(value.AsBool(&bool_value));
+        THROW_IF_TRITON_ERROR(TRITONSERVER_InferenceRequestSetBoolParameter(
```
oandreeva-nv, line 391: Basic question: what was the reason of not supporting floats?

kthui: It is because the core only has APIs for string, int and bool.
https://github.com/triton-inference-server/core/blob/main/src/tritonserver.cc#L1974-L2000

kthui: I think the user can always use a string to represent floats.

oandreeva-nv: > It is because the core only has APIs for string, int and bool.
I saw that, hoped there might be more details behind this choice
> I think the user can always use a string to represent floats.
agree, if they need a WAR, this might be a potential solution

kthui: >>> Basic question: what was the reason of not supporting floats?

>> It is because the core only has APIs for string, int and bool.
https://github.com/triton-inference-server/core/blob/main/src/tritonserver.cc#L1974-L2000

> I saw that, hoped there might be more details behind this choice

I don't know more details behind it. Maybe @tanmayv25 or @Tabrizian might know more details on the data types behind parameters?

src/pb_stub.cc
```diff
@@ -1485,7 +1502,8 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
           py::arg("flags").none(false) = 0, py::arg("timeout").none(false) = 0,
           py::arg("preferred_memory").none(false) =
               PreferredMemory(PreferredMemory::DEFAULT, 0),
-          py::arg("trace").none(false) = InferenceTrace())
+          py::arg("trace").none(false) = InferenceTrace(),
+          py::arg("parameters").none(false) = py::dict())
```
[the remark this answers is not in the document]
kthui: Thanks for pointing this out. From the [accepted answer](https://stackoverflow.com/a/26320917) on the post, the issue only exists if the default empty dictionary is modified, but the receiving function defines the parameter as [`const py::dict& parameters`](https://github.com/triton-inference-server/python_backend/pull/313/files/4d1bcc67f444f575ca129456e2a1294a0f30d233#diff-f2bb64cbc9d4dc5d938ce965813b23178f53732e02e8c3a2afa707e287e92e57R1467), meaning the default empty dictionary cannot be modified and we cannot run into this issue.
Unless there are other issues with using empty dictionary as initializer or it is a common practice to use `none`, I would be hesitate to make it more complicated than it needs to.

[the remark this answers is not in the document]
Tabrizian: The issue is that in the future someone may make some changes to modify the parameters before passing them to the shared memory and not being aware of this issue. Using `None` would avoid running into these issues in the future.

src/request_executor.cc
Tabrizian: I think it was just not requested. If someone asks for it we can add support for this.

Tabrizian (commented):
Except what Kris pointed out, it looks good to me.

src/request_executor.cc
oandreeva-nv: CC @fpetrini15 , I believe this PR related to your ticket: https://jirasw.nvidia.com/browse/DLIS-5626

src/pb_stub.cc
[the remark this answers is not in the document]
kthui: Adjusted the argument "parameters" to allow safe modification in the future, without risking corruption of the default argument.

Tabrizian (approved):
Nice helper function!

PR #319, opened 2023-11-02
Add support for request rescheduling

krishung5:
Testing: https://github.com/triton-inference-server/python_backend/pull/319


src/infer_request.cc
```diff
@@ -175,6 +176,20 @@ InferRequest::Trace()
   return trace_;
 }
 
+uint32_t
+InferRequest::ReleaseFlags()
+{
+  request_release_flags_ = infer_request_shm_ptr_->request_release_flags;
```
nnshah1, line 182: Question for my learning: why do we use a temporary variable here and in set release flags?

Is that needed because of shared memory?

src/resources/triton_python_backend_utils.py
```diff
@@ -542,3 +542,4 @@ def add_output(self, output):
 TRITONSERVER_REQUEST_FLAG_SEQUENCE_START = 1
 TRITONSERVER_REQUEST_FLAG_SEQUENCE_END = 2
 TRITONSERVER_RESPONSE_COMPLETE_FINAL = 1
+TRITONSERVER_REQUEST_RELEASE_RESCHEDULE = 2
```
nnshah1, line 545: does it make sense to add release_all to be complete?

krishung5: Didn't expose the `release_all` flag since it's the default behavior, but I think we can add it just to make it complete.

src/infer_request.h
```diff
@@ -87,7 +88,8 @@ class InferRequest {
       const intptr_t request_address = 0,
       const PreferredMemory& preferred_memory =
           PreferredMemory(PreferredMemory::DEFAULT, 0),
-      const InferenceTrace& trace = InferenceTrace());
+      const InferenceTrace& trace = InferenceTrace(),
+      const uint32_t& request_release_flags = TRITONSERVER_REQUEST_RELEASE_ALL);
```
GuanLuo, line 92: Will a different flag be passed when creating a request object?

src/pb_stub.cc
```diff
@@ -793,26 +793,39 @@ Stub::ProcessRequests(RequestBatch* request_batch_shm_ptr)
...
       // Check the return type of execute function.
-      if (!py::isinstance<InferResponse>(response)) {
-        std::string str = py::str(response.get_type());
-        throw PythonBackendException(
-            std::string("Expected an 'InferenceResponse' object in the execute "
-                        "function return list, found type '") +
-            str + "'.");
+      InferRequest* infer_request = py_request_list[i].cast<InferRequest*>();
+      if (infer_request->ReleaseFlags() ==
+          TRITONSERVER_REQUEST_RELEASE_RESCHEDULE) {
```
GuanLuo, line 801: I don't think the Python backend should enforce how to send the response, here it enforces the response must be sent with a request release ALL flag.

There is legacy implementation that Python backend have restriction on the response behavior depending on whether the model is marked decoupled. But in terms of the backend API usage, it is completely fine for the user model to send one response on reschedule and send response end flag alone on executing the rescheduled request (at the end one and only one response is sent).

krishung5: For non-decoupled model, there is only one response per request. I think there are two possible cases:

1.  **First iteration**: request reschedule flag + no response returned(append `None` to the return list of responses)
**Final iteration**: response complete flag + response
2. **First iteration**: rescheduled flag + response
**Second iteration(optional)**: reschedule flag + no response returned
**Final iteration**: response complete flag + no response returned

The above cases should be valid for backend API usage. However, for Python backend, the `response_sender` is only exposed to users for decoupled model, which means currently it's not possible for case 2 - sending response and response complete flag in different iterations. I think it makes sense to support case 2, and we can modify the Python backend to also have a monitor thread for non-decoupled models so that case 2 can be satisfied.
@Tabrizian Will there be any issues if we expose `response_sender` to non-decoupled models to fulfill the use case?
cc @nnshah1

src/resources/triton_python_backend_utils.py
krishung5: Added release_all flag.

src/infer_request.h
krishung5: Good catch! No, there won't be any. Fixed the constructor.

README.md
```diff
@@ -623,6 +624,101 @@ for more details on how to host a decoupled model.
...
+    def execute(self, requests):
+        responses = []
+
+        for request in requests:
+            # Explicitly reschedule the first request
+            if self.idx == 0:
+                request.set_release_flags(
+                    pb_utils.TRITONSERVER_REQUEST_RELEASE_RESCHEDULE
+                )
+                responses.append(None)
```
Tabrizian, line 662: Why does it have to return None for rescheduled request?

src/pb_stub.cc
Tabrizian: @krishung5 could you please elaborate why the 1-1 mapping is limiting here? What is the problem that we're trying to solve. If the user wants to send less than one response, they can use decoupled mode.

README.md
```diff
@@ -623,6 +624,101 @@ for more details on how to host a decoupled model.
...
+                self.idx += 1
+            else:
+                responses.append(inference_response)
+
+        return responses
+```
```
Tabrizian, line 668: It would be good to have a paragraph explaining why the user needs to return `None` for rescheduled request (i.e., the rescheduled request is the same request and there can be only one response with final flag with every request).

README.md
```diff
@@ -623,6 +624,101 @@ for more details on how to host a decoupled model.
...
+                    response, flags=pb_utils.TRITONSERVER_RESPONSE_COMPLETE_FINAL
+                )
+                self.reset_flag = True
+            else:
+                request.set_release_flags(
+                    pb_utils.TRITONSERVER_REQUEST_RELEASE_RESCHEDULE
+                )
+                response_sender.send(response)
+
+        return None
```
Tabrizian, line 719: Remove this line.

src/python_be.h
```diff
@@ -281,6 +281,9 @@ class ModelInstanceState : public BackendModelInstance {
   std::unique_ptr<boost::asio::thread_pool> thread_pool_;
   std::unordered_map<void*, std::shared_ptr<InferPayload>> infer_payload_;
   std::unique_ptr<RequestExecutor> request_executor_;
+  std::mutex response_factory_map_mutex_;
+  std::unordered_map<intptr_t, TRITONBACKEND_ResponseFactory*>
+      response_factory_map_;
```
Tabrizian, line 286: Response factory doesn't seem to be used?

krishung5: It is used for decoupled models. e.g. https://github.com/triton-inference-server/python_backend/pull/319/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR378

README.md
krishung5: I was wondering why we would need to return `return None` here?

README.md
krishung5: Agreed. Update the doc.

Tabrizian (commented):
The code changes look good to me. I was wondering do you think there could be a memory leak in non-decoupled case with these changes? For non-decoupled we create N responses in the beginning: https://github.com/triton-inference-server/python_backend/blob/60a9091cc232f77645a88a19ac63809866319e50/src/python_be.cc#L1380

How is the `None` responses going to be treated by server? I didn't see any flags adjustments for non-decoupled mode for responses.

krishung5:
> The code changes look good to me. I was wondering do you think there could be a memory leak in non-decoupled case with these changes? For non-decoupled we create N responses in the beginning:
> 
> https://github.com/triton-inference-server/python_backend/blob/60a9091cc232f77645a88a19ac63809866319e50/src/python_be.cc#L1380
> 
> How is the `None` responses going to be treated by server? I didn't see any flags adjustments for non-decoupled mode for responses.

The backend will clean up the response if the associated request is rescheduled. Please see here: https://github.com/triton-inference-server/python_backend/pull/319/files#diff-ca67ec8693bbdfbd82a858e307f87fa39fc9187d39228f1afc5375837dc3f60dR1577-R1585

Tabrizian (approved):

src/pb_stub.cc
krishung5: Resolved offline.

README.md
krishung5: Resolved offline.

PR #317, opened 2023-11-01
Enhanced python_backend autocomplete

oandreeva-nv:
Added an ability to add optional input from `autocomplete` function + set `model_transaction_policy` 
Note: pre-commit hook fails not on my code

PR #320, opened 2023-11-03
Follow up to autocomplete pr #317

oandreeva-nv:
This is a follow up PR to : #317 . Unfortunately, I missed essential part.

After we retrieved updated model configuration from core during the call:
https://github.com/triton-inference-server/python_backend/blob/0f1221129a01b067b93a4abc4a2c30a9e2856e01/src/python_be.cc#L1864

`decoupled_` is not updated on the python backend site, since it is a part of `ModelState`. 

I've added `PropagateAutoCompletedConfig() ` that does exactly that, i.e. looks for `model_transaction_policy` in the model config (already filled by core), and if `decoupled` is found in `model_transaction_policy` , we update `decoupled_` property.

Actual functionality is tested with vllm backend on this PR: https://github.com/triton-inference-server/vllm_backend/pull/20

### DIFF
--- a/src/python_be.h
+++ b/src/python_be.h
@@ -250,9 +250,9 @@ class ModelState : public BackendModel {
   // Validate Model Configuration
   TRITONSERVER_Error* ValidateModelConfig();
 
-  // Update values set during auto_complete to `ModelState`
-  // Currently only updates `decoupled_`
-  TRITONSERVER_Error* PropagateAutoCompletedConfig();
+  // Overrides `BackendModel::SetModelConfig` to also
+  // set `ModelState::decoupled_`
+  TRITONSERVER_Error* SetModelConfig();
 
   // Auto-complete stub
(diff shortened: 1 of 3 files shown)


src/python_be.h
```diff
@@ -247,6 +250,10 @@ class ModelState : public BackendModel {
   // Validate Model Configuration
   TRITONSERVER_Error* ValidateModelConfig();
 
+  // Overrides `BackendModel::SetModelConfig` to also
```
oandreeva-nv, line 253: I believe in this case, it is more like: // Hides `BackendModel::SetModelConfig` ... ,
since base class' function is not virtual. Happy to change

dyastremsky (approved):
Great work!

krishung5 (approved):

src/python_be.h
tanmayv25: No.. This looks fine to me.

tanmayv25 (approved):

PR #325, opened 2023-12-06
Fix BLS decoupled segfault and hang

krishung5:
This PR addresses the BLS decoupled segfault by correctly recording the `InferPayload` with the accurate address

Additionally, it resolves a hang occurring in the following scenario:
~~When the stub pushes a message of type `PYTHONSTUB_ResponseSend` to the queue, it [waits](https://github.com/triton-inference-server/python_backend/blob/main/src/response_sender.cc#L120) for the Python backend to complete. The queue is monitored by `DecoupledMessageQueueMonitor`. When there are numerous concurrent BLS requests, all threads in the `boost::asio::thread_pool` become occupied by the [ExecuteBLSRequest](https://github.com/triton-inference-server/python_backend/blob/main/src/python_be.cc#L819-L824) function. Consequently, the [ResponseSendDecoupled](https://github.com/triton-inference-server/python_backend/blob/main/src/python_be.cc#L811-L814) function is pushed into the queue, awaiting the next available thread. At this point, the stub process is blocked.~~

~~On the server side, within `InferResponseComplete`, for decoupled responses, it calls the callback function `SendBLSDecoupledResponse`, which pushes a message and [waits](https://github.com/triton-inference-server/python_backend/blob/main/src/python_be.cc#L1850) for the stub to pick it up. As the stub is still blocked, the server hangs.~~

~~One potential solution is to use separate thread pools to eliminate the dependency between them. I am open to any suggestions and discussions.~~

**Updated version**

By default, on the PYBE side, there are 32 threads in the `boost::asio::thread_pool`. When numerous concurrent BLS requests occur, all threads in the `boost::asio::thread_pool` become occupied by the [ExecuteBLSRequest](https://github.com/triton-inference-server/python_backend/blob/main/src/python_be.cc#L819-L824) function. These threads are waiting for the backend invoked by the BLS request to execute inference and return the response via the `InferResponseComplete` callback function.

Within the callback, for decoupled responses, we call `SendBLSDecoupledResponse`, which sends a decoupled response to the stub and [waits](https://github.com/triton-inference-server/python_backend/blob/main/src/python_be.cc#L1850) for the stub to process the response. Meanwhile, on the stub side, when [loading the response from shm](https://github.com/triton-inference-server/python_backend/blob/main/src/pb_stub.cc#L1340), it requires the [GIL](https://github.com/triton-inference-server/python_backend/blob/main/src/infer_response.cc#L159).

If, at this moment, a `ResponseSender` tries to [send a response to the parent process](https://github.com/triton-inference-server/python_backend/blob/main/src/response_sender.cc#L118) beforehand, it acquires the GIL. The `ResponseSender` will then be waiting for the parent process to [process the response](https://github.com/triton-inference-server/python_backend/blob/main/src/python_be.cc#L814) sent by itself. The GIL will be held until there is an available thread.

Typically, there is only a single thread to execute the inference and invoke the callback when returning responses. Therefore, those 32 threads cannot proceed since they are all waiting on the `InferResponseComplete` callback, leading to a deadlock.

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -465,7 +465,6 @@ InferRequest::Exec(const bool is_decoupled)
   });
 
   try {
-    py::gil_scoped_release release;
     ipc_message = IPCMessage::Create(shm_pool, true /* inline_response */);
     bool has_exception = false;
     PythonBackendException pb_exception(std::string{});
@@ -508,6 +507,13 @@ InferRequest::Exec(const bool is_decoupled)
 
     // Send the BLS request to the parent process and wait for the response.
     {
+      // Release the GIL before sending the message to the stub process. This
+      // avoids a potential deadlock situation in the parent process, where
+      // every thread in the thread pool is indirectly waiting for a function in
+      // the stub process that acquires the GIL. Meanwhile, the current thread,
+      // which holds the GIL, is also waiting for the parent side to have the
+      // next available thread to pick up the job during resource contention.
+      py::gil_scoped_release release;
       bi::scoped_lock<bi::interprocess_mutex> lock{
           *(ipc_message->ResponseMutex())};
       stub->SendIPCMessage(ipc_message);
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -442,6 +442,13 @@ InferRequest::GetResponseSender()
 std::shared_ptr<InferResponse>
 InferRequest::Exec(const bool is_decoupled)
 {
+  // Release the GIL. This avoids a potential deadlock situation in the parent
+  // process, where every thread in the thread pool is indirectly waiting for a
+  // function in the stub process that acquires the GIL. Meanwhile, the current
+  // thread, which holds the GIL, is also waiting for the parent side to have
+  // the next available thread to pick up the job during resource contention.
+  py::gil_scoped_release release;
+
   // BLS should not be used in "initialize" or "finalize" function.
   std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
   if (!stub->IsInitialized() || stub->IsFinalizing()) {
@@ -507,13 +514,6 @@ InferRequest::Exec(const bool is_decoupled)
 
     // Send the BLS request to the parent process and wait for the response.
     {
-      // Release the GIL before sending the message to the stub process. This
-      // avoids a potential deadlock situation in the parent process, where
-      // every thread in the thread pool is indirectly waiting for a function in
-      // the stub process that acquires the GIL. Meanwhile, the current thread,
-      // which holds the GIL, is also waiting for the parent side to have the
-      // next available thread to pick up the job during resource contention.
-      py::gil_scoped_release release;
       bi::scoped_lock<bi::interprocess_mutex> lock{
           *(ipc_message->ResponseMutex())};
       stub->SendIPCMessage(ipc_message);
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -752,7 +752,7 @@ ModelInstanceState::ExecuteBLSRequest(
         if (is_decoupled && (infer_response->Id() != nullptr)) {
           // Need to manage the lifetime of InferPayload object for bls
           // decoupled responses.
-          infer_payload_[reinterpret_cast<void*>(&infer_payload)] =
+          infer_payload_[reinterpret_cast<intptr_t>(infer_payload.get())] =
               infer_payload;
         }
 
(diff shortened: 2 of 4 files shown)


src/python_be.cc
```diff
@@ -430,8 +430,22 @@ ModelInstanceState::LaunchStubProcess()
   StartMonitor();
   RETURN_IF_ERROR(Stub()->Launch());
 
-  thread_pool_ = std::make_unique<boost::asio::thread_pool>(
-      model_state->StateForBackend()->thread_pool_size);
+  // Use the default thread pool for BLS requests, and the decoupled response
+  // thread pool for decoupled responses to avoid deadlock between the stub and
+  // the parent process. Only use the default thread pool when there is only one
+  // thread in the thread pool.
+  int thread_pool_size = model_state->StateForBackend()->thread_pool_size;
```
nnshah1, line 437: the deadlock is caused by resource contention for threads? And so we now create two disjoint threadpools to ensure responses and requests can always progress?

I don't quite follow the 2/3 ratio and why we wouldn't always have two pools to avoid deadlock (even in the case the setting is for 1 thread).

I can follow up offline -

src/python_be.cc
```diff
@@ -752,7 +766,7 @@ ModelInstanceState::ExecuteBLSRequest(
         if (is_decoupled && (infer_response->Id() != nullptr)) {
           // Need to manage the lifetime of InferPayload object for bls
           // decoupled responses.
-          infer_payload_[reinterpret_cast<void*>(&infer_payload)] =
+          infer_payload_[reinterpret_cast<intptr_t>(infer_payload.get())] =
```
nnshah1, line 769: why change from void* ? can we use the type of infer_payload.get() directly?

src/python_be.h
```diff
@@ -286,7 +286,8 @@ class ModelInstanceState : public BackendModelInstance {
   std::unique_ptr<IPCMessage> received_message_;
   std::vector<std::future<void>> futures_;
   std::unique_ptr<boost::asio::thread_pool> thread_pool_;
-  std::unordered_map<void*, std::shared_ptr<InferPayload>> infer_payload_;
+  std::unique_ptr<boost::asio::thread_pool> decoupled_response_thread_pool_;
+  std::unordered_map<intptr_t, std::shared_ptr<InferPayload>> infer_payload_;
```
nnshah1, line 290: I see- we changed to intptr_t for consistency with the other map?

krishung5: Yes, that's correct.

src/python_be.cc
krishung5: The 2/3 ratio is just that I thought the default_thread_pool should have more threads since it handles both non-decoupled and decoupled BLS requests.

> why we wouldn't always have two pools to avoid deadlock (even in the case the setting is for 1 thread

We definitely can always have two pools and each of them has one thread. This would make the logic easier. Just wanted to make it consistent to the number of threads that users put in via the command line. i.e. when users specify only one thread but there're actually two.

nnshah1: if we decide to go with always having it (my inclination) - I think we can add to release notes.

krishung5:
I updated the description regarding the deadlock situation. I missed mentioning the GIL part. I hope this version is clearer. Shout out to @tanmayv25, who helped me clarify the backend mechanism in detail!

I think instead of using two thread pools to ensure that decoupled send and BLS requests can always proceed, we can just release the GIL when sending the message to the other process. This way, the other processes that acquire the GIL can proceed, which avoids the deadlock situation.

src/infer_request.cc
```diff
@@ -508,6 +507,13 @@ InferRequest::Exec(const bool is_decoupled)
 
     // Send the BLS request to the parent process and wait for the response.
     {
+      // Release the GIL before sending the message to the stub process. This
+      // avoids a potential deadlock situation in the parent process, where
+      // every thread in the thread pool is indirectly waiting for a function in
+      // the stub process that acquires the GIL. Meanwhile, the current thread,
+      // which holds the GIL, is also waiting for the parent side to have the
+      // next available thread to pick up the job during resource contention.
+      py::gil_scoped_release release;
```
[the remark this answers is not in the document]
krishung5: Reverted this change back. We can just release the GIL in the beginning.

Tabrizian (approved):

src/python_be.cc
krishung5: Resolving this as in the latest change doesn't require the thread pool changes.

PR #327, opened 2023-12-12
Fix segfault for decoupled models

krishung5:
This PR fixed the segfault issues:
- when an exception is raised within a decoupled model. We expected users to handle error by returning an error response with a final flag like this https://github.com/triton-inference-server/server/blob/main/qa/L0_backend_python/decoupled/models/decoupled_execute_error/1/model.py#L95-L106. This fix should also cover the use case where an exception is raised by calling `raise Exception("Intentional Error")` without sending any final flag via response sender.
- when there are lots of concurrent requests to decoupled models, the request object may have the same address which causes issues when using that address as id to retrieve the response factory from the map. This PR removed the usage of response_factory_map. The backend will create a new response factory for each request, no matter if the request is a rescheduled one or not, and clean up the response_factory when `response_sender` goes out of scope in the Python model to avoid memory growth. 

Added testing: https://github.com/triton-inference-server/server/pull/6686


src/python_be.cc
```diff
@@ -1361,18 +1361,24 @@ ModelInstanceState::ProcessRequestsDecoupled(
...
   if (response_batch.data_->has_error) {
+    for (auto& infer_request : pb_infer_requests) {
+      // Reset the release flags for all the requests.
+      infer_request->SetReleaseFlags(TRITONSERVER_REQUEST_RELEASE_ALL);
+      // Clean up the response factory map.
+      {
+        std::lock_guard<std::mutex> guard{response_factory_map_mutex_};
+        response_factory_map_.erase(
+            reinterpret_cast<intptr_t>(infer_request->RequestAddress()));
+      }
```
rmccorm4, line 1372: Would it make sense to encapsulate this in some `infer_request->Reset()` ?

krishung5: I think having them as is makes more sense to me since `response_factory_map_` and the mutex are not part of `InferRequest`, but open to any suggestion.

rmccorm4: Ah didn't look too closely, makes sense as is.

src/python_be.cc
```diff
@@ -1361,18 +1361,24 @@ ModelInstanceState::ProcessRequestsDecoupled(
   reporter.SetBatchStatistics(request_count);
 
   if (response_batch.data_->has_error) {
+    for (auto& infer_request : pb_infer_requests) {
+      // Reset the release flags for all the requests.
+      infer_request->SetReleaseFlags(TRITONSERVER_REQUEST_RELEASE_ALL);
+      // Clean up the response factory map.
+      {
+        std::lock_guard<std::mutex> guard{response_factory_map_mutex_};
```
nnshah1, line 1369: would it be better to hold the mutex across the whole operation? Seems odd at first glance to lock and unlock for each request -

src/python_be.cc
krishung5: Makes sense! Encapsulated this in a separate function. I think it makes more sense to move that function to the outer scope so that the requests and response factory can be cleaned up at one place.

> Are the request objects re-used or something here? Curious why we are "resetting" release flags in general.

The request objects will be used at the end of `TRITONBACKEND_ModelInstanceExecute` to get the flags when calling `TRITONBACKEND_RequestRelease`. Resetting the flags here because the flags may be set to "rescheduled" in the model.

src/python_be.cc
krishung5: The mutex is only required for the operation on the map. Holding the mutex across the entire operation might slightly block other places where something is being inserted or read from the map.

Tabrizian (approved):
Is it going to be cherry-picked for 23.12?

rmccorm4 (approved):

krishung5:
> Is it going to be cherry-picked for 23.12?

This is to fix a segfault so I think we could if we have not staged the container. @tanmayv25 May I on your thoughts on this?

src/python_be.cc
nnshah1: I think acquiring and releasing may have additional overhad in itself - but not positive

src/ipc_message.h
```diff
@@ -54,7 +54,8 @@ typedef enum PYTHONSTUB_commandtype_enum {
   PYTHONSTUB_AutoCompleteRequest,
   PYTHONSTUB_AutoCompleteResponse,
   PYTHONSTUB_LogRequest,
-  PYTHONSTUB_CleanupRequest,
+  PYTHONSTUB_BLSDecoupledInferPayloadCleanup,
+  PYTHONSTUB_BLSDecoupledResponseFactoryCleanup,
```
Tabrizian, line 58:
```suggestion
  PYTHONSTUB_DecoupledResponseFactoryCleanup,
```

Tabrizian (approved):

src/ipc_message.h
krishung5: Will address this in a separate PR later.

PR #332, opened 2023-12-26
Bumping min required cxx standard to 17

oandreeva-nv:
This PR is one of the series PRs to update Triton to C++17 standard.

Introduced logic: either set `TRITON_MIN_CXX_STANDARD` to 17 or use cached value. Later,`TRITON_MIN_CXX_STANDARD`  is used to set min required standard through `target_compile_features`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,9 @@ cmake_minimum_required(VERSION 3.17)
 
 project(tritonpythonbackend LANGUAGES C CXX)
 
+# Use C++17 standard as Triton's minimum required.
+set(TRITON_MIN_CXX_STANDARD 17 CACHE STRING "The minimum C++ standard which features are requested to build this target.")
+
 #
 # Options
 #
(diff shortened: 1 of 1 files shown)


GuanLuo (approved):

CMakeLists.txt
```diff
@@ -231,14 +234,14 @@ add_library(
   TritonPythonBackend::triton-python-backend ALIAS triton-python-backend
 )
 
-target_compile_features(triton-python-backend PRIVATE cxx_std_11)
+target_compile_features(triton-python-backend PRIVATE cxx_std_${TRITON_MIN_CXX_STANDARD})
```
Tabrizian, line 237: Why not use the existing `CMAKE_CXX_STANDARD` variable. For example, `cxx_std_${CMAKE_CXX_STANDARD}`?

Tabrizian: Seems like some of the libraries follow this pattern: https://github.com/IntelRealSense/librealsense/blob/ea975f8641a83a38a3f2ef3baa4cc42e94a75219/CMake/lrs_macros.cmake#L24

oandreeva-nv: we do not explicitly define and set `CMAKE_CXX_STANDARD`, nor do we propagate it.To use `CMAKE_CXX_STANDARD`, we would need to set  it first or see if it was passed by user through arguments, since we want to move to C++17, we will also need to check if passed `CMAKE_CXX_STANDARD` is >=17. Plus, from what I've researched so far, if `CMAKE_CXX_STANDARD` is used, it is typically used through `set(CMAKE_CXX_STANDARD  <STANDARD>)`, which is too restrictive, as I mentioned in the slack discussion:

it will force the project to be compiled in specified standard only, it may cause some troubles with later upgrades of Ubuntu or other upgrades.  target_compile_features just specifies [minimum required standard](https://www.scivision.dev/cmake-force-old-language-standard/). Thus, we are not restricted. Some of related discussions on this topic: https://sarcasm.github.io/notes/dev/cmake.html , https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1525 and https://github.com/AcademySoftwareFoundation/aswf-sample-project/pull/8#issuecomment-952962520

Tabrizian: How about keeping everything as-is (was) and just making sure that `CMAKE_CXX_STANDARD` is propagated to all the repos? We can set it once during the build.py script by specifying `-DCMAKE_CXX_STANDARD=11,17,etc`.

oandreeva-nv: What if users will build with cmake explicitly? Then we would need to add more code that checks that `CMAKE_CXX_STANDARD` is set, and that the value is >= 17. Plus, propagation would mostly be done through cmake cache, i.e. code will still be similar to what I currently have. 

I don't see any benefit in this, could you please elaborate your concerns?

oandreeva-nv: > keeping everything as-is (was)

Could you please clarify this as well? Currently, min required standard is 11, which is hard-coded with the following:
`target_compile_features(triton-python-backend PRIVATE cxx_std_11)`

Tabrizian: I think now I have a better idea about the challenges involved. thanks for clarifying. Resolving.

Tabrizian (approved):

CMakeLists.txt
oandreeva-nv: No problem! Thanks for the discussion.

PR #334, opened 2024-01-08
Move from jfrog artifactory to archives.boost.io to fix boost download

rmccorm4:
Confirmed sha256sum is the same locally

Server: https://github.com/triton-inference-server/server/pull/6775
Python Backend: https://github.com/triton-inference-server/python_backend/pull/334
Gitlab: !1027

Based on issues with jfrog similar to this: https://github.com/boostorg/boost/issues/849#issue-2069720679

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,7 +100,7 @@ FetchContent_MakeAvailable(dlpack)
 #
 ExternalProject_Add(
   boostorg
-  URL https://boostorg.jfrog.io/artifactory/main/release/1.79.0/source/boost_1_79_0.tar.gz
+  URL https://archives.boost.io/release/1.79.0/source/boost_1_79_0.tar.gz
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -100,7 +100,7 @@ FetchContent_MakeAvailable(dlpack)
 #
 ExternalProject_Add(
   boostorg
-  URL https://boostorg.jfrog.io/artifactory/main/release/1.79.0/source/boost_1_79_0.tar.gz
+  URL https://archives.boost.io/release/1.79.0/source/boost_1_79_0.tar.gz
```
mc-nv, line 103: Can we align versions with installed one in container which 1.80.0 ?

rmccorm4: If it's OK with you, I'll leave this one as-is to see if it fixes CI. 

Then I'll make a separate PR to lift version to 1.80 to match. I don't know if we could be doing something smarter to re-use assets from server build rather than re-download here though.

Tabrizian (approved):

PR #336, opened 2024-01-11
Fix decoupled ensemble IO shared memory leak

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/6784

The shm block that holds an shm address to another shm block is no longer accessed after dereference. Although the block will eventually be cleaned up on next decoupled inference request or model unload, it is still a leak for this current decoupled inference request and cause the decoupled ensemble IO test to constantly fail on shm leak detected.


kthui:
Will be part of another PR.

PR #337, opened 2024-01-17
Improve decoupled shm handling

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/6803

Free decoupled shm memory as soon as the memory will not be accessed any more.

### DIFF
--- a/src/ipc_message.cc
+++ b/src/ipc_message.cc
@@ -59,11 +59,10 @@ IPCMessage::Create(
 std::unique_ptr<IPCMessage>
 IPCMessage::LoadFromSharedMemory(
     std::unique_ptr<SharedMemoryManager>& shm_pool,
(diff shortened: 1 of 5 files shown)


rmccorm4 (approved):
Happy to see a 1-line fix!

Is this the small "growth" (~dozen bytes or so) we saw consistently in this test? Or does it address any other growths or issues as well?

kthui:
> Is this the small "growth" (~dozen bytes or so) we saw consistently in this test? Or does it address any other growths or issues as well?

It's the consistent small growth on the decoupled IO test. Might fixed other decoupled issue, but I didn't test for it explicitly.

krishung5 (approved):
Great catch!

Tabrizian (approved):
It is not really a leak based on the offline conversation so perhaps change the PR title to 
"Release received message to avoid shm probe failures".

PR #339, opened 2024-02-01
Add support for correlation string in BLS calls in python backend

pskiran1:


src/pb_stub.cc
```diff
@@ -1670,6 +1684,7 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
           py::return_value_policy::reference_internal)
       .def("request_id", &InferRequest::RequestId)
       .def("correlation_id", &InferRequest::CorrelationId)
+      .def("correlation_id_string", &InferRequest::CorrelationIdString)
```
Tabrizian, line 1687: Let's only have a single `correlation_id` field and it should return string or int depending on which one has been set.

src/python_be.cc
```diff
@@ -363,8 +363,13 @@ ModelInstanceState::SaveRequestsToSharedMemory(
     RETURN_IF_ERROR(TRITONBACKEND_RequestId(request, &id));
 
     uint64_t correlation_id;
```
Tabrizian, line 365: initialize to zero?

pskiran1:
Closing the draft PR.
Please find the https://github.com/triton-inference-server/python_backend/pull/344 with more enhancements.

PR #340, opened 2024-02-15
[Windows Port] IPC Capability Check

fpetrini15:


src/python_be.cc
```diff
@@ -512,6 +512,25 @@ ModelInstanceState::GetInputTensor(
   cpu_only_tensors = true;
 #endif  // TRITON_ENABLE_GPU
 
+// For Windows, force CPU tensors if IPC is not supported on
+// the target GPU device
+#if defined(TRITON_ENABLE_GPU) && defined(_WIN32)
+  if (src_memory_type == TRITONSERVER_MEMORY_GPU) {
```
fpetrini15, line 518: Merge if-statements for legibility.

nv-kmcgill53: This is likely fine. The compiler should be able to optimize this and if not, it's a well predicted branch

PR #344, opened 2024-03-08
Add Correlation Id string support for BLS

pskiran1:


src/correlation_id.h
```diff
@@ -0,0 +1,93 @@
...
+  /// \param handle Shared memory handle of the CorrelationId.
+  /// \return Returns the CorrelationId in the specified handle
+  /// location.
+  static std::unique_ptr<CorrelationId> LoadFromSharedMemory(
+      std::unique_ptr<SharedMemoryManager>& shm_pool,
+      bi::managed_external_buffer::handle_t handle);
+
+  // Functions that help determine exact type of Correlation Id
+  CorrelationIdDataType Type() const { return id_type_; }
+  bool IsNotEmpty() const { return ((id_string_ != "") || (id_uint_ != 0)); }
```
krishung5, line 67: I was wondering if we are using this function somewhere?

src/infer_request.cc
```diff
@@ -396,6 +375,12 @@ InferRequest::InferRequest(
   trace_ = infer_request_shm_ptr_->trace;
   request_release_flags_ = infer_request_shm_ptr_->request_release_flags;
 
+  if (correlation_id_shm->Type() == CorrelationIdDataType::UINT64) {
+    correlation_id_ = CorrelationId(correlation_id_shm->UnsignedIntValue());
+  } else {
+    correlation_id_ = CorrelationId(correlation_id_shm->StringValue());
+  }
+
```
krishung5, line 383: Can we use constructor to initialize `correlation_id_`?

krishung5 (changes requested):
nit: Need to update the copyright of the files.

src/correlation_id.h
pskiran1: At present, we are not using this function and removed it for now.

krishung5 (approved):
Great work, Sai!

src/python_be.cc
```diff
@@ -362,9 +363,19 @@ ModelInstanceState::SaveRequestsToSharedMemory(
...
-    uint64_t correlation_id;
-    RETURN_IF_ERROR(
-        TRITONBACKEND_RequestCorrelationId(request, &correlation_id));
+    uint64_t correlation_id_uint = 0;
+    CorrelationId correlation_id;
+
+    auto error =
+        TRITONBACKEND_RequestCorrelationId(request, &correlation_id_uint);
+    if (error != nullptr) {
+      const char* correlation_id_string = "";
```
rmccorm4, line 372: @krishung5 @pskiran1 is this a memory leak not deleting the TRITONSERVER_Error ?

krishung5: Ah.. this is a really great catch, thanks Ryan. I think there will be a memory leak when the error is not deleted. @pskiran1 Can you create a follow-up PR to fix this?

pskiran1: Thank you @rmccorm4, I have created a follow-up PR #345 to delete the error.

PR #346, opened 2024-03-14
Exposing trace context to python backend

oandreeva-nv:
Added needed logic to pass trace context to the stub proces + accessing it from the stub process.

Current discussion: `get_trace` does it need a `mode` field: https://github.com/triton-inference-server/server/pull/6985/files#r1525367279

Most likely `mode` field will be removed from `get_trace` api.

Related PRs:
core: https://github.com/triton-inference-server/core/pull/334
server: https://github.com/triton-inference-server/server/pull/6985

### DIFF
--- a/src/infer_trace.h
+++ b/src/infer_trace.h
@@ -34,6 +34,7 @@
 namespace triton { namespace backend { namespace python {
 
 struct InferenceTraceShm {
+  bi::managed_external_buffer::handle_t trace_context_shm_handle;
   // The address of the 'TRITONSERVER_InferTrace' object.
   void* triton_trace;
 };
@@ -62,19 +63,26 @@ class InferenceTrace {
       bi::managed_external_buffer::handle_t handle);
 
   void* TritonTrace() { return triton_trace_; }
+  void SetContext(std::string trace_context) { trace_context_ = trace_context; }
+  const std::string& Context() const { return trace_context_; }
 
   bi::managed_external_buffer::handle_t ShmHandle() { return shm_handle_; }
 
  private:
   // The private constructor for creating a InferenceTrace object from shared
   // memory.
-  InferenceTrace(AllocatedSharedMemory<InferenceTraceShm>& infer_trace_shm);
+  InferenceTrace(
+      AllocatedSharedMemory<InferenceTraceShm>& infer_trace_shm,
+      std::unique_ptr<PbString>& trace_context_shm);
+
   void* triton_trace_;
+  std::string trace_context_;
 
   // Shared Memory Data Structures
   AllocatedSharedMemory<InferenceTraceShm> infer_trace_shm_;
   InferenceTraceShm* infer_trace_shm_ptr_;
   bi::managed_external_buffer::handle_t shm_handle_;
+  std::unique_ptr<PbString> trace_context_shm_;
 };
 
 }}};  // namespace triton::backend::python
--- a/src/infer_trace.h
+++ b/src/infer_trace.h
@@ -1,0 +1,80 @@
+// Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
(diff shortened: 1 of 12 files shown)


src/python_be.cc
```diff
@@ -387,10 +387,21 @@ ModelInstanceState::SaveRequestsToSharedMemory(
...
+      LOG_MESSAGE(TRITONSERVER_LOG_ERROR, TRITONSERVER_ErrorMessage(err));
       TRITONSERVER_ErrorDelete(err);
     }
+    const char* val = nullptr;
+    if (triton_trace != nullptr) {
+      err = TRITONSERVER_InferenceTraceContext(triton_trace, &val);
+      if (err != nullptr) {
+        LOG_MESSAGE(TRITONSERVER_LOG_ERROR, TRITONSERVER_ErrorMessage(err));
+        TRITONSERVER_ErrorDelete(err);
+      }
```
rmccorm4, line 399: I think `LOG_IF_ERROR` does this, something like:
```suggestion
      LOG_IF_ERROR(TRITONSERVER_InferenceTraceContext(triton_trace, &val), "failed to get trace context");
```

src/pb_stub.cc
```diff
@@ -1611,7 +1611,17 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
+      .def(
+          "get_context",
+          [](InferenceTrace& self, const std::string mode) -> py::object {
+            auto context = (mode == "opentelemetry") ? self.Context() : "";
+            if (context != "") {
+              return py::str(context);
+            }
+            return py::none();
+          },
+          py::arg("mode").none(false) = "triton");
```
rmccorm4, line 1624: I'm in favor of removing mode for now: https://github.com/triton-inference-server/server/pull/6985/files#r1525487116

src/python_be.cc
```diff
@@ -387,10 +387,21 @@ ModelInstanceState::SaveRequestsToSharedMemory(
     auto err = TRITONBACKEND_RequestTrace(request, &triton_trace);
     if (err != nullptr) {
       triton_trace = nullptr;
+      LOG_MESSAGE(TRITONSERVER_LOG_ERROR, TRITONSERVER_ErrorMessage(err));
```
rmccorm4, line 390: Wasn't this error intentionally silenced as described by comment above?

oandreeva-nv: This one should just print it, not fail a response. The comment above explains why we don't do:
```
RETURN_IF_ERROR(TRITONBACKEND_RequestTrace(request, &triton_trace));
```
here. I think at least printing error message should be helpful with the debug.

oandreeva-nv: Actually, I'll remove it. Otherwise when Triton is compiled without tracing, it'll keep logging message.

rmccorm4: Yeah that's what I was worried about

oandreeva-nv: Good catch!

krishung5 (approved):
LGTM, especially on the shm part, great work 🚀 

I understand that this is only stage 1, but I was wondering if we plan to add some documents on how to interact with the trace context? (maybe we'd want to wait until the stage 2 is completed?) I think the demo code in the server PR would be really helpful(at least for my own understanding)

kthui (approved):

src/pb_stub.cc
oandreeva-nv: updated

oandreeva-nv:
@krishung5, yes good point. I'll update docs in a separate PR, so that this can be cherry-picked faster

src/python_be.cc
oandreeva-nv: updated

src/infer_trace.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include "pb_string.h"
+#include "pb_utils.h"
+
+namespace triton { namespace backend { namespace python {
+
+struct InferenceTraceShm {
+  bi::managed_external_buffer::handle_t trace_context_shm_handle;
+  // The address of the 'TRITONSERVER_InferTrace' object.
+  void* triton_trace;
```
rmccorm4, line 39: Can you remind me why this is `void*` ? To avoid C API references out of process in stub?

rmccorm4 (approved):
Looks great! Had a question, but unrelated to approval: https://github.com/triton-inference-server/python_backend/pull/346/files#r1526885602

src/infer_trace.h
oandreeva-nv: yes, stub does not have an access to TRITONSERVER_InferTrace

PR #350, opened 2024-04-01
Decoupled Async Execute

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/7062

Add support for decoupled async execute function on `model.py` and enable overlapping between different calls into the execute function by having them execute on the same event loop. Also, enable async exec BLS function for decoupled async execute function.

Closes: https://github.com/triton-inference-server/server/issues/3482

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,9 +620,22 @@ full power of what can be achieved from decoupled API. Read
 [Decoupled Backends and Models](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/decoupled_models.md)
 for more details on how to host a decoupled model.
 
-##### Known Issues
+##### Async Execute
 
-* Currently, decoupled Python models can not make async infer requests.
+Starting from 24.04, `async def execute(self, requests):` is supported for
+decoupled Python models. Its coroutine will be executed by an AsyncIO event loop
+shared with requests executing in a model instance. The next request for the
+model instance can start executing while the current request is waiting.
+
+This is useful for minimizing the number of model instances for models that
+spend the majority of its time waiting, given requests can be executed
+"concurrently" by AsyncIO. To take full advantage of the "concurrency", it is
+vital for the async execute function to not block the event loop from making
+progress while it is waiting, i.e. downloading over the network.
+
+Limitations:
+* The server/backend do not control how many requests can be executed
+"concurrently" by a model instance.
 
 #### Request Rescheduling
 
--- a/README.md
+++ b/README.md
@@ -624,8 +624,8 @@ for more details on how to host a decoupled model.
 
 Starting from 24.04, `async def execute(self, requests):` is supported for
 decoupled Python models. Its coroutine will be executed by an AsyncIO event loop
-shared with requests executing in a model instance. The next request for the
-model instance can start executing while the current request is waiting.
+shared with requests executing in the same model instance. The next request for
+the model instance can start executing while the current request is waiting.
 
 This is useful for minimizing the number of model instances for models that
 spend the majority of its time waiting, given requests can be executed
--- a/README.md
+++ b/README.md
@@ -629,15 +629,15 @@ the model instance can start executing while the current request is waiting.
 
 This is useful for minimizing the number of model instances for models that
 spend the majority of its time waiting, given requests can be executed
-"concurrently" by AsyncIO. To take full advantage of the "concurrency", it is
-vital for the async execute function to not block the event loop from making
-progress while it is waiting, i.e. downloading over the network.
+concurrently by AsyncIO. To take full advantage of the concurrency, it is vital
+for the async execute function to not block the event loop from making progress
+while it is waiting, i.e. downloading over the network.
 
 Notes:
 * The model should not modify the running event loop, as this might cause
 unexpected issues.
-* The server/backend do not control how many requests can be executed
-"concurrently" by a model instance.
+* The server/backend do not control how many requests are added to the event
+loop by a model instance.
 
 #### Request Rescheduling
 
--- a/README.md
+++ b/README.md
@@ -633,7 +633,9 @@ spend the majority of its time waiting, given requests can be executed
 vital for the async execute function to not block the event loop from making
 progress while it is waiting, i.e. downloading over the network.
 
-Limitations:
+Notes:
+* The model should not modify the running event loop, as this might cause
+unexpected issues.
 * The server/backend do not control how many requests can be executed
 "concurrently" by a model instance.
 
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ any C++ code.
       - [Request Cancellation Handling](#request-cancellation-handling)
       - [Decoupled mode](#decoupled-mode)
         - [Use Cases](#use-cases)
-        - [Known Issues](#known-issues)
+        - [Async Execute](#async-execute)
       - [Request Rescheduling](#request-rescheduling)
     - [`finalize`](#finalize)
   - [Model Config File](#model-config-file)
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -904,33 +904,30 @@ Stub::RunCoroutine(py::object coroutine)
   py::object py_future = py::module_::import("asyncio").attr(
       "run_coroutine_threadsafe")(coroutine, loop);
 
-  {
-    std::lock_guard<std::mutex> lock(async_event_futures_mu_);
-
-    std::shared_ptr<std::future<void>> shared_future(new std::future<void>());
-    std::future<void> c_future = std::async(
-        std::launch::async, [this, shared_future, py_future]() mutable {
-          {
-            py::gil_scoped_acquire gil_acquire;
-            try {
-              py_future.attr("result")();
-            }
-            catch (const PythonBackendException& pb_exception) {
-              LOG_ERROR << pb_exception.what();
-            }
-            catch (const py::error_already_set& error) {
-              LOG_ERROR << error.what();
-            }
-            py_future = py::none();
+  std::shared_ptr<std::future<void>> shared_future(new std::future<void>());
+  std::future<void> c_future = std::async(
+      std::launch::async, [this, shared_future, py_future]() mutable {
+        {
+          py::gil_scoped_acquire gil_acquire;
+          try {
+            py_future.attr("result")();
           }
-          {
-            std::lock_guard<std::mutex> lock(async_event_futures_mu_);
-            async_event_futures_.erase(shared_future);
+          catch (const PythonBackendException& pb_exception) {
+            LOG_ERROR << pb_exception.what();
           }
-        });
-    *shared_future = std::move(c_future);
-    async_event_futures_.emplace(std::move(shared_future));
-  }
+          catch (const py::error_already_set& error) {
+            LOG_ERROR << error.what();
+          }
+          py_future = py::none();
+        }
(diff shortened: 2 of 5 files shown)


src/pb_stub.cc
```diff
@@ -870,6 +879,59 @@ Stub::ProcessRequests(RequestBatch* request_batch_shm_ptr)
...
+          py_future = py::none();
+        }
+        std::vector<std::shared_ptr<std::future<void>>> empty;
+        {
+          std::lock_guard<std::mutex> lock(async_event_futures_mu_);
+          done_async_event_futures_.swap(empty);
+          done_async_event_futures_.emplace_back(std::move(shared_future));
+        }
+      });
+  *shared_future = std::move(c_future);
```
Tabrizian, line 930: could you elaborate why we need a separate thread for every future that just waits for the future is C++?

Tabrizian (commented):
Is removing the bls async_exec limitation going to be a separate ticket?

kthui:
> Is removing the bls async_exec limitation going to be a separate ticket?

The limitation is already removed.
https://github.com/triton-inference-server/python_backend/pull/350/files#diff-f2bb64cbc9d4dc5d938ce965813b23178f53732e02e8c3a2afa707e287e92e57L1732-L1736

Tests:
https://github.com/triton-inference-server/server/pull/7062/files#diff-733cdb25e87f4d1cf11e7ba218d45ca862a54fb6b909e6a304f105b8dcc57c44R88-R98

src/pb_stub.cc
kthui: Since we want to enable overlapping in calls to async execute function and the function does not return anything meaningful (always return `None`), the stub process will report the execution is completed while leaving the async execute coroutine to execute in the background and allowing it to generate new responses until it returns. Two things can happen while the coroutine is in the background
1. It returns normally - not a concern.
2. It raises an exception. 

The purpose of the c future is to keep track of the py future until its coroutine is done. Without the tracking, any exception raised by the py future will be lost. The c future is blocked on the `py_future.result()` call, until the py future is completed. If the py future raises an exception, the c future will capture it and prints it to the log.

Since `py_future.result()` is blocking, separate threads are necessary. Why not just use the `py_future.done()` check, which is non-blocking? The done check only returns `True` if the coroutine completes successfully (or cancelled) and does not raise an exception. The goal is to capture the exception, so the only alternative is `py_future.exception()` which is equally blocking as the `py_future.result()`, so `py_future.result()` is picked for its easy to use, as it reports  anything abnormal as an exception.

Tabrizian:
@kthui Ah, sorry I missed that. Thanks for the correction!

src/pb_stub.cc
Tabrizian: Can the user utilize the ` loop.set_exception_handler(handle_exception)` to handle the exceptions? We could have a default one that logs the exceptions?

src/pb_stub.cc
```diff
@@ -870,6 +879,59 @@ Stub::ProcessRequests(RequestBatch* request_batch_shm_ptr)
...
+          catch (const py::error_already_set& error) {
+            LOG_ERROR << error.what();
+          }
+          py_future = py::none();
+        }
+        std::vector<std::shared_ptr<std::future<void>>> empty;
+        {
+          std::lock_guard<std::mutex> lock(async_event_futures_mu_);
+          done_async_event_futures_.swap(empty);
+          done_async_event_futures_.emplace_back(std::move(shared_future));
```
GuanLuo, line 927: I don't think you need a vector of futures, just a shared_pointer to previously holding future and swap in the future associated to current task when the current task is done should be sufficient

src/pb_stub.cc
kthui: I don't think the `def handle_exception(loop, context)` set by `loop.set_exception_handler(handle_exception)` will get called. For example,
```python
#!/usr/bin/env python3

import asyncio
import threading

def get_async_event_loop():
    loop = asyncio.new_event_loop()
    asyncio.set_event_loop(loop)
    t = threading.Thread(target=loop.run_forever, daemon=True)
    t.start()
    return loop

def custom_exception_handler(loop, context):
    print("----- Exception caught by custom_exception_handler -----")
    print(context)

def run_coroutine(loop, coroutine_):
    py_future = asyncio.run_coroutine_threadsafe(coroutine_, loop)

    try:
        py_future.result()
    except Exception as e:
        print("----- Exception caught from py_future -----")
        print(e)

async def raise_exception():
    print("async raise_exception() start")
    await asyncio.sleep(1)
    print("async raise_exception() raise exception")
    raise ValueError("dummy exception")
    print("async raise_exception() end")

if __name__ == "__main__":
    loop = get_async_event_loop()
    loop.set_exception_handler(custom_exception_handler)
    run_coroutine(loop, raise_exception())
```
```
$ python3 aio_exception_handler.py
async raise_exception() start
async raise_exception() raise exception
----- Exception caught from py_future -----
dummy exception
```
Note: `loop.run_forever()` is a blocking until `loop.stop()` is called.

src/pb_stub.cc
kthui: Very slick idea!

src/pb_stub.cc
Tabrizian: I see, that's strange. @kthui can we perhaps have a background thread that monitors a vector of errors and logs them if there are any?

README.md
```diff
@@ -620,9 +620,22 @@ full power of what can be achieved from decoupled API. Read
...
+Starting from 24.04, `async def execute(self, requests):` is supported for
+decoupled Python models. Its coroutine will be executed by an AsyncIO event loop
+shared with requests executing in the same model instance. The next request for
+the model instance can start executing while the current request is waiting.
+
+This is useful for minimizing the number of model instances for models that
+spend the majority of its time waiting, given requests can be executed
+"concurrently" by AsyncIO. To take full advantage of the "concurrency", it is
+vital for the async execute function to not block the event loop from making
+progress while it is waiting, i.e. downloading over the network.
```
Tabrizian, line 634: Would be good to add a note that the user should not modify the running event loop otherwise there might be some unexpected issues.

src/pb_stub.cc
kthui: > have a background thread that monitors a vector of errors and logs them if there are any

The issue is how to get a vector of errors? There are two ways to get the error:
1. [`py_future.result()`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result) which raises an exception for any error; or
2. [`py_future.exception()`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.exception) which return the error, except raise exception if cancelled.

Unfortunately, both APIs are blocking, so there is a need for a thread to wait for each `py_future` until it is completed.
Fortunately, the threads are C++ threads, so any overhead should be minimal compare to those in Python.

There are only 3 APIs that are non-blocking, but non (or any combinations) of them can tell if there is an exception:
* [cancelled](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.cancelled)
* [running](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.running)
* [done](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.done)

The definition of `done` there is the coroutine completed running and either it is cancelled or completed successfully, which means a coroutine that raised an exception is not done, leaving us with the blocking options above.

README.md
kthui: Sure, this is a good point!
[Document model should not modify event loop](https://github.com/triton-inference-server/python_backend/pull/350/commits/00f0cb8873651fd0d93bb7273727178e6814d9ab)

src/pb_stub.cc
Tabrizian: Thanks for super detailed explanation. Could we perhaps use `add_done_callback`?

kthui: Sure. I still have no luck getting the exception logging to match exactly what we have from the others, but I could get it to print sufficient details, i.e.
```
...
E0408 21:43:36.182801 6016 pb_stub.cc:120] Traceback (most recent call last):
  File "/opt/tritonserver/qa/L0_backend_python/async_execute/models/async_execute_decouple/1/model.py", line 44, in execute
    self.raise_value_error(requests)
  File "/opt/tritonserver/qa/L0_backend_python/async_execute/models/async_execute_decouple/1/model.py", line 77, in raise_value_error
    raise ValueError("wait_secs cannot be negative")
ValueError: wait_secs cannot be negative

...
```

README.md
```diff
@@ -620,9 +620,24 @@ full power of what can be achieved from decoupled API. Read
...
 
-* Currently, decoupled Python models can not make async infer requests.
+Starting from 24.04, `async def execute(self, requests):` is supported for
+decoupled Python models. Its coroutine will be executed by an AsyncIO event loop
+shared with requests executing in the same model instance. The next request for
+the model instance can start executing while the current request is waiting.
+
+This is useful for minimizing the number of model instances for models that
+spend the majority of its time waiting, given requests can be executed
+"concurrently" by AsyncIO. To take full advantage of the "concurrency", it is
```
Tabrizian, line 632:
```suggestion
concurrently by AsyncIO. To take full advantage of the concurrency, it is
```

README.md
```diff
@@ -620,9 +620,24 @@ full power of what can be achieved from decoupled API. Read
...
+vital for the async execute function to not block the event loop from making
+progress while it is waiting, i.e. downloading over the network.
+
+Notes:
+* The model should not modify the running event loop, as this might cause
+unexpected issues.
+* The server/backend do not control how many requests can be executed
```
Tabrizian, line 639:
```suggestion
* The server/backend do not control how many requests are added to
```

README.md
```diff
@@ -620,9 +620,24 @@ full power of what can be achieved from decoupled API. Read
...
+"concurrently" by a model instance.
```
Tabrizian, line 640:
```suggestion
the event loop by a model instance.
```

src/pb_stub.cc
```diff
@@ -881,6 +945,10 @@ void
 Stub::Finalize()
 {
   finalizing_ = true;
+  // Stop async event loop if created.
+  if (!py::isinstance<py::none>(async_event_loop_)) {
+    async_event_loop_.attr("stop")();
+  }
```
Tabrizian, line 951: Do we need to wait for the thread to join as well?

README.md
kthui: [Update docs](https://github.com/triton-inference-server/python_backend/pull/350/commits/2b387c5db241b5c16e07a0f89fd0757d9752083b)

README.md
kthui: [Update docs](https://github.com/triton-inference-server/python_backend/pull/350/commits/2b387c5db241b5c16e07a0f89fd0757d9752083b)

README.md
kthui: [Update docs](https://github.com/triton-inference-server/python_backend/pull/350/commits/2b387c5db241b5c16e07a0f89fd0757d9752083b)

src/pb_stub.cc
kthui: I don't think it is needed, because the thread is started as daemon and it only does one thing
```python
async_event_loop_.run_forever()
```
which will return after calling
```python
async_event_loop_.stop()
```
and then the thread automatically exits.

Tabrizian (approved):

PR #351, opened 2024-04-05
Randomize Python backend shared memory region naming

Tabrizian:
I did some testing with 10 regions and everything seems to work fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1067,11 +1067,13 @@ will create additional threads instead of spawning separate processes.
 
 ## Running Multiple Instances of Triton Server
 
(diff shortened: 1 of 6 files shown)


rmccorm4 (commented):
Should we update that README section on shm region name collision and setting a prefix and such?

kthui (approved):
LGTM! We could probably mark the setting prefix option on the README as deprecated, because it is no longer necessary to avoid naming collision.

Tabrizian:
@rmccorm4 @jackyh docs are updated.

PR #353, opened 2024-04-11
Decoupled Async Execute (#350)

kthui:
* Add async decoupled execute

* Enable decoupled bls async exec

* Improve handling for async execute future object

* Add docs for async execute for decoupled model

* Fix link on docs

* Improve docs wording

* Improve destruction steps for async execute future object

* Piggy back on GIL for protection

* Document model should not modify event loop

* Use Python add_done_callback

* Protect infer_payload_

* Use traceback API that supports Python 3.8 and 3.9

* Update docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ any C++ code.
       - [Request Cancellation Handling](#request-cancellation-handling)
       - [Decoupled mode](#decoupled-mode)
         - [Use Cases](#use-cases)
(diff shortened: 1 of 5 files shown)


kthui:
Pick https://github.com/triton-inference-server/python_backend/pull/350 into r24.04

Tabrizian (approved):

PR #354, opened 2024-04-16
Reset async_event_loop_ only if initialized

kthui:
Shout-out to @fpetrini15 for finding the auto-complete stub process does not go through the `Initialize()` function but always go through the `Finalize()` function.

This fix the issue by checking the `initialized_` flag to determine if the `Initialize()` function is called before manipulating with the `async_event_loop_` variable initialized on the `Initialize()` function.

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -945,17 +945,19 @@ void
 Stub::Finalize()
 {
   finalizing_ = true;
-  // Stop async event loop if created.
-  if (!py::isinstance<py::none>(async_event_loop_)) {
-    async_event_loop_.attr("stop")();
-  }
-  // Call finalize if exists.
-  if (initialized_ && py::hasattr(model_instance_, "finalize")) {
-    try {
-      model_instance_.attr("finalize")();
+  if (initialized_) {
+    // Stop async event loop if created.
+    if (!py::isinstance<py::none>(async_event_loop_)) {
+      async_event_loop_.attr("stop")();
     }
-    catch (const py::error_already_set& e) {
-      LOG_INFO << e.what();
+    // Call finalize if exists.
+    if (py::hasattr(model_instance_, "finalize")) {
+      try {
+        model_instance_.attr("finalize")();
+      }
+      catch (const py::error_already_set& e) {
+        LOG_INFO << e.what();
+      }
     }
   }
 #ifdef TRITON_ENABLE_GPU


rmccorm4:
Sharing some notes for future reference on why the if condition was still entered even without an async event loop defined:
```cpp
if (!py::isinstance<py::none>(async_event_loop_)) {
      async_event_loop_.attr("stop")();
}
```

This looks like a valid check, because earlier in the code we initialize the event loop to `None` in `Stub::Initialize()`:
```cpp
async_event_loop_ = py::none();
```

However, as Jacky mentioned - we don't actually call `Initialize()` in the first round of loading the model code for handling auto complete. Since the object is left uninitialized, the pybind object default construction is actually akin to a `nullptr` rather than `py::none` as it would typically behave in Python. This is described in the Pybind [Gotchas](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/object.html#gotchas) docs:
> When a wrapper type is default-constructed, it is not a valid Python object (i.e. it is not py::none()). It is simply the same as PyObject* null pointer. To check for this, use static_cast<bool>(my_wrapper).

So in this case:
```cpp
async_event_loop_ == nullptr;
async_event_loop_ != py::none();
```

So it passes the condition for `if (!py::isinstance<py::none>(async_event_loop_))`.

fpetrini15 (approved):
LGTM! Tested changes locally and no longer collect core dumps.

src/pb_stub.cc
```diff
@@ -945,17 +945,19 @@ void
...
-  if (!py::isinstance<py::none>(async_event_loop_)) {
-    async_event_loop_.attr("stop")();
-  }
-  // Call finalize if exists.
-  if (initialized_ && py::hasattr(model_instance_, "finalize")) {
-    try {
-      model_instance_.attr("finalize")();
+  if (initialized_) {
+    // Stop async event loop if created.
+    if (!py::isinstance<py::none>(async_event_loop_)) {
```
rmccorm4, line 950: Tangentially related question - is there a reasoning for not initializing all the `py::object` member variables to `py::none()` in the class declaration given that we know `py::object` [default construction](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/object.html#gotchas) leads to possibly unexpected behavior as `nullptr`s?

```cpp
class Stub {
...
private:
...
  py::object model_instance_ = py::none();
  py::object deserialize_bytes_ = py::none();
  py::object serialize_bytes_ = py::none();
  py::object async_event_loop_ = py::none();
}
```

This would also avoid the unexpected entering of this condition.

Tabrizian (approved):

PR #356, opened 2024-05-03
Add error handling in case of AutocompleteStub Failure for DLIS-5819

indrajit96:
Tested various failures in Load/Unload API
Handling for error in LaunchAutoCompleteStubProcess() was not complete causing the HTTP request to hang, in case of faulty autocomplete function.
Attached new behaviour in the image
Screenshot 2024-05-03 at 10 29 37 AM

Followup question:

I saw similar handling in ALL other backends as well, I think a fix would be needed in all of them.

### DIFF
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -2042,7 +2042,7 @@ ModelState::LaunchAutoCompleteStubProcess()
         std::string("unexpected nullptr in BackendModelException"));
     RETURN_IF_ERROR(ex.err_);
     return TRITONSERVER_ErrorNew(
-    TRITONSERVER_ERROR_INTERNAL, TRITONSERVER_ErrorMessage(ex.err_));    
+        TRITONSERVER_ERROR_INTERNAL, TRITONSERVER_ErrorMessage(ex.err_));
   }
 
   return nullptr;
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -2041,8 +2041,6 @@ ModelState::LaunchAutoCompleteStubProcess()
         ex.err_ == nullptr, TRITONSERVER_ERROR_INTERNAL,
         std::string("unexpected nullptr in BackendModelException"));
     RETURN_IF_ERROR(ex.err_);
-    return TRITONSERVER_ErrorNew(
-        TRITONSERVER_ERROR_INTERNAL, TRITONSERVER_ErrorMessage(ex.err_));
   }
 
   return nullptr;
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -2041,6 +2041,8 @@ ModelState::LaunchAutoCompleteStubProcess()
         ex.err_ == nullptr, TRITONSERVER_ERROR_INTERNAL,
         std::string("unexpected nullptr in BackendModelException"));
     RETURN_IF_ERROR(ex.err_);
+    return TRITONSERVER_ErrorNew(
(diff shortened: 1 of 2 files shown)


src/python_be.cc
```diff
@@ -2041,6 +2041,8 @@ ModelState::LaunchAutoCompleteStubProcess()
         ex.err_ == nullptr, TRITONSERVER_ERROR_INTERNAL,
         std::string("unexpected nullptr in BackendModelException"));
     RETURN_IF_ERROR(ex.err_);
+    return TRITONSERVER_ErrorNew(
```
tanmayv25, line 2044: Trying to understand here. Shouldn't the RETURN_IF_ERROR call in L2043 detect ex.err_ is not nullptr and returned the error object?

Tabrizian (changes requested):
Is there a test case that I can learn more about the issue.

src/python_be.cc
rmccorm4: +1, I had the same thought

indrajit96: Hi @rmccorm4 and @tanmayv25,
Initially my assumption was RETURN_IF_ERROR(ex.err_); is not returning for some reason hence I added a ruturn statement manually.
This seemed to fix the issue.

Later turned out the way I was reproducing the issue was inconsistent and this statement has no impact on the bug :)
After spending quite some time, I was able to consistently reproduce.

Steps to repro

1. Load a "correct" model succesfully.
2. Try loading a "in-correct/faulty" model after step 1. Triton will hang.

If we directly try to  
Load a "in-correct/faulty" model Triton erros out and returns 400 (expected and correct behavior)

krishung5: @indrajit96 Thanks for the details! Are you able to get the backtrace and see which line causes the hang? Is the hang still related to the auto-complete stub after the new repro? If disabling the auto-complete, will the hang still occur?

indrajit96: Resolved this thread after discussion with @krishung5 and also reverted that chunk of code

src/stub_launcher.cc
```diff
@@ -442,7 +442,9 @@ StubLauncher::Launch()
         stub_message_queue_.reset();
         parent_message_queue_.reset();
         memory_manager_.reset();
-        WaitForStubProcess();
+        if (stub_pid_ != 0) {
```
krishung5, line 445: Can we add a comment here to explain that the check is required because the stub may have been terminated if there's some error occurred - just for future reference.

Tabrizian (approved):
Nice fix!

tanmayv25 (approved):

src/stub_launcher.cc
```diff
@@ -442,7 +442,11 @@ StubLauncher::Launch()
...
```
rmccorm4, line 445: Question for my own curiosity @indrajit96 @Tabrizian @krishung5: 

Are the other spots that call `WaitForStubProcess()` not at risk for similar issues?

Does it make sense to just move this if statement within `WaitForStubProcess()` instead? Should it ever wait on PID 0?

krishung5: Great question! I think usually `WaitForStubProcess()` comes with killing the stub process, so it's the error handling cases that we need to be cautious. I think it makes sense to move the check within `WaitForStubProcess()`. The stub pid is reset to 0 when it's killed, so it shouldn't be waiting on PID 0.

indrajit96: +1, Will move it inside and test once

krishung5 (approved):

PR #361, opened 2024-05-22
Fix decoupled batch statistics to account for implicit batch size

kthui:
Previous PRs:
* https://github.com/triton-inference-server/server/pull/7257
* https://github.com/triton-inference-server/python_backend/pull/360

The decoupled metrics reporting should consider implicit batch size when reporting batch statistics. Previously, the batch statistics uses the number of requests as batch size, which does not account for batched request(s). Now, it is modified to take the number of requests multiplied by its respective implicit batch size.

Next PRs:
* https://github.com/triton-inference-server/server/pull/7258
* https://github.com/triton-inference-server/python_backend/pull/362

PR #362, opened 2024-05-22
Fix decoupled gpu output error handling

kthui:
Previous PR: https://github.com/triton-inference-server/python_backend/pull/361
Related PR: https://github.com/triton-inference-server/server/pull/7258

Fix two things:
* The decoupled GPU tensors output does not check if the GPU output buffers are successfully allocated. This adds the check.
* If there is an error in buffer, it will throw an exception instead of logging the exception. The stub/parent turn flag will be updated correctly if an exception is raised or completed successfully.

Enhance one thing:
* The error message returned to the request will match the full one printed to the log, instead of only returning the basic message from the model.

Next PRs:
* https://github.com/triton-inference-server/server/pull/7292
* https://github.com/triton-inference-server/python_backend/pull/363

### DIFF
--- a/src/response_sender.cc
+++ b/src/response_sender.cc
@@ -147,9 +147,26 @@ ResponseSender::Send(
   }
 
   if (has_gpu_output) {
+    ScopedDefer _([send_message_payload] {
+      bi::scoped_lock<bi::interprocess_mutex> guard{send_message_payload->mu};
+      send_message_payload->is_stub_turn = false;
+      send_message_payload->cv.notify_one();
+      while (!send_message_payload->is_stub_turn) {
+        // Wait for the stub process to send the response and populate error
+        // message if any.
+        send_message_payload->cv.wait(guard);
+      }
+    });
+
     AllocatedSharedMemory<GPUBuffersShm> gpu_buffers_handle =
         shm_pool_->Load<GPUBuffersShm>(
             send_message_payload->gpu_buffers_handle);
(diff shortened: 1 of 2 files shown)


src/response_sender.cc
```diff
@@ -147,22 +147,38 @@ ResponseSender::Send(
...
+        // Wait for the stub process to send the response and populate error
+        // message if any.
+        send_message_payload->cv.wait(guard);
+      }
+    });
+
     AllocatedSharedMemory<GPUBuffersShm> gpu_buffers_handle =
         shm_pool_->Load<GPUBuffersShm>(
             send_message_payload->gpu_buffers_handle);
+    if (!gpu_buffers_handle.data_->success) {
```
Tabrizian, line 164: Great enhancement for adding error handling in GPU buffer loading. Do we also have any test coverage for this part?

kthui: Yes, the existing [test_dlpack_tensor_error()](https://github.com/triton-inference-server/server/blob/main/qa/L0_backend_python/python_test.py#L206) will test this part. The test currently test the non-decoupled data pipeline, once this is merged, it will test the decoupled data pipeline which requires this change to pass.

kthui:
Rebase onto https://github.com/triton-inference-server/python_backend/tree/jacky-res-sender-main for merging.

Tabrizian (approved):

PR #364, opened 2024-06-01
Add support for response sender in the default mode

kthui:
Related PR: https://github.com/triton-inference-server/server/pull/7311

Add support for non-decoupled models to use response sender. Removed the non-decoupled data pipeline.

Items to be done on another ticket:
* Re-implement restart functionality to the decoupled data pipeline. [This functionality is never mentioned on documentation, so it is ok to skip it for now]

Logistics:
* All commits on this PR came from previously approved PRs (less copyright/documentation updates), the purpose of this PR is to merge the feature into the main branch.

### DIFF
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -571,7 +571,8 @@ ModelInstanceState::ExecuteBLSRequest(
           reinterpret_cast<bi::managed_external_buffer::handle_t*>(
               request_batch.data_.get() + sizeof(RequestBatch));
       infer_request = InferRequest::LoadFromSharedMemory(
-          Stub()->ShmPool(), *request_handle, false /* open_cuda_handle */);
+          Stub()->ShmPool(), *request_handle, false /* open_cuda_handle */,
+          nullptr /* is_model_decoupled */);
 
       // If the BLS inputs are in GPU an additional round trip between the
       // stub process and the main process is required. The reason is that we
--- a/src/python_be.cc
+++ b/src/python_be.cc
@@ -153,124 +153,6 @@ ModelInstanceState::SetErrorForResponseSendMessage(
   }
 }
 
-void
-ModelInstanceState::SendMessageAndReceiveResponse(
-    bi::managed_external_buffer::handle_t message,
-    bi::managed_external_buffer::handle_t& response, bool& restart,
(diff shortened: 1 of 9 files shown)


Tabrizian (approved):

src/python_be.cc
```diff
@@ -408,24 +290,15 @@ ModelInstanceState::SaveRequestsToSharedMemory(
...
-    } else {
-      infer_request = std::make_unique<InferRequest>(
-          id, correlation_id, pb_input_tensors, requested_output_names,
-          model_state->Name(), model_state->Version(), parameters_string, flags,
-          request_timeout, 0 /* response_factory_address */,
-          reinterpret_cast<intptr_t>(request),
-          PreferredMemory(PreferredMemory::kDefault, 0), trace);
-    }
+    TRITONBACKEND_ResponseFactory* factory_ptr;
+    RETURN_IF_ERROR(TRITONBACKEND_ResponseFactoryNew(&factory_ptr, request));
```
krishung5, line 294: Question for my understanding - I see that we are creating ResponseFactory for each request regardless of if the model is decoupled or not, later on the ResponseFactory will be cleaned up when the response sender goes out of scope. Is this correct?

src/python_be.cc
```diff
@@ -449,11 +322,9 @@ ModelInstanceState::LaunchStubProcess()
   thread_pool_ = std::make_unique<boost::asio::thread_pool>(
       model_state->StateForBackend()->thread_pool_size);
 
-  if (model_state->IsDecoupled()) {
-    decoupled_thread_ = true;
-    decoupled_monitor_ =
-        std::thread(&ModelInstanceState::DecoupledMessageQueueMonitor, this);
-  }
+  queue_monitor_thread_ = true;
+  decoupled_monitor_ =
```
krishung5, line 326: Would it make sense to rename it to `queue_monitor` or something to avoid confusion?

krishung5 (approved):

src/python_be.cc
kthui: Sure, updated.

krishung5 (approved):

Tabrizian (approved):

PR #365, opened 2024-06-06
fix: [precaution fix] Capture Python futures while running in the background

kthui:
#### What does the PR do?
The `py_future` object generated by `RunCoroutine()` is lost upon the function returns if the coroutine is to be ran in the background. This can be problematic as some sources suggest the `run_coroutine_threadsafe()` only maintain a weak reference to the `py_future` object that it returns. Therefore, this PR will implement the workaround suggested on the [official Python documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) which stores the `py_future` in a Python set and remove it upon the done callback is invoked, so the `py_future` cannot be dangling while waiting for it to execute in the background.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
N/A

N/A

#### Test plan:
No additional tests were added. The current tests, i.e. L0_backend_python should provide sufficient coverage for any new issue introduced by the change.

- CI Pipeline ID:
#15608296

#### Caveats:
We can explore if the Python library actually has weak reference while waiting for background futures.

#### Background
The future object returned from `asyncio.run_coroutine_threadsafe(...)` is not captured by the stub process while waiting for the future object to complete execution in the background. If the `run_coroutine_threadsafe()` only maintains a weak reference to the future object, the future object may be deleted by the time the result is ready.

N/A

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -537,7 +537,7 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
   c_python_backend_utils.attr("shared_memory") = py::cast(shm_pool_.get());
 
   async_event_loop_ = py::none();
-  background_futures_ = py::module_::import("builtins").attr("set")();
+  background_futures_ = py::set();
 
   py::object TritonPythonModel = sys.attr("TritonPythonModel");
   deserialize_bytes_ = python_backend_utils.attr("deserialize_bytes_tensor");
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -829,6 +829,15 @@ Stub::RunCoroutine(py::object coroutine, bool in_background)
 void
 Stub::BackgroundFutureDone(const py::object& py_future)
 {
+  ScopedDefer _([this, &py_future] {
+    // Remove future from background
+    try {
+      background_futures_.attr("remove")(py_future);
+    }
+    catch (const py::error_already_set& error) {
+      LOG_ERROR << "Cannot remove future from background; " << error.what();
+    }
+  });
   // TODO: Why using `py_future.result()` with error hangs on exit?
   try {
     py::object exception = py_future.attr("exception")();
@@ -849,13 +858,6 @@ Stub::BackgroundFutureDone(const py::object& py_future)
   }
   catch (const py::error_already_set& error) {
     LOG_ERROR << error.what();
-  }
-  // Remove future from background
-  try {
-    background_futures_.attr("remove")(py_future);
-  }
-  catch (const py::error_already_set& error) {
-    LOG_ERROR << "Cannot remove future from background; " << error.what();
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/pb_stub.cc
```diff
@@ -556,6 +537,7 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
   c_python_backend_utils.attr("shared_memory") = py::cast(shm_pool_.get());
 
   async_event_loop_ = py::none();
+  background_futures_ = py::module_::import("builtins").attr("set")();
```
Tabrizian, line 540:
```suggestion
  background_futures_ = py::set();
```

https://github.com/pybind/pybind11/blob/9b3a2000653a002abe0fe96ce8bd002f9a67f3d4/tests/test_pytypes.cpp#L160-L163

src/pb_stub.cc
```diff
@@ -838,11 +820,45 @@ Stub::RunCoroutine(py::object coroutine, bool in_background)
...
         py::module_::import("c_python_backend_utils")
             .attr("async_event_future_done_callback"));
+    background_futures_.attr("add")(py_future);
     return py::none();
   }
   return py_future.attr("result")();
 }
 
+void
+Stub::BackgroundFutureDone(const py::object& py_future)
```
Tabrizian, line 830: nit: I don't think we require a new function. Let's have the body in the `AsyncEventFutureDoneCallback`.

src/pb_stub.cc
kthui: Nice finding! Something I was looking for but couldn't find on the official docs. Updated: [Use pybind11 provided python set](https://github.com/triton-inference-server/python_backend/pull/365/commits/063fdfe6d8f040cdd1ea6a722bee123f5236ef97)

src/pb_stub.cc
```diff
@@ -838,11 +820,45 @@ Stub::RunCoroutine(py::object coroutine, bool in_background)
...
+  }
+  catch (const PythonBackendException& pb_exception) {
+    LOG_ERROR << pb_exception.what();
+  }
+  catch (const py::error_already_set& error) {
+    LOG_ERROR << error.what();
+  }
+  // Remove future from background
+  try {
+    background_futures_.attr("remove")(py_future);
```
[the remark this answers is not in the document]
kthui: I wonder if we are going to see an exception that is neither a `PythonBackendException` nor `py::error_already_set`? Yes, we always want to remove it to be on the safe side, [Scoped defer background future removal](https://github.com/triton-inference-server/python_backend/pull/365/commits/57d9d9600dc81a3c83e1b66304fcd63b2f896749)

GuanLuo (commented):
Can you update the title / description following the PR template?

Tabrizian (approved):

src/pb_stub.cc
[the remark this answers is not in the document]
Tabrizian: > I wonder if we are going to see an exception that is neither a PythonBackendException nor py::error_already_set

We're going to return from the function which should run the `ScopedDefer` task. I think we would remove the future even when the exception is not caught.

krishung5 (approved):

PR #366, opened 2024-06-10
fix: Models should filter outputs based on requested outputs

kthui:
#### What does the PR do?
Models on Python backend should filter outputs based on requested outputs from the client.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/server/pull/7338

N/A

#### Test plan:
Add additional tests to L0_backend_python to affirm the behavior of non-decoupled and decoupled models on filtering outputs from requested outputs.

- CI Pipeline ID:
Refer to the related server PR.

#### Caveats:
N/A

#### Background
The non-decoupled data pipeline previously does filter outputs based on requested output, so this PR essentially brings this behavior back after switching non-decoupled models to use decoupled data pipeline.

N/A

### DIFF
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -68,14 +68,15 @@ InferRequest::InferRequest(
     }
   }
 
-  inputs_ = inputs;
-  requested_output_names_ = requested_output_names;
+  inputs_ = inputs;                                  // TODO: do we need this?
+  requested_output_names_ = requested_output_names;  // TODO: do we need this?
 #ifdef TRITON_PB_STUB
   pb_cancel_ =
       std::make_shared<PbCancel>(response_factory_address_, request_address_);
   response_sender_ = std::make_shared<ResponseSender>(
       request_address_, response_factory_address_, nullptr /* is_decoupled */,
-      Stub::GetOrCreateInstance()->SharedMemory(), pb_cancel_);
+      RequestedOutputNames(), Stub::GetOrCreateInstance()->SharedMemory(),
+      pb_cancel_);
 #endif
 }
 
@@ -390,7 +391,8 @@ InferRequest::InferRequest(
       std::make_shared<PbCancel>(response_factory_address_, request_address_);
   response_sender_ = std::make_shared<ResponseSender>(
       request_address_, response_factory_address_, is_model_decoupled,
-      Stub::GetOrCreateInstance()->SharedMemory(), pb_cancel_);
+      RequestedOutputNames(), Stub::GetOrCreateInstance()->SharedMemory(),
+      pb_cancel_);
 #endif
 }
 
--- a/src/infer_request.cc
+++ b/src/infer_request.cc
@@ -68,8 +68,6 @@ InferRequest::InferRequest(
     }
   }
 
-  inputs_ = inputs;                                  // TODO: do we need this?
-  requested_output_names_ = requested_output_names;  // TODO: do we need this?
 #ifdef TRITON_PB_STUB
   pb_cancel_ =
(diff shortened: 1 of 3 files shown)


src/response_sender.cc
```diff
@@ -110,6 +119,16 @@ ResponseSender::UpdateStateAndCounters(
   number_of_response_sent_++;
 }
 
+void
+ResponseSender::PruneNonRequestedOutputs(
+    const std::shared_ptr<InferResponse>& infer_response) const
+{
+  // TODO: should this be limited to non decoupled only?
+  if (!IsDecoupled() && infer_response) {
```
Tabrizian, line 127: If we remove the `!IsDecoupled()` is it going to just work for decoupled?

src/infer_request.cc
```diff
@@ -68,14 +68,15 @@ InferRequest::InferRequest(
     }
   }
 
-  inputs_ = inputs;
-  requested_output_names_ = requested_output_names;
+  inputs_ = inputs;                                  // TODO: do we need this?
```
Tabrizian, line 71: Why do we need `TODO`? It is clearly required.

kthui: The two variables are already initialized to the ones on the constructor argument, i.e.
```c++
InferRequest::InferRequest(...) : ..., inputs_(inputs), requested_output_names_(requested_output_names), ...
{
  ...
  inputs_ = inputs;
  requested_output_names_ = requested_output_names;
}
```
I am not sure why we are assigning them again? The second copy seems unnecessary?

rmccorm4: Looks redundant to me when you put it like that :nerd_face: 

This PR LGTM, I'm fine with removing this redundancy if compilation and tests all look good :+1:

GuanLuo: Should be able to remove

GuanLuo (approved):

src/response_sender.cc
```diff
@@ -123,6 +126,9 @@ ResponseSender::Send(
 
   CheckResponseSenderArguments(infer_response, flags);
   UpdateStateAndCounters(infer_response, flags);
+  if (infer_response) {
+    infer_response->PruneOutputTensors(requested_output_names_);
```
GuanLuo, line 130: `requested_output_names_` is guaranteed to be full set of outputs if not specified by the client?

src/infer_request.cc
Tabrizian: I see, yes, we can remove these. I thought Jacky was referring to the attributes themselves (i.e. remove `inputs_` and `requested_output_names_` and not the copy)..

Tabrizian (approved):

src/response_sender.cc
kthui: Yes, because underneath it is calling [this two C APIs](https://github.com/triton-inference-server/core/blob/main/src/backend_model.cc#L1292-L1333) and the `ImmutableRequestedOutputs()` used by both will check if `requested_outputs_` is empty and use `original_requested_outputs_` if so.

src/infer_request.cc
kthui: Removed: [[chore] Remove redundant copy](https://github.com/triton-inference-server/python_backend/pull/366/commits/b745ae54906a2b36c72a6f5a29b38411f99f06f9)

rmccorm4 (approved):

PR #367, opened 2024-06-12
fix: Models should filter outputs based on requested outputs (#366)

kthui:
* Prune non requested outputs from non-decoupled models

* Prune non requested outputs from decoupled models

* [chore] Remove redundant copy


kthui:
Pick https://github.com/triton-inference-server/python_backend/pull/366 into r24.06

PR #372, opened 2024-07-30
perf: Release GIL during cancellation check

kthui:
#### What does the PR do?

Release the GIL while checking for cancellation. Another Python thread can be performing useful work while the cancellation status is being checked.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [x] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

https://github.com/triton-inference-server/vllm_backend/pull/54

Can start by verifying the cancellation function, where the GIL is released during its execution, does not access any Python objects.

#### Test plan:

This is a performance improvement, so any issue should be covered by existing test cases.

- CI Pipeline ID: 17061062

#### Caveats:

N/A

#### Background

Need to improve the performance on vLLM backend.

N/A

### DIFF
--- a/src/pb_cancel.cc
+++ b/src/pb_cancel.cc
@@ -1,4 +1,4 @@
-// Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


Tabrizian (approved):

src/pb_cancel.cc
```diff
@@ -57,6 +57,9 @@ PbCancel::ShmPayload()
 bool
 PbCancel::IsCancelled()
 {
+  // Release the GIL. Python objects are not accessed during the check.
```
oandreeva-nv, line 60: If it's not too much work, can we elaborate performance benefits in the comment above or in the PR description?

kthui: Sure added to the PR description.

oandreeva-nv (approved):
LGTM!

PR #374, opened 2024-08-07
feat: Add histogram metric type

yinggeh:
#### What does the PR do?
Support histogram metric type and its functions.

#### Checklist
- [ ] All template sections are filled out.

#### Commit Type:
- [x] feat

#### Related PRs:
https://github.com/triton-inference-server/vllm_backend/pull/56
https://github.com/triton-inference-server/core/pull/386
https://github.com/triton-inference-server/server/pull/7525

n/a

#### Test plan:
n/a

- CI Pipeline ID:
17487728

#### Caveats:
n/a

#### Background
Customer requested histogram metrics from vLLM.

n/a

### DIFF
--- a/src/metric.cc
+++ b/src/metric.cc
@@ -1,4 +1,4 @@
-// Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
--- a/src/metric.cc
+++ b/src/metric.cc
@@ -358,6 +358,16 @@ Metric::Increment(const double& value)
 void
 Metric::SetValue(const double& value)
 {
+  // SetValue and Observe share the same C API TRITONSERVER_MetricSet.
+  // Throws if SetValue is called by a histogram metric.
+  TRITONSERVER_MetricKind kind;
+  THROW_IF_TRITON_ERROR(TRITONSERVER_GetMetricKind(
+      reinterpret_cast<TRITONSERVER_Metric*>(metric_address_), &kind));
+  if (kind == TRITONSERVER_METRIC_KIND_HISTOGRAM) {
+    throw PythonBackendException(
+        "TRITONSERVER_METRIC_KIND_HISTOGRAM does not support SetValue");
+  }
+
   auto triton_metric = reinterpret_cast<TRITONSERVER_Metric*>(metric_address_);
   THROW_IF_TRITON_ERROR(TRITONSERVER_MetricSet(triton_metric, value));
 }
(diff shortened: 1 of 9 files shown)


oandreeva-nv:
I think we need general tests for this on python backend as well

src/pb_stub.cc
```diff
@@ -1824,11 +1824,13 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
   py::class_<Metric, std::shared_ptr<Metric>>(module, "Metric")
       .def("increment", &Metric::SendIncrementRequest)
       .def("set", &Metric::SendSetValueRequest)
+      .def("observe", &Metric::SendObserveRequest)
       .def("value", &Metric::SendGetValueRequest);
 
   py::enum_<MetricKind>(module, "MetricKind")
       .value("COUNTER", MetricKind::kCounter)
       .value("GAUGE", MetricKind::kGauge)
+      .value("HISTOGRAM", MetricKind::kHistogram)
```
rmccorm4, line 1833: +1 on @oandreeva-nv comment for testing in python backend too:

Can add to the existing custom metrics tests:
- https://github.com/triton-inference-server/server/blob/main/qa/L0_backend_python/custom_metrics/test.sh
- https://github.com/triton-inference-server/server/blob/main/qa/python_models/custom_metrics/model.py

krishung5:
+1 on the test. Also I think would be great if we can add the documentation for the new type of metrics: https://github.com/triton-inference-server/python_backend/tree/main/examples/custom_metrics

src/metric.cc
```diff
@@ -281,6 +362,13 @@ Metric::SetValue(const double& value)
   THROW_IF_TRITON_ERROR(TRITONSERVER_MetricSet(triton_metric, value));
 }
 
+void
+Metric::Observe(const double& value)
```
Tabrizian, line 366: Why not just use `Set`?

yinggeh: In Prometheus, gauge calls `Set` to set itself to a value, while histogram and summary kinds call `Observe` to record a value internally. I think `Set` and `Observe` have different meanings although they are the same API. It also makes the call more clear to people who's familiar with Prometheus metrics. Open to suggestions.
https://github.com/jupp0r/prometheus-cpp/blob/master/core/include/prometheus/histogram.h#L58
https://github.com/jupp0r/prometheus-cpp/blob/master/core/include/prometheus/summary.h#L84

src/pb_stub.cc
yinggeh: Tests added https://github.com/triton-inference-server/server/pull/7525

src/metric.cc
Tabrizian: Why we don't do it in core then?

Tabrizian: This is an internal implementation, let's keep it consistent with what we have in core.

GuanLuo: Agree, please keep in sync with the core API.

yinggeh:
> +1 on the test. Also I think would be great if we can add the documentation for the new type of metrics: https://github.com/triton-inference-server/python_backend/tree/main/examples/custom_metrics

Updated. See [commit](https://github.com/triton-inference-server/python_backend/pull/374/commits/e36181229e55ff0b07e12d4cb49949cd55a008b6)

GuanLuo (approved):

krishung5 (approved):

Tabrizian (approved):

src/metric.cc
yinggeh: Resolved.

PR #390, opened 2024-12-11
fix: Fix requested output deleting extra outputs

kthui:
#### What does the PR do?
RCA: When deleting NumPy arrays in C++, GIL must be held.

Fixes:
* Ensure GIL is held when deleting NumPy arrays behind tensor objects - fix RCA.
* Update shutdown step to ensure Python objects are actually deleted during destructor execution.

#### Checklist
- [ ] Verified that the PR passes existing CI.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/server/pull/7866

Look at this PR first, and then move on to the related PR for the new test.

#### Test plan:
A new test is added to the related PR that enhances requested output testing.

- CI Pipeline ID: 21321591

#### Caveats:
N/A

#### Background
N/A

N/A

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -1032,9 +1032,9 @@ Stub::~Stub()
 
   {
     py::gil_scoped_acquire acquire;
-    async_event_loop_ = py::none();
-    background_futures_ = py::none();
-    model_instance_ = py::none();
+    py::object async_event_loop_local(std::move(async_event_loop_));
+    py::object background_futures_local(std::move(background_futures_));
+    py::object model_instance_local(std::move(model_instance_));
   }
   stub_instance_.reset();
   stub_message_queue_.reset();
--- a/src/pb_tensor.cc
+++ b/src/pb_tensor.cc
@@ -503,6 +503,14 @@ PbTensor::~PbTensor() noexcept(false)
 {
   pb_memory_.reset();
   DeleteDLPack();
+
+#ifdef TRITON_PB_STUB
+  {
+    py::gil_scoped_acquire acquire;
+    py::array numpy_array_local(std::move(numpy_array_));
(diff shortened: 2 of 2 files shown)


src/pb_stub.cc
```diff
@@ -1032,9 +1032,9 @@ Stub::~Stub()
 
   {
     py::gil_scoped_acquire acquire;
-    async_event_loop_ = py::none();
-    background_futures_ = py::none();
-    model_instance_ = py::none();
+    py::object async_event_loop_local(std::move(async_event_loop_));
+    py::object background_futures_local(std::move(background_futures_));
+    py::object model_instance_local(std::move(model_instance_));
```
krishung5, line 1037: Just for my own understanding - does this mean that `async_event_loop_ = py::none();` will not actually delete the python object?

kthui: It may not. When I tried the set to none method in the tensor destructor, the object was not destructed.

krishung5 (approved):
Nice find, Jacky!

src/pb_stub.cc
nv-kmcgill53: You have to decrement the reference of the object for it to be garbage collected. Once the reference count of the object goes to zero, then the object will be picked for garbage collection. My guess is that assigning these variables to `py::none()` didn't decrement the reference count on these objects if you were to inspect the properties at runtime again. I think this idea is also reinforced since you use the `std::move` operator which does reference counting on the smart pointers being used. All this should be taken with a caveat that I haven't worked in this part of the code and is only based on this PR and your conversations.

PR #394, opened 2025-01-24
feat: Add parameters support to InferResponse

kthui:
#### What does the PR do?
Add support for setting response parameters in regular and decoupled Python backend model response(s).

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/server/pull/7964

Start with the test cases on the related server PR, and then `pb_stub.cc` -> `infer_response.cc` -> `python_be.cc`.

#### Test plan:
New tests are added to the related server PR.

- CI Pipeline ID: 22968260

#### Caveats:
Responses to BLS models are not populated with the response parameters, if any. (DLIS-7864)

#### Background
N/A

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,6 +803,9 @@ You can read more about the inference response parameters in the [parameters
 extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_parameters.md)
 documentation.
 
+Inference response parameters is currently not supported on BLS inference
+responses received by BLS models.
+
 ## Managing Python Runtime and Libraries
 
 Python backend shipped in the [NVIDIA GPU Cloud](https://ngc.nvidia.com/)
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <!--
-# Copyright 2020-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2020-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -54,6 +54,7 @@ any C++ code.
     - [`finalize`](#finalize)
   - [Model Config File](#model-config-file)
   - [Inference Request Parameters](#inference-request-parameters)
+  - [Inference Response Parameters](#inference-response-parameters)
   - [Managing Python Runtime and Libraries](#managing-python-runtime-and-libraries)
     - [Building Custom Python Backend Stub](#building-custom-python-backend-stub)
     - [Creating Custom Execution Environments](#creating-custom-execution-environments)
@@ -784,6 +785,21 @@ request = pb_utils.InferenceRequest(parameters={"key": "value"}, ...)
 ```
 
 You can read more about the inference request parameters in the [parameters
+extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_parameters.md)
+documentation.
+
+## Inference Response Parameters
+
+Inference response parameters may be optionally set during the construction of
+an inference response object. The parameters should be a dictionary of key value
+pairs, where keys are `str` and values are `bool`, `int` or `str`. For example,
+```python
+response = pb_utils.InferenceResponse(
+    output_tensors, parameters={"key": "value"}
+)
+```
+
+You can read more about the inference response parameters in the [parameters
 extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_parameters.md)
 documentation.
 
--- a/src/infer_response.cc
+++ b/src/infer_response.cc
@@ -169,7 +169,8 @@ InferResponse::LoadFromSharedMemory(
             response_shm.data_.get() + sizeof(ResponseShm));
     {
 #ifdef TRITON_PB_STUB
-      // Need to acquire the GIL to avoid hangs.
+      // PbTensor::LoadFromSharedMemory() will construct Python objects if
+      // called from pb_stub, which requires holding the GIL.
       py::gil_scoped_acquire acquire;
 #endif
       for (size_t idx = 0; idx < requested_output_count; ++idx) {
--- a/src/infer_response.cc
+++ b/src/infer_response.cc
@@ -1,4 +1,4 @@
-// Copyright 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2021-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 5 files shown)


src/infer_response.cc
```diff
@@ -58,6 +60,12 @@ InferResponse::OutputTensors()
   return output_tensors_;
 }
 
+std::string&
```
rmccorm4, line 63: Should it be mutable or `const`?

src/infer_response.h
```diff
@@ -72,9 +73,10 @@ class InferResponse {
  public:
   InferResponse(
       const std::vector<std::shared_ptr<PbTensor>>& output_tensors,
-      std::shared_ptr<PbError> error = nullptr,
+      std::shared_ptr<PbError> error = nullptr, std::string parameters = "",
       const bool is_last_response = true, void* id = nullptr);
   std::vector<std::shared_ptr<PbTensor>>& OutputTensors();
+  std::string& Parameters();
```
rmccorm4, line 79: Looks like this parameters string is expected to be JSON serializable - can you add some docstrings describing that expectation?

src/infer_response.cc
rmccorm4, line 1: Can we add a README section for response parameters similar to the one for request parameters? https://github.com/triton-inference-server/python_backend?tab=readme-ov-file#inference-request-parameters

src/pb_stub.cc
```diff
@@ -1869,16 +1869,43 @@ PYBIND11_EMBEDDED_MODULE(c_python_backend_utils, module)
...
+                PyDefaultArgumentToMutableType<py::dict>(parameters_);
+            for (const auto& pair : parameters) {
+              if (!py::isinstance<py::str>(pair.first)) {
+                throw PythonBackendException(
+                    "Expect parameters keys to have type str, found type " +
+                    std::string(py::str(pair.first.get_type())));
+              }
+              if (!py::isinstance<py::bool_>(pair.second) &&
+                  !py::isinstance<py::int_>(pair.second) &&
+                  !py::isinstance<py::str>(pair.second)) {
```
rmccorm4, line 1886: Should we have any protection against modifying reserved parameter keys such as `triton_final_response` from user-code?

kthui: I did a quick check on the current behavior, if it is a
* regular model: `triton_final_response` set by the backend is passed to the client unmodified;
* decoupled model: `triton_final_response` is overwritten by the frontend, i.e. [gRPC here](https://github.com/triton-inference-server/server/blob/r24.12/src/grpc/stream_infer_handler.cc#L769)

> Should we have any protection against modifying reserved parameter keys such as triton_final_response from user-code?

I don't think it is the responsibility of the backend to safeguard any frontend "reserved" parameter keys, because the response object is originally created at the backend and the core does not limit which key can/cannot be set on the C-API response object, then it is up to the frontend to handle the case where a parameter key it wants to set is already set on the response object it is returning.

README.md
```diff
@@ -787,6 +788,21 @@ You can read more about the inference request parameters in the [parameters
...
+pairs, where keys are `str` and values are `bool`, `int` or `str`. For example,
+```python
+response = pb_utils.InferenceResponse(
+    output_tensors, parameters={"key": "value"}
+)
+```
+
+You can read more about the inference response parameters in the [parameters
+extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_parameters.md)
+documentation.
```
rmccorm4, line 804: Let's add a note here on the limitation with BLS models until implemented

src/infer_response.cc
kthui: It should be `const`. Updated [Mark response parameters accessor const and JSON serializable](https://github.com/triton-inference-server/python_backend/pull/394/commits/f23114704d0a6135962b72ccb88b55425c48b5b0)

src/infer_response.h
kthui: Yes, added the comment: [Mark response parameters accessor const and JSON serializable](https://github.com/triton-inference-server/python_backend/pull/394/commits/f23114704d0a6135962b72ccb88b55425c48b5b0)

src/pb_stub.cc
rmccorm4: Sounds good to me - as long as we gave it some thought 🙂  We can always add a check later if we run into an issue.

README.md
kthui: Sure, added [[Docs] Note BLS response parameters are not populated currently](https://github.com/triton-inference-server/python_backend/pull/394/commits/47e71e0c6334e4b8fa1f491db47a1782c739e910)

src/infer_response.cc
```diff
@@ -154,33 +167,43 @@ InferResponse::LoadFromSharedMemory(
     bi::managed_external_buffer::handle_t* tensor_handle_shm =
         reinterpret_cast<bi::managed_external_buffer::handle_t*>(
             response_shm.data_.get() + sizeof(ResponseShm));
+    {
```
rmccorm4, line 170: Why does the `PbTensor::LoadFromSharedMemory` block for tensor need scoped GIL held, but the `PbString::LoadFromSharedMemory` block for parameters doesn't?

If you have more context on the scoped gil usage that was previously here, I think this comment could be improved - (needs to be held because XYZ is also contending for the GIL?):
```
      // Need to acquire the GIL to avoid hangs 
```

rmccorm4 (commented):
General look&feel LGTM - left a couple questions. Mostly defer to @krishung5 for an expert opinion

src/infer_response.cc
kthui: The `PbTensor::LoadFromSharedMemory` may [access Python object](https://github.com/triton-inference-server/python_backend/blob/r24.12/src/pb_tensor.cc#L667), which requires holding the GIL, but the parameters are all C++ objects, so GIL is not required.

I'll update the comment to better describe the GIL requirement.

krishung5 (approved):
🚀

src/infer_response.cc
kthui: Updated comment: https://github.com/triton-inference-server/python_backend/pull/394/commits/1acb95691182a9b3ff0be3eaac72ba8b4dc03273

Tabrizian:
Thanks Jacky!

README.md
```diff
@@ -803,6 +803,9 @@ You can read more about the inference response parameters in the [parameters
 extension](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_parameters.md)
 documentation.
 
+Inference response parameters is currently not supported on BLS inference
```
nnshah1, line 806: not sure this is clear ....

PR #401, opened 2025-05-08
Fix: Launch the python backend on Windows when spaces exist in the path.

vinitkmagnet:
There were no quotes around any of the paths in the stub launcher. As a result, if you tried launching the stub, it would always fail. Normally this is not an issue in Linux, but having spaces is very common in Windows. Added quotes around all the pathing of the launcher code to resolve this problem.

### DIFF
--- a/src/stub_launcher.cc
+++ b/src/stub_launcher.cc
@@ -240,10 +240,10 @@ StubLauncher::Launch()
   std::string launch_command;
 
   std::stringstream ss;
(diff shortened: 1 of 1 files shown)


vinitkmagnet:
Hi @yinggeh and @mattwittwer, I have created a new PR in to the requested branch https://github.com/triton-inference-server/python_backend/pull/412. Hope this is what you were requesting. If not, let me know and I can correct my mistake.

PR #430, opened 2026-02-16
feat: Support user-defined `is_ready()` in Python backend readiness checks

pskiran1:
This PR adds support for user-defined `is_ready()` functions in Python backend models, allowing custom readiness checks beyond basic stub health verification. The implementation uses IPC (Inter-Process Communication) between the parent process and the Python stub to invoke the user's readiness function when `TRITONBACKEND_ModelInstanceReady` (health endpoint - `v2/models/<model>/ready`) is called.

**Changes:**
- Added IPC infrastructure for user-defined model readiness checks with a 5-second timeout
- Cached the presence of `is_ready()` function during stub initialization to optimize the common case (fast path when no custom function exists)
- Implemented thread-safe result caching to avoid multiple concurrent IPC calls for the same readiness check

### DIFF
--- a/src/ipc_message.h
+++ b/src/ipc_message.h
@@ -69,8 +69,7 @@ typedef enum PYTHONSTUB_commandtype_enum {
   PYTHONSTUB_ModelReadinessRequest,
   PYTHONSTUB_IsRequestCancelled,
   PYTHONSTUB_CancelBLSInferRequest,
-  PYTHONSTUB_UserModelReadinessRequest,
-  PYTHONSTUB_UserModelReadinessResponse
+  PYTHONSTUB_UserModelReadinessRequest
 } PYTHONSTUB_CommandType;
 
 ///
--- a/src/ipc_message.h
+++ b/src/ipc_message.h
@@ -1,4 +1,4 @@
-// Copyright 2021-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -68,7 +68,9 @@ typedef enum PYTHONSTUB_commandtype_enum {
   PYTHONSTUB_UnloadModelRequest,
   PYTHONSTUB_ModelReadinessRequest,
   PYTHONSTUB_IsRequestCancelled,
-  PYTHONSTUB_CancelBLSInferRequest
+  PYTHONSTUB_CancelBLSInferRequest,
+  PYTHONSTUB_UserModelReadinessRequest,
+  PYTHONSTUB_UserModelReadinessResponse
 } PYTHONSTUB_CommandType;
 
 ///
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -1591,15 +1591,13 @@ Stub::ProcessUserModelReadinessRequest(std::unique_ptr<IPCMessage>& ipc_message)
     readiness_payload = readiness_message.data_.get();
   }
   catch (const PythonBackendException& pb_exception) {
-    LOG_ERROR
-        << "Failed to process model readiness request: "
-        << pb_exception.what();
+    LOG_ERROR << "Failed to process model readiness request: "
+              << pb_exception.what();
     return;
   }
 
   if (ipc_message->ResponseMutex() == nullptr) {
-    LOG_ERROR
-        << "Failed to process model readiness request";
+    LOG_ERROR << "Failed to process model readiness request";
     return;
   }
 
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -596,7 +596,7 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
   }
 
   // Cache whether is_model_ready() function defined in the Python model.
-  ipc_control_->stub_has_model_ready_fn =
+  ipc_control_->stub_has_user_model_readiness_fn =
       py::hasattr(model_instance_, "is_model_ready");
 
   initialized_ = true;
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -1,4 +1,4 @@
-// Copyright 2021-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -594,6 +594,10 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
   if (py::hasattr(model_instance_, "initialize")) {
     model_instance_.attr("initialize")(model_config_params);
   }
+
+  // Cache whether is_model_ready() function defined in the Python model.
+  ipc_control_->stub_has_model_ready_fn =
+      py::hasattr(model_instance_, "is_model_ready");
 
   initialized_ = true;
 }
@@ -1350,6 +1354,9 @@ Stub::ParentToStubMQMonitor()
       case PYTHONSTUB_CommandType::PYTHONSTUB_InferStreamExecResponse: {
         ProcessBLSResponseDecoupled(ipc_message);
       } break;
+      case PYTHONSTUB_CommandType::PYTHONSTUB_UserModelReadinessRequest: {
+        ProcessUserModelReadinessRequest(ipc_message);
+      } break;
       default:
         break;
     }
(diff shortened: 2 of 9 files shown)


src/python_be.cc
```diff
@@ -895,6 +895,241 @@ ModelInstanceState::ProcessCancelBLSRequest(
...
+
+  try {
+    ipc_message =
+        IPCMessage::Create(Stub()->ShmPool(), true /* inline_response */);
+    readiness_message =
+        Stub()->ShmPool()->Construct<UserModelReadinessMessage>();
+    readiness_payload = readiness_message.data_.get();
+
+    // Initialize payload fields
+    new (&(readiness_payload->mu)) bi::interprocess_mutex;
```
whoisj, line 1055: we know that no other thread is waiting on this mutex before we `new` it?

Same goes for the CV below?

yinggeh:
Does `is_model_ready` always return boolead result? To be consistent with `Initialize`, `Execute` and `Finalize`, consider `is_ready`, `get_readiness`, etc.

src/python_be.cc
```diff
@@ -895,6 +895,241 @@ ModelInstanceState::ProcessCancelBLSRequest(
...
+          // otherwise shared-memory may be deallocated before stub reads it.
+          ScheduleUserModelReadinessCleanupTask(
+              std::move(ipc_message), std::move(readiness_message));
+
+          return TRITONSERVER_ErrorNew(
+              TRITONSERVER_ERROR_UNAVAILABLE,
+              "Timed out waiting for user-defined is_model_ready() response");
+        }
+      }
+    }
```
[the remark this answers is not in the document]
pskiran1: Added IsStubProcessAlive() check.

src/pb_stub.cc
```diff
@@ -1573,6 +1580,99 @@ Stub::ProcessBLSResponseDecoupled(std::unique_ptr<IPCMessage>& ipc_message)
...
+        shm_pool_->Load<UserModelReadinessMessage>(ipc_message->Args());
+    readiness_payload = readiness_message.data_.get();
+  }
+  catch (const PythonBackendException& pb_exception) {
+    return;
+  }
+
+  if (ipc_message->ResponseMutex() == nullptr) {
+    return;
+  }
```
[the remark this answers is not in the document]
pskiran1: Added error log

src/python_be.cc
pskiran1: `Construct<UserModelReadinessMessage>()` gives a fresh shared-memory allocation for this request. At that point, no handle is published yet, so no other thread/process can access this payload. We initialize `mu` and `cv` before pushing the IPC message handle to the queue.

pskiran1:
> Does `is_model_ready` always return boolead result? To be consistent with `Initialize`, `Execute` and `Finalize`, consider `is_ready`, `get_readiness`, etc.

Yes, `is_model_ready` must always return a boolean. We enforce this in `pb_stub.cc using py::isinstance<py::bool_>(result)`, and any non-boolean return is treated as an error/not-ready path.

I have chosen the name `is_model_ready` as it is more descriptive and matches endpoint naming `(/v2/models/<model>/ready, TRITONBACKEND_ModelInstanceReady)`. Please let me know if you recommend changing it.

whoisj (approved):
LGTM, please address @yinggeh's comments before merging.

yinggeh:
Need to update docs on usage and examples.

src/pb_utils.h
```diff
@@ -53,6 +53,10 @@ namespace triton { namespace backend { namespace python {
 
 namespace bi = boost::interprocess;
 
+// Timeout for user-defined model readiness requests
+// and mutex locks (in milliseconds)
+constexpr uint64_t kUserModelReadinessTimeoutMs = 5000;
```
pskiran1, line 58: @yinggeh, @whoisj, I currently set the ready request timeout to 5 seconds to ensure users keep it lightweight. If it times out, it returns the model as not ready. Please let me know if this timeout still seems too long and whether we should reduce it to around 1–2 seconds.

yinggeh:
I would like to see use cases listed in the doc. How can this feature be useful to users?

src/ipc_message.h
```diff
@@ -68,7 +68,8 @@ typedef enum PYTHONSTUB_commandtype_enum {
   PYTHONSTUB_UnloadModelRequest,
   PYTHONSTUB_ModelReadinessRequest,
   PYTHONSTUB_IsRequestCancelled,
-  PYTHONSTUB_CancelBLSInferRequest
+  PYTHONSTUB_CancelBLSInferRequest,
+  PYTHONSTUB_UserModelReadinessRequest
```
yinggeh, line 72: What's the difference between `PYTHONSTUB_ModelReadinessRequest` and `PYTHONSTUB_UserModelReadinessRequest`?

yinggeh:
> > Does `is_model_ready` always return boolead result? To be consistent with `Initialize`, `Execute` and `Finalize`, consider `is_ready`, `get_readiness`, etc.
> 
> Yes, `is_model_ready` must always return a boolean. We enforce this in `pb_stub.cc using py::isinstance<py::bool_>(result)`, and any non-boolean return is treated as an error/not-ready path.
> 
> I have chosen the name `is_model_ready` as it is more descriptive and matches endpoint naming `(/v2/models/<model>/ready, TRITONBACKEND_ModelInstanceReady)`. Please let me know if you recommend changing it.

I prefer `set_readiness` than `is_model_ready`.

src/ipc_message.h
pskiran1: `PYTHONSTUB_ModelReadinessRequest` is a pre-existing command that travels from the _stub to the parent_. It is used by `pb_utils.is_model_ready(model_name, model_version)` in user Python code to query Triton core about whether a different model is ready (e.g., for BLS dependency checks).

`PYTHONSTUB_UserModelReadinessRequest` is the new command introduced in this PR. It travels in the opposite direction from the _parent to the stub_. It is sent when the `v2/models/<model>/ready` health endpoint is called, and it invokes the user-defined `is_model_ready()` function on this model instance.

pskiran1:
> > > Does `is_model_ready` always return boolead result? To be consistent with `Initialize`, `Execute` and `Finalize`, consider `is_ready`, `get_readiness`, etc.
> > 
> > 
> > Yes, `is_model_ready` must always return a boolean. We enforce this in `pb_stub.cc using py::isinstance<py::bool_>(result)`, and any non-boolean return is treated as an error/not-ready path.
> > I have chosen the name `is_model_ready` as it is more descriptive and matches endpoint naming `(/v2/models/<model>/ready, TRITONBACKEND_ModelInstanceReady)`. Please let me know if you recommend changing it.
> 
> I prefer `set_readiness` than `is_model_ready`.

I think `set_readiness` implies a setter that actively changes state (e.g., set_readiness(True/False)), but `is_model_ready` function is a read-only check, it takes no arguments and returns a boolean. The `is_` prefix aligns with the standard Python convention for boolean predicates (`is_alive()`, `isdigit()`, `isfile()`, etc.) and is consistent with the existing Triton client API (`client.is_model_ready()`, `client.is_server_ready()`).

pskiran1:
> I would like to see use cases listed in the doc. How can this feature be useful to users?

I added some generic use cases to the documentation.

src/pb_utils.h
yinggeh: 5 sec is fine

src/stub_launcher.h
```diff
@@ -136,6 +136,15 @@ class StubLauncher {
   // Is Healthy
   bool IsHealthy() { return is_healthy_; }
 
+  // Returns true if is_model_ready() is defined in the Python model.
+  // Reads directly from shared memory (set by the stub during initialization).
+  // Returns false if not set yet (safe default).
+  bool HasUserModelReadinessFunction() const
+  {
+    return ipc_control_ ? ipc_control_->stub_has_user_model_readiness_fn
+                        : false;
```
whoisj:
```suggestion
    return ipc_control_ && ipc_control_->stub_has_user_model_readiness_fn;
```

yinggeh:
I am interested in one corner case -- what's the result of `model/ready` request if `is_model_ready` returns `true` but model is not fully loaded or python stub process is not responsive? I expect response is `false` during such case.

yinggeh:
> > > > Does `is_model_ready` always return boolead result? To be consistent with `Initialize`, `Execute` and `Finalize`, consider `is_ready`, `get_readiness`, etc.
> > > 
> > > 
> > > Yes, `is_model_ready` must always return a boolean. We enforce this in `pb_stub.cc using py::isinstance<py::bool_>(result)`, and any non-boolean return is treated as an error/not-ready path.
> > > I have chosen the name `is_model_ready` as it is more descriptive and matches endpoint naming `(/v2/models/<model>/ready, TRITONBACKEND_ModelInstanceReady)`. Please let me know if you recommend changing it.
> > 
> > 
> > I prefer `set_readiness` than `is_model_ready`.
> 
> I think `set_readiness` implies a setter that actively changes state (e.g., set_readiness(True/False)), but `is_model_ready` function is a read-only check, it takes no arguments and returns a boolean. The `is_` prefix aligns with the standard Python convention for boolean predicates (`is_alive()`, `isdigit()`, `isfile()`, etc.) and is consistent with the existing Triton client API (`client.is_model_ready()`, `client.is_server_ready()`).

Forgive my nit-picking and thanks for the explanation. Since this function will never be changed once it goes public, I want to make sure the naming has the best description. I agree `set_readiness` may imply state changes and I am fine with going `is_xxx`  naming convention.

I am still a bit uncertain on `is_model_ready` which I feel a bit verbose. For example, is it more concise to go with `is_ready` since the other functions are not named with `initialize_model` or `execute_request`, etc. I am looking forward your input @whoisj.

whoisj:
> I am still a bit uncertain on is_model_ready which I feel a bit verbose. For example, is it more concise to go with is_ready since the other functions are not named with initialize_model or execute_request, etc. I am looking forward your input @whoisj.

In a more generic context, `is_model_ready()` would be the correct name, however this is effectively a callback delegate that we're asking the user to provide. In this case, I believe @yinggeh is probebaly correct and `is_ready()` is a more apt name for the function because we're specifically asking their entire solution "is it ready yet?", this could include one or more models, or even zero models.

whoisj (approved):
LGTM

yinggeh:
Please update PR title/decription, Slack review, ticket with the new API name `is_ready`.

pskiran1:
> I am interested in one corner case -- what's the result of `model/ready` request if `is_model_ready` returns `true` but model is not fully loaded or python stub process is not responsive? I expect response is `false` during such case.

I verified the above scenarios, and it correctly returned `false` for all of them.

yinggeh (approved):

PR #432, opened 2026-02-23
Fix: Inconsistent doc

mehadi92:
`Note: Async BLS is not supported on Python 3.6 or lower due to the async keyword and asyncio.run being introduced in Python 3.7.` 

vs 

`Async BLS is not supported when running a Python model in decoupled mode.`


menon92:
@pskiran1 please check this

PR #433, opened 2026-03-17
fix: Use Correct ptr Type to Avoid UAF

whoisj:
This change changes the managed pointer type used to manage the lifecyle of the stub instance from unique to shared. This avoids the dereference of the oft used unique pointer reference after the unique pointer has been free / deallocated.

CI Pipeline ID: 46362355

### DIFF
--- a/src/pb_stub.cc
+++ b/src/pb_stub.cc
@@ -130,7 +130,7 @@ PyParametersToJSON(const py::dict& parameters)
 void
 AsyncEventFutureDoneCallback(const py::object& py_future)
 {
-  std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
+  auto stub = Stub::GetOrCreateInstance();
   stub->BackgroundFutureDone(py_future);
 }
 
@@ -514,7 +514,7 @@ Stub::AutoCompleteModelConfig(
   python_backend_utils.def(
       "get_model_dir",
       []() {
-        std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
+        auto stub = Stub::GetOrCreateInstance();
         return stub->GetModelDir();
       },
       py::return_value_policy::reference);
@@ -568,7 +568,7 @@ Stub::Initialize(bi::managed_external_buffer::handle_t map_handle)
   python_backend_utils.def(
       "get_model_dir",
       []() {
-        std::unique_ptr<Stub>& stub = Stub::GetOrCreateInstance();
+        auto stub = Stub::GetOrCreateInstance();
         return stub->GetModelDir();
       },
       py::return_value_policy::reference);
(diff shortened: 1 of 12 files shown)


src/pb_stub.cc
```diff
@@ -2148,7 +2148,7 @@ main(int argc, char** argv)
 #endif
   std::atomic<bool> background_thread_running = {true};
   std::thread background_thread =
-      std::thread([&parent_pid, &background_thread_running, &stub, &logger] {
+      std::thread([stub, &parent_pid, &background_thread_running, &logger] {
```
yinggeh, line 2157: Does the order matter here?

src/pb_stub.cc
```diff
@@ -2180,7 +2180,6 @@ main(int argc, char** argv)
 
             // Destroy stub and exit.
             logger.reset();
-            stub.reset();
```
yinggeh, line 2183: Why is the explicit reset removed for stub?

whoisj: I don't see the point of this call. The reference is going to out of scope immediately afterwards anyways.

src/pb_stub.cc
whoisj: the compiler complained about the copy being mixed in with the references, so I just moved it to avoid the warning.

src/pb_stub.cc
whoisj: I stand correct, and I fixed up the logic.

src/pb_stub.cc
mudit-eng: Ordering does not matter for lambda capture but since stub got changes to shared_ptr, it is passed by value instead of a reference. This will increase the reference count and avoids dangling reference risk.

src/pb_stub.cc
```diff
@@ -1073,16 +1073,22 @@ Stub::~Stub()
   memory_manager_message_queue_.reset();
 }
 
-std::unique_ptr<Stub> Stub::stub_instance_;
+static std::shared_ptr<triton::backend::python::Stub> stub_instance{nullptr};
```
mudit-eng, line 1076: Did the compiler complain about std::shared_ptr<Stub>?

Why didn't we make the change to pb_stub.h so that we can use `stub_instance_`?

src/pb_stub.h
mudit-eng: Do we keep this line? 

`static std::unique_ptr<Stub> stub_instance_;`

yinggeh:
Copyrights

src/pb_stub.h
whoisj: I've removed this old field. static member fields are not really member fields, and exposing the static variable via the header provides a mechanism for unintentional risks. Better to declare static global variables in the .cc file that uses them.

whoisj:
> Copyrights

Good catch, thanks. Fixed up.

src/pb_stub.cc
whoisj: removed it.

mudit-eng: Curious, why did you write `std::shared_ptr<triton::backend::python::Stub>` and not `std::shared_ptr<Stub>`? Was compiler complaining?

yinggeh (approved):

mrpre:
@whoisj 
Hi, can you confirm whether this patch is intended to fix the following crash pattern?

In our old python_backend version, Stub::GetOrCreateInstance() returns std::unique_ptr<Stub>&. The main thread runs `stub->RunCommand()`, while the background health thread captures
`&stub` and calls `stub.reset()` when the parent process is detected dead.

Our core shows:
```
Thread 1:
__new_sem_wait_slow
-> Stub::PopMessage()
-> Stub::RunCommand()
```
The faulting instruction is `lock add QWORD PTR [rbx], rax`, with `rbx == si_addr`, `si_code == SEGV_MAPERR`. The address is the semaphore inside the shared memory message queue.

Another thread:
```
main::{lambda()#1}
-> Stub::~Stub()
-> SharedMemoryManager::~SharedMemoryManager()
```
So it looks like the health thread destroys Stub/SharedMemoryManager and unmaps shm while the main thread is still blocked in sem_wait on `mq_shm_ptr_->sem_full`.

Does this commit’s change from `unique_ptr&` to `shared_ptr` by value cover this exact use-after-free / use-after-unmap scenario?

PR #436, opened 2026-05-08
Env amends

Ero-Sennin9:


src/pb_env.cc
```diff
@@ -239,13 +242,16 @@ EnvironmentManager::EnvironmentManager()
   strcpy(base_path_, tmp_dir_template);
 }
 
-std::string
-EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
+std::shared_ptr<Environment> // TODO: write logic with shared and weak ptrs in this method
+EnvironmentManager::GetEnvironment(ModelState* model_state)
```
aleksn7: Let’s preserve the old name

src/pb_env.cc
```diff
@@ -270,23 +276,27 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
                      "not contain compressed path. Path: ") +
          canonical_env_path)
             .c_str());
-    return canonical_env_path;
+    return std::make_shared<Environment>(canonical_env_path, canonical_env_path);
   }
-  const auto env_itr = env_map_.find(canonical_env_path);
+
+  std::string model_id = model_state.GetModelId();
+  std::string env_key =  model_state->Name() + "-" + model_id;
```
aleksn7: Why do we need to add model name here? Multiple models could refer to the same environment

src/pb_env.cc
```diff
@@ -270,23 +276,27 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+  const auto env_itr = env_map_.find(env_key);
   if (env_itr != env_map_.end()) {
     // Check if the environment has been modified and would
-    // need to be extracted again.
-    if (env_itr->second.second == last_modified_time) {
+    // need to be extracted again (or the current environment has no owners anymore).
+    if (env_itr->second.first.expired()) {
```
aleksn7: This condition is executed only if env has already been destroyed and model has to recreate it

Also, when deleting the environment, we want to remove its key from the map


In this case, we can’t derive the environment directory name from the map size, because that could cause environments to overwrite each other.

src/pb_env.cc
```diff
@@ -270,23 +276,27 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+      env_map_.erase(env_itr);
+    } else if (env_itr->second.second == last_modified_time) {
       env_extracted = true;
```
aleksn7, line 301: We could store `last_modified_time` in Environment structure

src/pb_env.cc
```diff
@@ -270,23 +276,27 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
                      "not contain compressed path. Path: ") +
          canonical_env_path)
             .c_str());
-    return canonical_env_path;
+    return std::make_shared<Environment>(canonical_env_path, canonical_env_path);
```
aleksn7: We should return nullptr here instead

src/pb_env.cc
```diff
@@ -295,7 +305,7 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
             .c_str());
     std::string dst_env_path;
     if (re_extraction) {
-      dst_env_path = env_map_[canonical_env_path].first;
+      dst_env_path = env_map_[env_key].first;
```
aleksn7: I don't think this will compile
It's better to call the `->Path()` method here

src/pb_env.cc
```diff
@@ -305,30 +315,50 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+  return env_map_.find(env_key)->second.first.lock();
 }
 
 EnvironmentManager::~EnvironmentManager()
 {
   RecursiveDirectoryDelete(base_path_);
 }
+
+Environment::Environment(const std::string& source, const std::string& path) : source_(source), path_(path) {
+  if (source == path) {
```
aleksn7: Let's check this condition before creating the object.
Also, in this case, destructor will be called with an invalid `Path` value

src/pb_env.h
```diff
@@ -27,9 +27,12 @@
 #pragma once
 #include <climits>
 #include <map>
+#include <memory>
 #include <mutex>
 #include <string>
 
+#include "python_be.h"
```
aleksn7: We don't need this include if we remove `model_state` from the `GetEnvironment` signature

src/stub_launcher.cc
```diff
@@ -592,20 +592,19 @@ StubLauncher::Launch()
 TRITONSERVER_Error*
 StubLauncher::GetPythonEnvironment(ModelState* model_state)
 {
-  std::string python_execution_env = "";
   try {
-    python_execution_env =
-        model_state->StateForBackend()->env_manager->ExtractIfNotExtracted(
-            python_execution_env_);
+    python_execution_env_ =
+        std::move(model_state->StateForBackend()->env_manager->GetEnvironment(model_state));
```
aleksn7: We don't need this `std::move` here because copy elision will work here

src/pb_env.cc
aleksn7: I think `EnvironmentManager` doesn't need to know about `model_state`

src/pb_env.cc
```diff
@@ -291,44 +300,84 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
-      ExtractTarFile(canonical_env_path_str, dst_env_path);
-    } else {
-      throw PythonBackendException(
-          std::string("Failed to create environment directory for '") +
-          dst_env_path.c_str() + "'.");
-    }
-    if (re_extraction) {
-      // Just update the last modified timestamp
-      env_map_[canonical_env_path].second = last_modified_time;
+    if (re_extraction ) {
```
aleksn7: Let's merge this condition with the condition on line [308](https://github.com/triton-inference-server/python_backend/pull/436/changes#diff-8e22e2a1f41271c8dd4a71fa49f87dae28619d2afe6893bc250cb50e4dda0fb2L297)

src/pb_env.cc
```diff
@@ -291,44 +300,84 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+      // Just replace with new environment (by updated source)
+      env->Update(last_modified_time);
     } else {
-      // Add the path to the list of environments
-      env_map_.insert({canonical_env_path, {dst_env_path, last_modified_time}});
+      // Add the environment to the list of environments
+      env = std::make_shared<Environment>(
+        canonical_env_path_str, dst_env_path, last_modified_time);
+      env->SetManager(this);
```
aleksn7: Why do we need a separate method for this? Let's pass it into the Environment constructor

src/pb_env.cc
```diff
@@ -291,44 +300,84 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+void
+Environment::Delete()
+{
+  RecursiveDirectoryDelete(path_.c_str());
+}
+
+Environment::~Environment()
+{
+  if (manager_ != nullptr) {
+    manager_->env_map_.erase(source_);
```
aleksn7: This map is protected by a mutex inside `manager_`
Let's add a method inside `EnvironmentManager` struct that locks the mutex and erases environment from the map.

src/pb_env.h
```diff
@@ -46,17 +47,46 @@ bool FileExists(std::string& path);
...
+        const std::string& source, const std::string& path,
+        const time_t& last_modified_time);
+    void SetManager(EnvironmentManager* manager) { manager_ = manager; }
+
+    void Update(const time_t& last_modified_time);
+
+    const std::string& Source() const { return source_; }
+    const std::string& Path() const { return path_; }
+    const time_t& LastModifiedTime() const { return last_modified_time_; }
+    explicit operator std::string() const { return Path(); }
```
aleksn7: Useless operator

src/pb_env.cc
```diff
@@ -270,19 +275,30 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
                      "not contain compressed path. Path: ") +
          canonical_env_path)
             .c_str());
-    return canonical_env_path;
+    return nullptr;
   }
-  const auto env_itr = env_map_.find(canonical_env_path);
+
+  std::string canonical_env_path_str(canonical_env_path);
+  std::string env_key = canonical_env_path_str;
```
aleksn7: Why do we need three identical variables `canonical_env_path_str`, `env_key` and `canonical_env_path`?

I would write something like this:

```c++
std::string canonical_env_path = [&]{
    char canonical_env_path[PATH_MAX + 1];
    char* err = realpath(env_path.c_str(), canonical_env_path);
    if (err == nullptr) {
        throw PythonBackendException(
            std::string("Failed to get the canonical path for ") + env_path + ".");
    }
    return std::string(canonical_env_path);
}();
```

And drop the other variables: `canonical_env_path_str`, `env_key`

src/pb_env.cc
```diff
@@ -270,19 +275,30 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+    env = env_itr->second.lock();
+
     // Check if the environment has been modified and would
-    // need to be extracted again.
-    if (env_itr->second.second == last_modified_time) {
+    // need to be extracted again (or the current environment has no owners
+    // anymore).
+
+    if (env == nullptr) {
+      // refer to case when env was not loaded
```
aleksn7: We will enter this condition if an environment with such a key existed and was dropped earlier
Not because the environment hasn't been loaded yet

src/pb_env.cc
```diff
@@ -239,13 +242,15 @@ EnvironmentManager::EnvironmentManager()
   strcpy(base_path_, tmp_dir_template);
 }
 
-std::string
-EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
+
+std::shared_ptr<
+    Environment>  // TODO: write logic with shared and weak ptrs in this method
```
aleksn7: Let's drop the useless comment here

src/pb_env.cc
Ero-Sennin9: it simulates when env was not loaded, because we are deleting notice about it in map

Ero-Sennin9: I said: "refers to case", but not "this is case when"

src/pb_env.cc
```diff
@@ -262,27 +270,37 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+        ("Returning canonical path since EXECUTION_ENV_PATH does "
+                     "not contain compressed path. Path: " +
          canonical_env_path)
             .c_str());
-    return canonical_env_path;
+    return nullptr;
   }
-  const auto env_itr = env_map_.find(canonical_env_path);
+
+  std::string& env_key = canonical_env_path;
```
aleksn7: Do we really need this variable?

src/pb_env.cc
```diff
@@ -291,44 +311,90 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
+          Environment(env_path, dst_env_path, last_modified_time);
+      // Add the environment to the list of environments
+      insert({env_key, new_env});
+      env = &env_map_[env_key]
     }
+  }
 
-    std::string canonical_env_path_str(canonical_env_path);
+  if (!re_extraction) {
+    env->AddOwner();
```
aleksn7: We should increment reference count on each `ExtractIfNotExtracted` call

src/pb_env.cc
```diff
@@ -239,50 +242,67 @@ EnvironmentManager::EnvironmentManager()
...
+        ("Returning canonical path since EXECUTION_ENV_PATH does "
+         "not contain compressed path. Path: " +
          canonical_env_path)
             .c_str());
-    return canonical_env_path;
+    return nullptr;
   }
-  const auto env_itr = env_map_.find(canonical_env_path);
+
+  return EnvironmentGuard(*this, canonical_env_path);
```
aleksn7: Let's create environment here and pass it explicitly to the constructor

src/pb_env.h
```diff
@@ -46,17 +48,77 @@ bool FileExists(std::string& path);
...
+    public:
+      EnvironmentGuard(EnvironmentManager* manager, Environment* environment);
+      const EnvironmentProxy* operator->() const { return &environment_proxy_; }
+      const EnvironmentProxy& operator*() const { return environment_proxy_; }
+      ~EnvironmentGuard();
+
+   private:    
+    EnvironmentManager* manager_;
+    Environment* environment_;
+    EnvironmentProxy environment_proxy_;
```
aleksn7, line 113: I don't think that we need a separate field for this
We could build it on the fly directly inside the operator’s body.

src/pb_env.cc
```diff
@@ -239,50 +242,68 @@ EnvironmentManager::EnvironmentManager()
...
+  // Lock the mutex. Only a single thread should modify the map.
+  std::lock_guard<std::mutex> lk(mutex_);
+
+  time_t last_modified_time;
+  LastModifiedTime(env_path, &last_modified_time);
+
+  bool env_extracted = false;
+  bool re_extraction = false;
+
+  const std::string& env_key = env_path;
```
aleksn7, line 291: Let's use env_path everywhere instead of `env_key`

src/pb_env.h
Ero-Sennin9: Okay, how we would return pointer to the temporal object of the function scope?

src/pb_env.cc
```diff
@@ -239,50 +243,67 @@ EnvironmentManager::EnvironmentManager()
...
+  auto env_itr = env_map_.find(env_key);
+  Environment* env = nullptr;
   if (env_itr != env_map_.end()) {
+    env = &env_itr->second;
+
     // Check if the environment has been modified and would
-    // need to be extracted again.
-    if (env_itr->second.second == last_modified_time) {
+    // need to be extracted again (or the current environment has no owners
+    // anymore).
```
Ero-Sennin9: old comment that is need to be deleted

src/pb_env.cc
Ero-Sennin9: But this is done this way already, am I right?

src/pb_env.cc
```diff
@@ -291,44 +312,114 @@ EnvironmentManager::ExtractIfNotExtracted(std::string env_path)
...
-          std::string(base_path_) + "/" + std::to_string(env_map_.size());
+      std::string dst_env_path =
+          std::string(base_path_) + "/" + std::to_string(env_path_counter_);
+      ++env_path_counter_;
+
+      // Add the environment to the list of environments
+      env_itr =
+          env_map_
+              .try_emplace(env_key, env_path, dst_env_path, last_modified_time)
+              .first;
```
Ero-Sennin9: what a mess with the formatting

src/pb_env.h
```diff
@@ -27,7 +27,9 @@
 #pragma once
 #include <climits>
 #include <map>
+#include <memory>
```
Ero-Sennin9: for what

src/pb_env.h
Ero-Sennin9: I assume we don't need environment_ field, we need from env only source, that can be provided by proxy

src/pb_env.h
```diff
@@ -46,17 +48,92 @@ bool FileExists(std::string& path);
...
   ~EnvironmentManager();
+
+ private:
+  void DropEnvironment(Environment& environment);
+  Environment& GetEnvironment(const std::string& env_path);
+
+  size_t env_path_counter_ = 0;
+  std::map<std::string, Environment> env_map_;
+  char base_path_[PATH_MAX + 1];
+  std::mutex mutex_;
```
Ero-Sennin9, line 134: maybe rename to point out that this mutex is for map asynchronous safety

src/pb_env.h
```diff
@@ -46,17 +48,92 @@ bool FileExists(std::string& path);
...
+  std::optional<EnvironmentGuard> ExtractIfNotExtracted(
+      const std::string& env_path);
+
   ~EnvironmentManager();
+
+ private:
+  void DropEnvironment(Environment& environment);
+  Environment& GetEnvironment(const std::string& env_path);
+
+  size_t env_path_counter_ = 0;
```
Ero-Sennin9: useless field

src/pb_env.cc
Ero-Sennin9: notice it

src/pb_env.cc
Ero-Sennin9: notice it

src/pb_env.h
Ero-Sennin9: leave it for readability

src/pb_env.h
Ero-Sennin9: obviously not!

PR #437, opened 2026-05-20
fix: Validate archive entry paths when unpacking `EXECUTION_ENV_PATH`

pskiran1:
Adds `ARCHIVE_EXTRACT_SECURE_NODOTDOT`, `ARCHIVE_EXTRACT_SECURE_NOABSOLUTEPATHS`, and `ARCHIVE_EXTRACT_SECURE_SYMLINKS` to the `libarchive` options used when extracting the `EXECUTION_ENV_PATH` tarball, so that archive entries are placed under the model's per-stub extraction directory. Also checks the return of `archive_write_disk_set_options` and surfaces its error string.

### DIFF
--- a/src/pb_env.cc
+++ b/src/pb_env.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2021-2026, NVIDIA CORPORATION. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


whoisj (approved):
LGTM. Do we have a test where this failure occurs?

If not, please create a Linear ticket to provide one and add it to our backlog.

src/pb_env.cc
yinggeh:
```suggestion
// Copyright (c) 2021-2026, NVIDIA CORPORATION. All rights reserved.
```

yinggeh (approved):

